PR #17, opened 2018-07-03
Fix release date stamp and tag deployment dependency

rbayliss:
This PR fixes the release dating and also fixes the dependency issue we had with tag deployment.

Steps to test:
- [x] Visit https://mayflower.digital.mass.gov/b/patternlab/fix_release_dating/index.html and verify that the date is correct.  
- [x] Visit https://circleci.com/gh/massgov/mayflower/214 and verify that the only failure comes from NPM, which won't let us deploy over the 5.29.0 version.


mrossi113 (approved):
👍 LGTM! This will fix deployment and sync master back to develop.

PR #19, opened 2018-07-05
release minor version to s3

clairesunstudio:
This pr allows pointing to a static mayflower patternlab minor version

### DIFF
--- a/patternlab/styleguide/tools/gulp/PatternLab.js
+++ b/patternlab/styleguide/tools/gulp/PatternLab.js
@@ -74,14 +74,14 @@ class PatternLabRegistry extends MayflowerRegistry {
         taker.task("patternlab:release", taker.series(
             maybe(isStableRelease,
                 taker.series(
-                    this.buildPatternlabTask("%domain%/v/%%major%/%%minor%"),
+                    this.buildPatternlabTask("%domain%/v/%major%/%minor%"),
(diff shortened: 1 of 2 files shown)


patternlab/styleguide/tools/gulp/PatternLab.js
```diff
@@ -72,6 +72,13 @@ class PatternLabRegistry extends MayflowerRegistry {
 
         // Perform a release to S3.
         taker.task("patternlab:release", taker.series(
+            maybe(isStableRelease,
+                taker.series(
+                    this.buildPatternlabTask("%domain%/v/%%major%/%%minor%"),
```
rbayliss: Ouch, I just realized we've had a typo for a while in this file: `%%major%` should be `%major%` and `%%minor` should be `%minor%`.  As it currently stands, this would build pattern lab with a base url of `https://mayflower.digital.mass.gov/v/%0/%1`, and push to `v/0/%1`. 

Also, it might be even more useful for people if they could rely on a single, set version rather than the minor.  Eg: `v/5.30.0` instead of `v/5/30`.  I think you could do this by replacing what you've got here with `v/%tag%`.

clairesunstudio: updated per your comment f0fdbd2 and def73cf

rbayliss (approved):
Looks awesome, thanks!

PR #25, opened 2018-07-06
Share release notes

clairesunstudio:
- share release notes between projects, updated changelog to:
```
___DESCRIPTION___
Change_type (see below)
Change_impact (see below)
- (Project_prefix) DP-1234: Adds apples to apple trees for admin apple pickers #PR#
```
- remove circle config file in patternlab


changelogs/versioning.md
```diff
@@ -1,9 +1,9 @@
 # Semantic Versioning
 
-The Mayflower Style Library is using [Semantic Versioning](http://semver.org/) to convey meaning about the underlying code and what has been modified from one version to the next to help consumers keep their sites up to date with the latest features and bug fixes.
+The Mayflower Component Libraries is using [Semantic Versioning](http://semver.org/) to convey meaning about the underlying code and what has been modified from one version to the next to help consumers keep their sites up to date with the latest features and bug fixes.
```
rbayliss: Should be "Component Libraries are using"

patternlab/README.md
```diff
@@ -22,7 +22,7 @@ Every time a branch passes CircleCI it will deploy the code to a the following U
 
 ### Production Deployment
 
-Mayflower release managers with the necessary repo permissions can see [Release docs](docs/release.md) for steps on deploying code to production (i.e. do a release).
+Mayflower release managers with the necessary repo permissions can see [Release docs](../release-notes.md) for steps on deploying code to production (i.e. do a release).
```
rbayliss: I don't think this change makes sense.  We're trying to link to the release docs, not the release notes.  Eventually those docs should live at the top level too, but they don't yet.

changelogs/versioning.md
```diff
@@ -1,9 +1,9 @@
 # Semantic Versioning
```
rbayliss: Shouldn't this live in the `docs` folder?  It doesn't make sense to me in the `changelogs` folder.

rbayliss (changes requested):
One thing that's always bugged me is that we don't use the standard `CHANGELOG.md` name though.  Can we fix that here?  I'd hate to see us carry this convention forward.

changelogs/versioning.md
clairesunstudio: right, i was skeptical about this too. I created a `.docs` at the root level to host shared docs like this one. is that a dumb idea?

rbayliss:
This looks good to me.  I fixed a couple of links to the versioning documentation.  I'm going to ask @mrossi113 to take a look here, since she's going to be updating release docs.  Melissa, can you S+M if this makes sense?

rbayliss (approved):

mrossi113 (approved):
👍 LGTM!

patternlab/docs/release.md
```diff
@@ -13,11 +13,11 @@ Mayflower release managers with the necessary repo permissions can follow these
     1. `cd styleguide`
 1. Create a release branch
     1. `git checkout -b release-#.#.#` where `#.#.#` is the next version (i.e. `5.0.0`).  Read more about [Mayflower and semantic versioning](docs/versioning.md) to ensure that your are creating the right type of version.
-    1. Add [release notes](https://github.com/massgov/mayflower/blob/dev/docs/change-log-instructions.md) to the top of [release notes](/release-notes.md) based on the "changelog.txt" files, remove all the "changelog.txt" files and then commit these changes.
+    1. Add [change logs](https://github.com/massgov/mayflower/blob/dev/changelogs) to the top of [CHANGELOG.md](../CHANGELOG.md) based on the "changelog.txt" files, remove all the "changelog.txt" files and then commit these changes.
```
mrossi113: I will fix this link for change logs it pointing to old URL in my PR to update the release process.

PR #66, opened 2018-07-13
DP-9579 Remove raw filter from _patterns/03-organisms/by-author/event-listing-interactive.twig

ygannett:

## Description

Remove raw filter from _patterns/03-organisms/by-author/event-listing-interactive.twig.

By removing the raw filter, the calendar icons in the date fields in the filter are not displayed, and the place holder text in the first field is not displayed from the JSON.

**With raw filter:**

dp-9579_event-filter-filter

**Without raw filter:**

dp-9579_event-listing-no-filter

The filter feature doesn't seem to be implemented in Drupal:

dp-9579_event-listing2

The change doesn't affect the display of the event listing data which is from JSON where the filter was originally applied and the change was applied, but it seems to be better to merge`patternlab/disable_twig_autoescape` to ensure any data from JSON (considering event description data includes links) to be displayed properly.

Template use in Mayflower:
```
 ~/MassIT/mayflower/patternlab/ [feature/DP-9579_organisms_event-listing-interactive] grep -r 'event-listing-interactive.twig' styleguide/source
styleguide/source/_patterns/05-pages/event-listing.twig:  {% include "@organisms/by-author/event-listing-interactive.twig" %}
styleguide/source/assets/js/modules/eventListingInteractive.js:    const rawData = ma.eventListingInteractive[i]; // Data object created in @organisms/by-author/event-listing-interactive.twig
 ~/MassIT/mayflower/patternlab/ [feature/DP-9579_organisms_event-listing-interactive]
```

## Related Issue / Ticket

- [DP-9579](https://jira.mass.gov/browse/DP-9579)

## Steps to Test

1. Go to event listing page and check for any display issue in the event listing information.

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


rbayliss:
Actually, it looks like this might be a valid use case for the raw filter.  We're outputting some JSON here, which is as escaped as it can be.  When we remove `|raw`, it outputs like this:
```
ma.eventListingInteractive.push({&quot;maxItems&quot;:10...
```
Which isn't valid JSON.  So anyway, I think we can actually settle this ticket by leaving the raw filter, but also leaving a Twig comment to explain why it's there.  Something like:
> This is an intentional use of the `raw` filter because the output of json_encode should not be tampered with by escaping.

ygannett:
Raw filter is back with the comment to keep it.

rbayliss (approved):
Makes sense to me.  I'm gonna go ahead and merge this, since it's just the addition of a comment.  Thanks!

PR #70, opened 2018-07-13
DP-9575 Remove raw filter from 02-molecules/header-contact.twig

ygannett:

## Description

Remove raw filter from 02-molecules/header-contact.twig

This template doesn't seem to be used in Mayflower:
```
 ~/MassIT/mayflower/ [feature/DP-9575_molecules-header-contact] grep -r 'header-contact.twig' patternlab/styleguide/source/_patterns
 ~/MassIT/mayflower/ [feature/DP-9575_molecules-header-contact] grep -r 'header-contact.twig' patternlab/styleguide/source/
 ~/MassIT/mayflower/ [feature/DP-9575_molecules-header-contact] grep -r '02-molecules/header-contact.twig' patternlab/styleguide/source/
 ~/MassIT/mayflower/ [feature/DP-9575_molecules-header-contact]
```

## Related Issue / Ticket

- [DP-9575](https://jira.mass.gov/browse/DP-9575)

## Steps to Test

1.  Go to the Event and the Location pages and check for any display issue at the header contact area.

## Screenshots
Event page
dp-9575_mf-event-header-contact

Location page
dp-9575_mf-location-header-contact

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/patternlab/changelogs/DP-9575.txt
+++ b/patternlab/changelogs/DP-9575.txt
@@ -1,0 +1,4 @@
+___DESCRIPTION___
+Removed
+Patch
+- DP-9575: Remove raw filter from 02-molecules/header-contact.twig.
--- a/patternlab/styleguide/source/_patterns/02-molecules/header-contact.twig
+++ b/patternlab/styleguide/source/_patterns/02-molecules/header-contact.twig
@@ -6,7 +6,7 @@
 <section class="ma__header-contact {{accordionParent}}">
   <div class="ma__header-contact__address"{% if schemaPageType %} property="address" typeof="PostalAddress"{% endif %}>
     {% if headerContact.title %}
-      <span{% if schemaPageType %} property="name"{% endif %}>{{headerContact.title | raw}}, </span>
+      <span{% if schemaPageType %} property="name"{% endif %}>{{ headerContact.title }}, </span>
     {% endif %}
     {% if headerContact.address2 %}
       <span>{% if schemaPageType %}<span property="streetAddress">{% endif %}{{headerContact.address1}}{% if schemaPageType %}</span>{% endif %} {{headerContact.address2}}, </span>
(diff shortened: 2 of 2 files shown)


rbayliss:
Huh!  It's not used at all on the Drupal side either.  I guess this is just there as a placeholder and that the HTML is duplicated in several places.  Anyway, looks good to me!

rbayliss (approved):

yriahi (approved):
LGTM

yriahi:
Same comment regarding raw from a previous PR. Not sure if this is a concern; but bringing up just in case.
I read the one pager about raw ([here](https://twig.symfony.com/doc/2.x/filters/raw.html)). Interesting that we are using _raw_ many places; is that a security concern? If so, can _ Twig_Sandbox_SecurityPolicy_ ([here](http://twig.readthedocs.io/en/latest/api.html)) help in such a use case?  Thanks.

rbayliss:
@yriahi - I think the reason we ended up with `raw` in so many places was that Patternlab escapes whatever you put into the JSON data files when it pumps it into Twig, so people were just defaulting to adding raw anytime they might have to pass an HTML string in (😱).  Since escaping in Patternlab is completely unnecessary (we aren't dealing with user provided data here), we've recently set the security policy on Patternlab to not escape anything.  On the Drupal side, having our data pass through the `raw` filter is indeed a security concern, which is a big part of why we're doing this removal.

rbayliss:
Also, it's worth noting that on the Drupal side, what we want to be passing into these templates is not a string - it's a markup object that Twig already understands as safe, and is capable of doing its own escaping.  But since objects can't pass through the `raw` filter, we're having to pre-escape a lot of stuff.  That's the other reason we're doing this... we want to pass those pre-secured objects in rather than simple strings.

PR #77, opened 2018-07-18
Release 5.32.0

mrossi113:
### Added
- (Patternlab) DP-5329: Adds browser history enabled pagination for listing pages. #8


mrossi113:
The DP-5329 passed ER in the release branch.

PR #81, opened 2018-07-19
refactor ButtonWIthIcon to replace ButtonSearch

clairesunstudio:
This PR refactors ButtonWithIcon to replace and archive the ButtonSearch component. It's a step towards creating a fully configurable button pattern.
It also fixes the spacing between the text and the icon in the search button in the Search UI. See the visual differences in Search:
before:
screen shot 2018-07-19 at 7 22 40 pm
after:
screen shot 2018-07-19 at 7 30 47 pm


react/src/components/atoms/buttons/ButtonWithIcon/index.js
```diff
@@ -21,28 +21,33 @@ class ButtonWithIcon extends React.Component {
...
       classNames += ' ma__icon-small';
     }
-    if (this.props.iconColor === 'green') {
+    if (iconColor === 'green') {
       classNames += ' ma__icon-green';
     }
+    if (icon.type.name === 'SvgSearch') {
+      classNames += ' ma__button-search';
+    }
+    console.log(type.name)
```
smurrayatwork: This needs to be removed.

react/src/components/atoms/buttons/ButtonWithIcon/index.js
```diff
@@ -21,28 +21,33 @@ class ButtonWithIcon extends React.Component {
     }
   }
   render() {
-    let classNames = this.props.classes.join(' ');
-    if (this.props.canExpand) {
+    const { classes, canExpand, expanded, capitalized, iconSize, iconColor, icon, type } = this.props
+    let classNames = classes.join(' ');
```
smurrayatwork: Instead of joining this at this line, why not join after all the logic for classNames is done? That way you can just push new items onto the classes array instead of appending potentially empty strings.

react/src/components/atoms/buttons/ButtonWithIcon/index.js
```diff
@@ -21,28 +21,33 @@ class ButtonWithIcon extends React.Component {
     }
   }
   render() {
-    let classNames = this.props.classes.join(' ');
-    if (this.props.canExpand) {
+    const { classes, canExpand, expanded, capitalized, iconSize, iconColor, icon, type } = this.props
```
smurrayatwork: Missing a semicolon here.

react/src/components/atoms/buttons/ButtonWithIcon/index.js
```diff
@@ -21,28 +21,33 @@ class ButtonWithIcon extends React.Component {
...
+    if (iconColor === 'green') {
       classNames += ' ma__icon-green';
     }
+    if (icon.type.name === 'SvgSearch') {
+      classNames += ' ma__button-search';
+    }
+    console.log(type.name)
     const buttonProps = {
-      type: 'submit',
+      type: type,
```
smurrayatwork: This should just be the type variable (object property shorthand).

react/src/components/atoms/buttons/ButtonWithIcon/index.js
```diff
@@ -67,18 +74,26 @@ ButtonWithIcon.propTypes = {
...
   capitalized: PropTypes.bool,
+  // Defines the size, default size fits the most square icons and "small" setting is specific for the chevron icon.
   iconSize: PropTypes.oneOf(['', 'small']),
-  iconColor: PropTypes.oneOf(['', 'green'])
+  // Defines the fill color of the svg, default color is $c-gray-dcdcdc.
+  iconColor: PropTypes.oneOf(['', 'green']),
+  /** The aria-label property is used to provide the label to any assistive
+   * technologies. This is useful if the text value is not descriptive of the
+   * button's functionality. */
+  ariaLabel: PropTypes.string,
```
smurrayatwork: Remove trailing comma.

smurrayatwork (changes requested):

react/src/components/molecules/HeaderSearch/index.js
```diff
@@ -62,7 +62,7 @@ HeaderSearch.propTypes = {
   /** The placeholder text for the input */
   placeholder: PropTypes.string,
   /** The Search button */
-  buttonSearch: PropTypes.shape(ButtonSearch.propTypes),
+  buttonSearch: PropTypes.shape(ButtonWithIcon.propTypes),
```
smurrayatwork, line 65: This prop is no longer being used in render. I don't believe that's intentional?

smurrayatwork (changes requested):

react/src/components/atoms/buttons/ButtonWithIcon/index.js
clairesunstudio: sure will update

pmallett-mc (dismissed):
Curious why the changes to className handling were reverted to all strings, but otherwise this seems good to go.

react/src/components/molecules/HeaderSearch/index.js
pmallett-mc: buttonSearch is accessed on line 49

clairesunstudio: I forgot to add the props before...

clairesunstudio:
@petermallett @smurrayatwork the reason i reverted the className handling change because for some reason the component rendered more than 1 time and each time the array gets more pushes and results in the following:
screen shot 2018-07-24 at 10 24 16 am
Not sure when it was doing that...

clairesunstudio:
@petermallett i did something to it by mistake... it's reverted

react/src/components/atoms/buttons/ButtonWithIcon/ButtonWithIcon.stories.js
```diff
@@ -37,6 +38,30 @@ storiesOf('atoms/buttons', module)
...
 
+      return(
+        <ButtonWithIcon {...props} />
+      );
+    })
+  )
+  .add(
+    'ButtonSearch',
+    withInfo(`<div>${buttonWithIconReadme}</div>`)(() => {
+      const props = {
```
avrilpearl, line 49: do you want all these available as knobs for buttonsearch - i would assume search icon etc should not be an option to play with in knobs cause this is a specific variant that needs that icon type

avrilpearl (approved):
one minor change on the search button story - but otherwise this looks good and does not cause any issues/follow up work on the search side so +++

avrilpearl (changes requested):
see comment above

avrilpearl (approved):

smurrayatwork (approved):

avrilpearl (approved):

PR #83, opened 2018-07-20
Patternlab/concatenate scss

clairesunstudio:
Combine all themes into index scss, prepping for componentized scss sharing:
To test:
-  [ ] visual regression testing to varify no change on mayflower patternlab UI
-  [ ] For patterns that had a theme, appended a `//theme` section

Question:
- Do we want to keep the gov theme and pilot override?
- Should we keep the theme separated or integrated in scss in the future?


clair0917 (approved):
looks great!

ygannett (approved):
Looks good to me!

PR #84, opened 2018-07-23
WIP react/DP-9399  add backstop WIP

pmallett-mc:

## Description
Testing React backstop/Circle config -- I may delete all this & make a clean PR some time down the road...

## Related Issue / Ticket
https://jira.mass.gov/browse/DP-9399

## Steps to Test
* `git checkout react/DP-9399--add-backstop--WIP`
* `nvm use && npm install`
* `npm run build-storybook` # the backstop tests will access the static build of the Storybook
* `npm run backstop`

## Screenshots
n/a

## Additional Notes:
n/a

* added npm run commands for backstop testing

#### @TODO
* add CHANGELOG


.circleci/config.yml
```diff
@@ -78,6 +78,23 @@ jobs:
...
+      - checkout
+      - restore_cache:
+          keys:
+            - mf-react-dependencies-{{ checksum "package.json" }}
+      - run: { name: 'NPM Install', command: 'cd react && npm install' }
+      - save_cache:
+          name: Save NPM cache
+          key: mf-react-dependencies-{{ checksum "package.json" }}
+          paths:
+            - node_modules
```
rbayliss: Each step executes in a new shell independent of the last. So (kinda crazy), but you have to `cd react` at each step.  In this case `paths` is relative to the root, so it should be `react/node_modules`.  Ugh.. I just realized that this is why our caching isn't working for the Patternlab build.

.circleci/config.yml
```diff
@@ -78,6 +78,23 @@ jobs:
...
+      - run: npm run storybook
```
rbayliss: `cd react && npm run storybook`

rbayliss: Also, the storybook script starts a server, right?  I think that'll be a problem here.  Would `build-storybook` work?

react/package.json
```diff
@@ -49,18 +52,19 @@
     "@storybook/addons": "^3.3.10",
     "@storybook/react": "^3.3.10",
     "babel-core": "^6.26.0",
+    "backstopjs": "^3.5.2",
```
rbayliss: You don't actually want to install backstop in your dev dependencies.  Much better to let it come with the docker container.  It'll install PhantomJS + SlimerJS + Chrome, which is bananas.  Running it in the docker container also has the benefit of making sure we're using a consistent version of chrome with the same fonts, etc installed.

react/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+
+services:
+  backstop:
+    image: backstopjs/backstopjs:v3.1.21
+    volumes:
+      - ./:/src
+    depends_on:
+      - web
+    shm_size: 1gb
+    entrypoint: [backstop]
```
rbayliss: Took me a sec to spot what you're doing with the configs (two different ones).  Totally up to you, but I find it easier to allow some variance within an individual config by allowing flags.  Eg: https://github.com/massgov/mass/blob/develop/backstop/backstop.js#L4-L6

Using something like that you can use the same config with flags (`--atoms` could trigger the addition of the atom patterns, etc).  Less code to manage, and cleaner commands.

react/backstop/data/engine_scripts/puppet/onReady.js
```diff
@@ -0,0 +1,6 @@
+module.exports = async (page, scenario, vp) => {
```
rbayliss: Totally up to you, but I find having scripts in there for puppeteer, Chromy, Phantom, and Slimer is overwhelming and confusing.  We're consolidating just to a single `ready.js` and a single `before.js` in `patternlab/backstop/scripts`.

react/backstop/backstop.js
```diff
@@ -0,0 +1,72 @@
...
+  scenarios,
+  paths: {
+    bitmaps_reference: `${__dirname}/data/bitmaps_reference`,
+    bitmaps_test: `${__dirname}/data/bitmaps_test`,
+    engine_scripts: `${__dirname}/engine_scripts`,
+    html_report: `${__dirname}/data/html_report`,
+    ci_report: `${__dirname}/data/ci_report`
+  },
+  report: ['browser', 'CI'],
+  engine: 'chrome',
```
rbayliss: You're getting CasperJS (Firefox) instead of Chromy/Puppeteer for these configurations. I recommend puppeteer - it's faster and more reliable than the others.  To get it, you have to bump the image in your `docker-compose.yml` to `backstopjs/backstopjs:v3.2.17`.

rbayliss (commented):
Sorry it took me a bit to get to this. I didn't actually see the path error you were hitting.  Maybe you fixed it.  In any case, my comments below. Happy to talk through this during a screenshare if you want.

.circleci/config.yml
pmallett-mc: Yea, I think this will need to be `build-storybook` & that build is what the nginx server will expose for the test job?

react/docker-compose.yml
pmallett-mc: I have them split because most of the atom screenshots only need a very small viewport & backstop doesn't let you override viewports per-scenario -- only per config.

was hoping to keep the overall runtime down. May not be worth the extra complexity/duplication?

react/backstop/data/engine_scripts/puppet/onReady.js
pmallett-mc: Agreed, I will clear these up -- none of these are running right now, howerver. They were the source of the path issue I was having. Specifically the lines:
```
  // onBeforeScript: 'puppet/onBefore.js',
  // onReadyScript: 'puppet/onReady.js',
```
in the backstop*.js config -- will try this again after I shuffle things around.

react/backstop/backstop.js
pmallett-mc: I was wondering what was up with that, thanks!

.circleci/config.yml
rbayliss: Yeah, exactly.  You want to build to static HTML, then have the nginx server serve it.

react/docker-compose.yml
pmallett-mc: Oh, I think I was missing the point here -- yes, I think I will try that ^.^ thanks again

pmallett-mc:
Closing this old PR down in favor of newer WIP branch/PR: #118

PR #85, opened 2018-07-23
React: Add Standard Accordion Components to Mayflower React

avrilpearl:
**Description:** This PR includes the addition of 4 new components that together create a standard accordion feature in Mayflower React. The design is pulled from Mayflower PL but were refactored in order to generate these new components. Also, in order to enable animations, this PR installs the react-transition-group library.

**Question to Validate with React Team** --> Is the approach to handling children in both accordion item and accordion wrapper appropriate here? Is the approach to animations here appropriate? Do we want to pull animation and focus scss in accordion item out to a higher level in the react library so it applies to all relevant components?

External Review Steps:
1. npm install
2. Go to @atoms/icons/SvgCircleChevron and confirm that the icon exist and matches [mayflower PL](https://mayflower.digital.mass.gov/?p=atoms-svg-circle-chevron) 
3. Go to @molecules/AccordionItem and confirm that a single accordion item is displayed, matching the appearance of a single item in [Mayflower PL](https://mayflower.digital.mass.gov/?p=organisms-collapsible-content-as-steps)
4. Go to @organisms/AccordionWrapper and confirm that multiple accordion items are grouped and follow the same behavior as [Mayflower PL](https://mayflower.digital.mass.gov/?p=organisms-collapsible-content-as-steps) 

Updated Gifs
accordionnoicon
accordionsecondary
accordiondefault


react/src/components/molecules/AccordionItem/index.js
```diff
@@ -0,0 +1,82 @@
...
+                {this.props.icon}
+              </div>
+            )}
+            <h2 className="ma__accordion-header__title">{this.props.title}</h2>
+          </button>
+        </header>
+        <Collapse in={this.state.open} dimension="height">
+          <div className="ma__accordion-content__body">
+            { React.Children.map(this.props.children, (child) => {
+              if (allowedChildren.includes(child.type.name)) {
```
clairesunstudio: maybe should also check for  `React.isValidElement(child))` before `child.type.name`

clairesunstudio: this could also be refactored into a utility function?

react/src/components/molecules/AccordionItem/index.js
```diff
@@ -0,0 +1,82 @@
...
+  handleClick() {
+    this.setState({
+      open: !this.state.open
+    });
+  }
+
+  render() {
+    const buttonClasses = this.state.open ? 'ma__accordion-header__button is-open' : 'ma__accordion-header__button';
+    const accordionClasses = this.props.border ? 'ma__accordion-item' : 'ma__accordion-item__borderless';
+    const allowedChildren = ['Paragraph', 'OrderedList', 'UnorderedList', 'Heading', 'Table'];
```
clairesunstudio: where does this list come from? also see this in patternlab --> https://mayflower.digital.mass.gov/?p=molecules-contact-us-collapsed-with-more-link

react/src/components/molecules/AccordionItem/index.js
```diff
@@ -0,0 +1,82 @@
...
+            aria-label={this.props.info}
+            onClick={this.handleClick}
+            aria-expanded={this.state.open}
+          >
+            { this.props.icon && (
+              <div className="ma__accordion-header__icon">
+                {this.props.icon}
+              </div>
+            )}
+            <h2 className="ma__accordion-header__title">{this.props.title}</h2>
```
clairesunstudio: heading level should probably be passable

clairesunstudio:
I think this is a good use of children to prevent prop drilling. I can see many organisms or higher components can be refactored to use children especially for those don't care what order of children are passed and don't need to carry any logic at the wrapper level. The react-transition-group can potentially solve all the animation problems we are running into, e.g. page scroll. For this specific case, since everything is purely css, maybe we could use ReactCSSTransitionGroup a higher level api to handle the animation. thoughts?

react/src/components/molecules/AccordionItem/index.js
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import SvgCircleChevron from '../../atoms/icons/SvgCircleChevron';
+import SvgChevron from '../../atoms/icons/SvgChevron';
+import Heading from '../../atoms/headings/Heading';
+import Collapse from '../../animations/Collapse';
+import './style.css';
```
pmallett-mc, line 8: I think this can change to the new method for including component styles now that #151 is merged

react/src/components/organisms/AccordionWrapper/index.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './style.css';
```
pmallett-mc, line 3: see #151 also

pmallett-mc (dismissed):
component styles changes

react/src/components/molecules/AccordionItem/index.js
avrilpearl: we can remove any check for children type if that is a better approach here..

react/src/components/molecules/AccordionItem/index.js
avrilpearl: added that to the wrapper - we can remove the child check if that is prefered here

react/src/components/molecules/AccordionItem/index.js
avrilpearl: not in here but i did update this in the style.scss file for this component

react/src/components/organisms/AccordionWrapper/index.js
avrilpearl: same comment as above

react/src/components/molecules/AccordionItem/style.scss
```diff
@@ -0,0 +1,181 @@
...
+
+      & > svg {
+        height: 25px;
+        width: 25px;
+        fill: $c-theme-secondary;
+      }
+    }
+
+    &__title {
+      @include ma-h5;
```
clairesunstudio: why import both h4 and h5 styles?

clairesunstudio:
Made some small changes to scss. The styles look good! There are a few warnings in AccordionItem. 
screen shot 2018-08-20 at 5 07 55 pm
This could be a follow up in design --> currently accordions don't have a hover state style, and for secondary accordion, it might not seem actionable.

react/src/components/molecules/AccordionItem/style.scss
clairesunstudio: if you want to include the margin, I'd recommend only import h4 and manually set margin

avrilpearl: i think i took this from a pl scss - but i will pull it out

react/backstop/backstop.js
```diff
@@ -10,15 +11,27 @@ let viewports;
 
 // Scan for component names and set up viewports.
 const componentsPath = `${__dirname}/../src/components/`;
+components = listDirs(componentsPath)
```
pmallett-mc: I think we should keep the main backstop.js as free from distraction as possible (it's already got more than we probably want with my split of atoms/non-atoms)-- it shouldn't care about how to get the list of components, that's a storybook-specific thing.

Instead of looping over the results of the other call into fs here & doing more fs access, let's rename storyBookBackstop listDirs to something like listComponents & fold this logic in there.

react/backstop/backstop.js
```diff
@@ -10,15 +11,27 @@ let viewports;
...
-  components = listDirs(componentsPath)
+  components = testList
     // Component directory names are capitalized.
     .filter((filePath) => (/^[A-Z]/.test(path.basename(filePath))))
     // Only test atoms with this backstop configuration.
     .filter((filePath) => (filePath.indexOf('/atoms/') > -1))
     // Skip table and media/Image, they need to be tested with larger viewports.
-    .filter((filePath) => ((filePath.indexOf('table') === -1) && (path.basename(filePath) !== 'Image')));
+    .filter((filePath) => ((filePath.indexOf('table') === -1) && (path.basename(filePath) !== 'Image')))
+    // Do not test components without stories;
```
pmallett-mc: Did this comment go with a line of code somewhere else?

react/src/components/animations/Collapse/index.js
```diff
@@ -0,0 +1,174 @@
...
+
+  return(
+    value +
+    parseInt(css(elem, margins[0]), 10) +
+    parseInt(css(elem, margins[1]), 10)
+  );
+}
+
+function triggerBrowserReflow(node) {
+  node.offsetHeight; // eslint-disable-line no-unused-expressions
```
pmallett-mc, line 29: Is this doing anything?

react/src/components/molecules/CalloutLink/CalloutLink.stories.js
```diff
@@ -8,7 +8,7 @@ import CalloutLink from './index';
 import CalloutLinkDocs from './CalloutLink.md';
 import calloutLinkOptions from './CalloutLink.knobs.options';
 
-storiesOf('molecules/CalloutLink', module).addDecorator(withKnobs)
```
pmallett-mc, line 11: I think these changes resulted in stories being listed in two places
doubled-stories

react/src/components/molecules/ImagePromo/ImagePromo.stories.js
```diff
@@ -34,12 +34,14 @@ const getCommonPropsWithKnobs = () => ({
   phone: null
 });
 
-storiesOf('molecules/ImagePromo', module).addDecorator(withKnobs)
```
pmallett-mc, line 37: see above comment

react/src/components/organisms/Footer/Footer.stories.js
```diff
@@ -13,7 +13,7 @@ import SocialLinksLiveData from '../../molecules/SocialLinks/SocialLinksLive.jso
 import stateSeal from '../../../../../assets/images/stateseal.png';
 
 
-storiesOf('organisms/Footer', module).addDecorator(withKnobs)
+storiesOf('organisms', module).addDecorator(withKnobs)
```
pmallett-mc, line 16: same issue as other example
double-footer

pmallett-mc (changes requested):
see code comments

react/src/components/animations/Collapse/index.js
avrilpearl: yeah this causes the redraw so the animation can happen

react/src/components/molecules/CalloutLink/CalloutLink.stories.js
avrilpearl: yeah this is so that backstop vr works correctly, it was getting the wrong stories before because of this nesting and that there is not a one to one relationship between src folder structure and storybook navigation - this was the simplest solution to solve this issue but we may want to refactor this in a followup

clairesunstudio: Could we do what we discussed about using the `add.("[ComponentName]")` in stories to create the reference url?

react/backstop/backstop.js
```diff
@@ -1,4 +1,4 @@
-const listDirs = require('./listDirs');
+const listComponents = require('./listComponents');
 const storyBookBackstop = require('./storyBookBackstop');
 const path = require('path');
 const fs = require('fs');
```
clairesunstudio: this line can be removed

clairesunstudio (approved):

pmallett-mc (approved):

react/src/components/organisms/AccordionWrapper/index.js
```diff
@@ -0,0 +1,49 @@
...
+            emphasize: props.emphasize,
+            border: props.border,
+            headerLevel: props.headerLevel
+          });
+          return clone;
+        }
+        return(
+          /* eslint-disable no-console */
+          console.log(`Warning! You cannot pass a ${child.type.name} child to AccordionWrapper`)
+          /* eslint-disable no-console */
```
pmallett-mc: I think this line was supposed to be `eslint-enable`

clairesunstudio: why should this be `enable`?

pmallett-mc: That or line 20 was maybe supposed to be `eslint-disable-next-line` instead of just disable

PR #93, opened 2018-07-26
Hotfix: Patternlab/gov theme refactor

clairesunstudio:
remove gov theme and all references


rbayliss (changes requested):
Looks awesome!  The only thing I'd request is that we also clean up the reference screenshots (the G2G ones).  Other than that, I think this is ready to merge.

rbayliss (approved):

PR #104, opened 2018-07-31
remove unused pilot theme

clairesunstudio:
clean up unused pilot theme compilation


mrossi113 (approved):
👍 LGTM! @clairesunstudio see my comment.

mrossi113:
@clairesunstudio in this PR do you want to clean up the `Go to classic Mass.gov`as well? That was from pilot.mass.gov in the header of each twig.
```
{% if not hideBackTo %}
<div class="ma__header__backto">
    <a href="http://www.mass.gov">Go to classic Mass.gov</a>
</div>
{% endif %}
```

rbayliss (approved):
Looks good to me!

clairesunstudio:
@mrossi113 i can address that in a separate pr. I think we can keep that header back to button but just not render it by passing `hideBackTo = true`. What do you think?

mrossi113:
Discussed with @clairesunstudio about the `Go to classic Mass.gov` this will be handled in another ticket. By keeping the functionality, but changing the text in the link.

PR #106, opened 2018-07-31
Patternlab: hide back to classic mass.gov button on all pages

clairesunstudio:
- hide back to button on all pages 
- make back to button configurable
To test: set "header" props in header.json
```    
"hideBackTo": false,
"backTo": {
      "text": "Mayflower Homepage",
      "url" : "https://www.mayflower.digital.mass.gov"
}
```


mrossi113 (approved):
👍 LGTM!

PR #107, opened 2018-07-31
DP-9527 set up React build

pmallett-mc:

## Description
Adding Circle builds for Mayflower /react

## Related Issue / Ticket
- [JIRA issue](https://jira.mass.gov/browse/DP-9527)

## Steps to Test
1.  react_build_storybook should pass when a branch is pushed
2. react_deploy_tag should deploy to NPM (if version is bumped) and S3 when a tag is pushed without anything after the semver (eg. "-alpha1")

## Screenshots
N/A

## Additional Notes:
N/A

* Circle build_branch and build_tag

#### @TODO

* test all added build jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,9 +91,6 @@ jobs:
           keys:
             - mfr-dependencies-{{ checksum "react/package.json" }}
       - run: cd react && npm install
-      - run:
-          name: Run tests
-          command: cd react && npm test
       - save_cache:
           paths:
             - react/node_modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,11 +7,6 @@ references:
   configure_git: &configure_git
     run: {name: 'Prepare Git', command: 'git config --global user.email "ci.bot@mass.gov" && git config --global user.name "MassGov Bot"'}
 
-react_defaults: &react_defaults
-  working_directory: ~/repo
-  docker:
-    - image: circleci/node:8.9.4
-
 jobs:
   patternlab_build:
     working_directory: /var/www/code
@@ -84,7 +79,9 @@ jobs:
       - run: { name: 'Release NPM', command: "cd patternlab/styleguide && node_modules/.bin/gulp npm:release" }
 
   react_build:
-    <<: *react_defaults
+    working_directory: ~/repo
+    docker:
+      - image: circleci/node:8.9.4
     steps:
       - checkout
       - restore_cache:
@@ -106,7 +103,9 @@ jobs:
           paths: .
 
   react_deploy_tag:
-    <<: *react_defaults
+    working_directory: ~/repo
+    docker:
+      - image: circleci/node:8.9.4
     steps:
       - attach_workspace:
           at: ~/repo
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,21 +83,22 @@ jobs:
       - run: { name: 'Release S3', command: "cd patternlab/styleguide && node_modules/.bin/gulp patternlab:release" }
       - run: { name: 'Release NPM', command: "cd patternlab/styleguide && node_modules/.bin/gulp npm:release" }
 
-  react_build:
+  react_build_storybook:
     <<: *react_defaults
     steps:
       - checkout
       - restore_cache:
           keys:
             - mfr-dependencies-{{ checksum "react/package.json" }}
+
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -103,5 +143,8 @@ workflows:
           filters:
             branches: { ignore: /.*/ }
             tags: { only: /.*/ }
-
-
+      - react_deploy_tag:
```
rbayliss: So a couple things here:
1. This job will run without a valid workspace to attach to (you have to have the `react_build_storybook` job listed as a job in this workflow too.
2. It looks like this always releases to the root of the S3 bucket.  Just a heads up we sometimes do experimental tags (eg: 6.0.0-alpha1), which would cause the bucket to get updated with something that may or may not be stable.  That's probably fine for NPM (you may have a use case for a pre-release tag), but I'd imagine you want to be more intentional with what happens in the S3 bucket.

Re: S3 bucket, maybe this is good enough for the first iteration.   It would be really great if we could sync up on the S3 publishing between React and Patternlab and handle them the same way, which probably means a script that can detect whether a given tag is stable or not and sync to a subdirectory for non-stable tags.

rbayliss (changes requested):
Comments below

.circleci/config.yml
```diff
@@ -78,6 +83,38 @@ jobs:
...
+          paths: .
+
+  react_deploy_tag:
+    <<: *react_defaults
+    steps:
+      - attach_workspace:
+          - at: ~/repo
+      - run: {name: 'Authenticate with registry', command: 'echo "//registry.npmjs.org/:_authToken=$npm_TOKEN" > ~/repo/react/.npmrc'}
+      - run: {name: 'Publish package', command: 'cd react && npm publish'}
+      - run: {name: 'Release S3', command: 'cd react && aws s3 sync storybook-static s3://mayflower-react.digital.mass.gov/ --delete'}
```
clairesunstudio: currently for mayflower react s3, we also invalidate cache on cloudfront --> `aws cloudfront create-invalidation --distribution-id [!!PROD_DISTRIBUTION_ID_HERE!!] --paths '/*'` should we also do that here?

.circleci/config.yml
pmallett-mc: Thanks @rbayliss I fixed up the build_tag jobs list in ae4b37235

Also added an ignore value to react_deploy_tag that should ignore versions with /-.*/ after the number, I think?

rbayliss: I think that regex needs an escape in it.  `-` is a special regex character.  Try `\-`

rbayliss: Also, yeah, the addition of the job looks good to me.

pmallett-mc: pedant moment: I think that's only true in a bracketed range (eg. [a-z]), but yea, escaping it to be 100% explicit doesn't hurt :) added in be3ebd9cf

rbayliss: Ha, fair enough.  You're probably right - I get nervous with the tag workflows though, since we don't see them run until we're in the middle of a release.

.circleci/config.yml
```diff
@@ -78,6 +83,39 @@ jobs:
...
+
+  react_deploy_tag:
+    <<: *react_defaults
+    steps:
+      - attach_workspace:
+          - at: ~/repo
+      - *configure_npm
+      - run: {name: 'Publish NPM package', command: 'cd react && npm publish'}
+      - run: {name: 'Sync storybook-static to S3', command: 'cd react && aws s3 sync storybook-static s3://mayflower-react.digital.mass.gov/ --delete'}
+      - run: cd react && aws configure set preview.cloudfront true
```
clairesunstudio: is `cd react` necessary?

clairesunstudio (commented):
looks good will do a pre-release to test it out. `cd react` doesn't seem necessary for the cloundfront config steps

.circleci/config.yml
pmallett-mc: No, but I thought it might be good to have all the commands consistent to (hopefully?) cut down on copy/paste errors in the future since almost every other command needs it.

clairesunstudio (approved):
tested with pre-release tag, works great!! once this is approved and merged will test with a stable release for react

rbayliss (approved):
Looks awesome!  Nice work!

PR #114, opened 2018-08-01
DP-9862: Remove all url references from Patternlab

rbayliss:

## Description

Patternlab's release workflow is greatly complicated by the fact that we have to recompile the application for each URL we release two. The reasons we had to do this were:

1. Assets (CSS, JS, etc) need to be loaded.
2. Schema.org metatags should point to themselves.

#1 is fixable by using relative paths to reference the assets (../../assets/images/foo.jpg rather than urlDomaindirectory/assets/images/foo.jpg). #2 just doesn't matter that much - it would be useful for schema.org metatag testing to have the urls pointing to the right places, but we rarely validate this, and it's certainly not worth the additional complexity it requires.

This is a prerequisite to simplifying the Patternlab build process in any way.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-9862)
- [Github issue]()

## Steps to Test

1. Visit the [S3 version](https://mayflower.digital.mass.gov/b/patternlab/DP-9862-remove-url-ref/index.html) of this branch.
2. Ensure that all images and assets load as they do on mayflower.digital.mass.gov.

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/changelogs/DP-9862.txt
+++ b/changelogs/DP-9862.txt
@@ -1,0 +1,4 @@
+___DESCRIPTION___
+Changed:
+Patch:
+- (Patternlab) DP-9862: Remove all absolute URL references from Patternlab data and templates.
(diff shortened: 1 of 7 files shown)


ygannett (approved):
LGTM.  The images are shown as same as on mayflower.digital.mass.gov.

mrossi113 (approved):
👍 LGTM!

PR #129, opened 2018-08-02
React/dp 9399  add backstop

pmallett-mc:

## Description
Adds backstop tests for Mayflower/react

## Related Issue / Ticket
https://jira.mass.gov/browse/DP-9399

## Steps to Test
* `git checkout react/DP-9399--add-backstop`
* `nvm use && npm install`
* `npm run build-storybook` # the backstop tests will access the static build of the Storybook
* `npm run backstop`

* Also tests in Circle on this PR should all pass

## Screenshots
n/a

## Additional Notes:
n/a

* added npm run commands for backstop testing in /react

#### @TODO
* add CHANGELOG


.circleci/config.yml
```diff
@@ -105,6 +108,28 @@ jobs:
...
+      - attach_workspace: {at: /home/circleci}
+      - setup_remote_docker
+      - run: |
+          cd react
+          docker-compose up --no-start web backstop
+          docker cp ./ $(docker-compose ps -q backstop):/src/
+          docker cp ./storybook-static/. $(docker-compose ps -q web):/usr/share/nginx/html
+      - run: |
+          cd react
+          docker-compose run backstop test --atoms && docker-compose run backstop test
```
rbayliss, line 123: This will fail if either of these tests fails, but splitting into two commands will cause the second report to be the one that's saved to the artifacts and reported on as JUnit.   I think it would probably be better to have an `--all` flag (or make that the default) to run them all in a single run so the report has everything.

rbayliss: This goes for `npm run backstop` as well - the report will only reflect the most recent one run.  I'd suggest finding a way to make the flags combine so `--atoms --other` will run all of the tests.   Are you sure you don't want to just combine these into a single test?  It seems like it'll make your workflow simpler - yes, it's a bit slower, but it seems worth it.

react/docs/visual-regression.md
```diff
@@ -0,0 +1,39 @@
...
+
+When you are ready to commit your work, you will want to update the reference screenshots with the new ones that reflect your changes.  To do that, run:
+
+```bash
+# Capture screenshots of the current state.
+docker-compose run backstop test --config=backstop/backstop.atoms.js
+docker-compose run backstop test --config=backstop/backstop.js
+# Accept these screenshots as the new references.
+docker-compose run backstop approve --config=backstop/backstop.atoms.js
+docker-compose run backstop approve --config=backstop/backstop.js
```
rbayliss: I think this section needs updating - backstop.atoms.js doesn't exist anymore.

react/backstop/backstop.js
```diff
@@ -0,0 +1,96 @@
...
+    html_report: `${__dirname}/data/html_report`,
+    ci_report: `${__dirname}/data/ci_report`
+  },
+  report: ['browser', 'CI'],
+  engine: 'puppeteer',
+  engineOptions: {
+    args: ['--no-sandbox']
+  },
+  asyncCaptureLimit: 5,
+  asyncCompareLimit: 50,
```
rbayliss, line 92: Just wanted to note that we've seen some issues hitting resource limits with these async comparisons.  I think in our case it has a lot to do with the pixel sizes of the pages we're capturing and comparing, so I wouldn't change them now, but keep in mind that the limits may need to drop if you start getting reports of hanging or oddly failing tests.

rbayliss (approved):
This is looking really good!  The one thing I'd suggest addressing is the reporting (explained below).  Other than that, I think you're in good shape here!

mrjmd:
Talked to @petermallett, this is still a WIP so holding off doing a full review until the work is complete, removing label for now.

react/backstop/backstop.js
pmallett-mc: Yea, I noticed they had to be lowered quite a bit for the PL build; I think our viewport sizes are small enough that this will run fine for now, though.

react/docs/visual-regression.md
pmallett-mc: Updated docs/visual-regression.md in 475c7c3f4

.circleci/config.yml
pmallett-mc: I agree simplifying this is probably desired & I will test the speed differences in the next round of updates for this. For now I think we're going to add them as-is so we can unblock work on sharing SCSS a bit.

I set it up in a single run step so that if --atoms fails, that is the report that will be written, the second test config will not be run.

Again, agreed the complexity is probably not worth the speed savings overall & will probably be cleaned up in the next update.

react/.eslintignore
```diff
@@ -1,5 +1,8 @@
-tests/
+backstop/
 es/
 lib/
+node_modules/
+storybook-static/
+tests/
```
mrjmd, line 6: We eslintignore tests on search as well, but I'm not sure if it's something we should be considering changing long term. It definitely had led to things slipping through the cracks (unused variables & imports, hanging spaces, etc). Thoughts?

react/backstop/backstop.js
```diff
@@ -0,0 +1,96 @@
...
+    .filter((filePath) => (/^[A-Z]/.test(path.basename(filePath))))
+    // Only test atoms with this backstop configuration.
+    .filter((filePath) => (filePath.indexOf('/atoms/') > -1))
+    // Skip table and media/Image, they need to be tested with larger viewports.
+    .filter((filePath) => ((filePath.indexOf('table') === -1) && (path.basename(filePath) !== 'Image')));
+
+  viewports = [{
+    label: 'small_atom',
+    width: 400,
+    height: 250
```
mrjmd, line 26: How arbitrary is the choice of viewport size here, or what motivated it?

react/backstop/scripts/onBefore.js
```diff
@@ -0,0 +1,3 @@
+module.exports = async (page, scenario, vp) => {
+  // await require('./loadCookies')(page, scenario);
```
mrjmd: I'm assuming this file is left here for future use?

mrjmd (approved):
All tests pass, however the readme instructions are a little unclear / didn't quite work for me as written. I tried to just run `npm run backstop` since it shows `npm-run-all` as the first step, however this didn't work. I then tried running `npm run build-storybook` as the prerequisite mentioned in the readme, however this failed because css assets weren't built. So to run tests my actual process was:

- cd react
- npm i
- npm run build-css
- npm run build-storybook
- npm run backstop

After that, everything worked fine.

pmallett-mc:
Temporarily re-adding the WIP label so we can test combined PL/react build/test

react/backstop/scripts/onBefore.js
pmallett-mc: Yea, though I can't imagine we will ever need to add any cookies; maybe should just remove it.

clairesunstudio (approved):
works great!

react/.eslintignore
pmallett-mc: I agree we should probably lint tests, too -- I'll note in the ticket for followup.

pmallett-mc: Also note: this is for the test that came with (I think?) nwb, not the backstop code in react/backstop

react/backstop/backstop.js
pmallett-mc: It's large enough to fit all the atoms; See DP-9399 todo's for followup.

PR #131, opened 2018-08-02
share assets 

clairesunstudio:
move sharable assets into root

- move fonts and scss partials into the assets folder in the root
- keep data, js, and images and index scss files in the assets folder under `patternlab/styleguide/source`. scss will be compiled in individual project using different compiling tools

Question: should images be moved into the root?


rbayliss (approved):
This looks good to me.  It keeps us in a working state on the PL side while moving the assets.  There are lots of little optimizations and changes we could make as a result, but I think it's probably worth moving this through as quickly as we can.

PR #140, opened 2018-08-07
DP 9871 updates for hotfix

tnathanjames:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](https://github.com/massgov/mayflower/blob/dev/docs/change-log-instructions.md)

## Description

Version and change log update for release 6.0.1

## Related Issue / Ticket

- [DP-9871](https://jira.mass.gov/browse/DP-9871)

## Steps to Test

1. 

## Screenshots
Use something like [licecap](http://www.cockos.com/licecap/) to capture gifs to demonstrate behaviors.

## Additional Notes:

Anything else to add?

#### Impacted Areas in Application

* 

#### @TODO

*

PR #133, opened 2018-08-03
DP-9871: Update pagination to replace state on page load

tnathanjames:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](https://github.com/massgov/mayflower/blob/dev/docs/change-log-instructions.md)

## Description

This updates the pagination JavaScript to replace state instead of push state when a page with pagination loads. This replace state instead of push state prevents extra entries into the history which was causing the need of multiple back button clicks to get to the previous page.

## Related Issue / Ticket

- [DP-9871](https://jira.mass.gov/browse/DP-9871)

## Steps to Test

1. Checkout the branch and build Mayflower pattern lab locally.
2. Go to a listing page like http://localhost:3000/patterns/05-pages-location-listing/05-pages-location-listing.html
3. Use the pager, click on results links and click back, and then confirm that the correct number of back button clicks gets you to the first page.
4. Pull this branch into Drupal with `composer require massgov/mayflower-artifacts:dev-patternlab/DP-9871-pagination-history`
5. Be sure to clear cache. You also may need to clear the browser cache on listing pages.
6. Test a listing page as defined in the related Jira ticket to confirm that it addresses the issue.

*Note that this does not prevent the ?page=1&_page=2 issue. That will need to be addressed separately.*

## Screenshots
Use something like [licecap](http://www.cockos.com/licecap/) to capture gifs to demonstrate behaviors.

## Additional Notes:

Anything else to add?

#### Impacted Areas in Application

* 

#### @TODO

*

PR #141, opened 2018-08-07
Sync Master back to develop for 6.0.1

mrossi113:
#133 & #140


mrossi113:
@clairesunstudio we needed a hotfix for Drupal Patternlab to get into the release today. Could you just review this for CircleCI is failing on react part and let me know if that okay?

clairesunstudio:
@mrossi113 I think it's probably failing because the versions are not bumped in the react package.json and package-lock.json. We should probably automate that in the future. But for now, versions needs to be updated in patternlab and react manually.

PR #151, opened 2018-08-09
create global scss from shared scss

clairesunstudio:
This is the first step of sharing scss in react. 

After this we can pull in component based scss via @import into component style.scss
Currently, atoms/forms.scss and atoms/_input-date.scss are pointing to shared images for background-image. The next step is to figure out how to resolve those background images paths and how to consume svg more efficiently in react 

Instructions for scss imports after this change:

create a `style.scss` for each component and add:
- `@import 'global';` (this replaces the basic variables imports: variables, colors and breakpoints)
- `@import [path-to-partial]` (e.g. `@import '01-atoms/divider'`)
-  add custom scss or overrides for react


clairesunstudio:
proposal to update green theme color for accessibility --> #154

clairesunstudio:
Need to upgrade screenshots after mayflower release to npm

pmallett-mc (approved):

PR #158, opened 2018-08-13
[React] Adds the TextArea atom to Mayflower React

avrilpearl:
## Description
Adds the TextArea form input atom to mayflower react. Need to confirm how we want to do error handling an potentially refactor this component before it is merged in.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-10023)

## Steps to Test
1. cd react directory of repo 
2. `npm install` & `npm start`
3. go to atoms/forms/TextArea

## Screenshots
textarea


pmallett-mc:
This is failing backstop tests because there is no reference image for the new atom. Instructions for updating those is here: https://github.com/massgov/mayflower/blob/develop/react/docs/visual-regression.md

clairesunstudio:
will follow up with a screenshots update after the mayflower release

react/src/components/atoms/forms/TextArea/index.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
+    if (nextProps.defaultText !== prevState.value) {
+      return{
+        value: nextProps.defaultText
+      };
+    }
+    return null;
+  }
```
pmallett-mc, line 20: We have this pattern in a lot of our components, but I wonder if we need it.
Outside of Storybook/Knobs, would this component ever need to reset its state because of a prop change?

**Anti-pattern: Erasing state when props change**
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-erasing-state-when-props-change

PR #159, opened 2018-08-13
[React] Error Message component

clairesunstudio:
error


avrilpearl:
@clairesunstudio  --> do you want to change the font color on different status?

screen shot 2018-08-13 at 2 35 47 pm

react/src/components/atoms/forms/ErrorMessage/index.js
```diff
@@ -0,0 +1,34 @@
...
+  message: PropTypes.string.isRequired,
+  /** Validation status */
+  status: PropTypes.oneOf(['error, success'])
+};
+
+ErrorMessage.defaultProps = {
+  status: 'error'
+};
+
+export default ErrorMessage;
```
avrilpearl: you need to add this to the index.js component export file...

react/src/components/atoms/forms/ErrorMessage/index.js
```diff
@@ -0,0 +1,34 @@
...
+import SvgInputSuccess from '../../icons/SvgInputSuccess';
+import './style.css';
+
+const ErrorMessage = ({inputId, message, status}) => {
+  const isSuccessful = status === 'success';
+  return(
+    <div
+      htmlFor={inputId}
+      aria-labelledby={inputId}
+      className={`ma__error-msg has-error ${isSuccessful ? 'ma__error-msg--success' : ''}`}
```
avrilpearl, line 15: should probably add something in style.scss for rendering the message text (font) green in a success state.

clairesunstudio:
The styling didn't work because mayflower-react is still pulling from mayflower package css and the styles were competing. Added temporary overrides to make it work now and will address that after the share scss pr gets in

smurrayatwork (changes requested):
The ErrorMessage component has a bunch of lint things that need fixing.

react/src/components/atoms/forms/ErrorMessage/index.js
```diff
@@ -0,0 +1,34 @@
...
+  )
+};
+
+ErrorMessage.propTypes = {
+  /** The ID of the corresponding input field */
+  inputId: PropTypes.string.isRequired,
+  /** The help text for the corresponding input field */
+  message: PropTypes.string.isRequired,
+  /** Validation status */
+  status: PropTypes.oneOf(['error, success'])
```
smurrayatwork: I think these should be two different strings?

react/src/components/atoms/forms/ErrorMessage/index.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import SvgInputError from '../../icons/SvgInputError';
+import SvgInputSuccess from '../../icons/SvgInputSuccess';
+import './style.css';
+
+const ErrorMessage = ({inputId, message, status}) => {
```
smurrayatwork: Should the inputId actually be the id attribute for the div being rendered?

smurrayatwork (changes requested):

react/src/components/atoms/forms/ErrorMessage/index.js
clairesunstudio: no that should be the id for `<input>` that the error message is tied to

react/src/components/atoms/forms/ErrorMessage/index.js
clairesunstudio: good catch

clairesunstudio: fixed this and the lint errors --> a77ab0d

smurrayatwork (approved):

react/src/components/atoms/forms/ErrorMessage/index.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import SvgInputError from '../../icons/SvgInputError';
+import SvgInputSuccess from '../../icons/SvgInputSuccess';
+import './style.css';
+
+const ErrorMessage = ({ inputId, message, status }) => {
```
avrilpearl: is there a reason you called the out here vs. as just props? - we are not doing this anywhere else so curious...

avrilpearl (approved):

PR #174, opened 2018-08-17
DP-10102: Revert "DP-9431: add sticky TOC to link list page in mayflower"

tnathanjames:
This reverts commit 9172376e3d977f277f14d7c64432961b503cb06a.

## Description

The change being reverted caused problems on the Drupal side with the sticky table of contents.

## Related Issue / Ticket

- [DP-10102](https://jira.mass.gov/browse/DP-10102)

## Steps to Test

1. 

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


patternlab/styleguide/source/assets/js/modules/stickyTOC.js
```diff
@@ -3,7 +3,7 @@ export default (function (window,document,$,undefined) {
   $('.pre-content .ma__sticky-toc').each(function() {
     const $toc = $('.ma__sticky-toc'),
           $tocContent = $('.ma__sticky-toc__links'),
-          $tocSections = $('.pre-content').next().find('.ma__comp-heading'),
+          $tocSections = $('.ma__information-details .page-content, .ma__stacked-row-section__container .ma__stacked-row-section__title').find('h2'),
```
ygannett, line 6: @tnathanjames For info details page, if `.ma__rich-text` under `.page-content` contains `h2`, the JS would pick it up with this code.  Here is the sample I created on feature5:  http://massgovfeature5.prod.acquia-sites.com/info-details/how-where-to-recycle

how___where_to_recycle___mass_gov

It's not supposed to have h2, but since we cannot control what authors put in RTE, it would be safe to exclude the `.ma__rich-text`.

Curated list wouldn't be the case since `<h2>` is in the separate container `.ma__stacked-row-section__title` from its associated body content `.main-content`.

ygannett (changes requested):
Please see my comment below.

mrossi113:
@tnathanjames could you update the react version for the following files`react/package.json` and `react/package-lock.json` needs to be updated to `6.2.3` as well for this hotfix to pass correctly after merging into master branch. Sorry the documentation for the release process has not been updated yet.

patternlab/styleguide/source/assets/js/modules/stickyTOC.js
tnathanjames: @ygannett Since this is something that is broken in production and since this PR is just reverting a previous change, we don't want to put additional changes into this. We don't want h3's in the content, but it could be seen as a feature instead of a bug to allow content authors to add additional headings to the TOC. I think we should proceed with this and we can come back for exclusion of h2's in rich text if we decide we want to later.

mrossi113 (approved):
👍 LGTM!

patternlab/styleguide/source/assets/js/modules/stickyTOC.js
ygannett: ok.

PR #179, opened 2018-08-20
React/dp 10098 remove mayflower npm dependency

clairesunstudio:
___DESCRIPTION___
Changed:
Minor:
- (React) DP-10098: Consume scss and images directly from shared assets.
          Removed: Removed Mayflower npm dependency in mayflower react.
          Added: Added `bourbon`, `normalize-scss` and `bourbon-neat` as react dev dependencies.

___POST DEPLOY STEPS___
`$npm install` in `react/`

For future reference:
Use variable `$assets-path` in scss to point to assets relative path

Follow up:
Import component level scss directly into the component js, and remove unused scss.
Export high level scss variables ?


rbayliss (approved):
Makes sense to me.  Not sure why the visual differences are there, but they seem trivial. You might want to check your `mismatchThreshold` if you're getting unexplainable diffs.

mrjmd (approved):
Redacting my previous comment as I realized it made no sense.

This is a 👍 from me.

PR #188, opened 2018-08-24
React/import componentize css

clairesunstudio:


react/src/components/atoms/Divider/style.scss
```diff
@@ -0,0 +1,29 @@
+@import 'global';
+
+.ma__divider {
```
clairesunstudio: @mark-casias you should be able to directly import the scss instead of copying all the styles into each component style.scss e.g. `@import '01-atoms/divider';`

clairesunstudio: keep @import 'global' to import variables for compiling, and @import directly from the assets/scss following example: `@import '01-atoms/error-msg';` Let me know if you have any questions!

mark-casias: I thought the goal of this ticket was to get the styles directly inside the components instead of depending on an external source. I actually removed the `@import '01-atoms/divider';` code in the first commit I did.

clairesunstudio: So once the styles are imported, the component will have the rendered styles as part of the exported component. This is just telling it to import and compile during the build

PR #193, opened 2018-08-27
DP-10086: SVG Icon Proof of Concept.

smurrayatwork:

## Description
A new component has been created, Icon, that replaces all the previously used SVG components. This allows us to store all SVG files in one directory within the component directory and pass props to modify that SVG's rendering. It makes use of [svg-sprite-loader](https://github.com/kisenka/svg-sprite-loader) to generate a SVG sprite on the fly that contains only the SVG assets that are needed on the page at the time.

All previously existing SVG components have been removed and replaced with Icon, so testing will need to cover quite a lot of components here.

svg-sprite-loader strips off the original attributes from the original svg element. This means that in some cases, you'll need to pass the svg's width and height to its svgWidth and svgHeight props. We could also make a new prop for CSS classes in the future (or now if needed).

Once this PR is reviewed and accepted, a new PR on massgov search will be needed to handle the changing of components.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-10086)

## Steps to Test

1. Pull the branch and run `npm install`.
2. Start up storybook (`npm run start`) and verify that there is now an Icon component under atoms/icons.
3. Click to the Icon component. Using knobs, verify that each SVG icon renders properly.
4. For each SVG icon, find where it exists in other components and verify that everything looks the same as before.

## Screenshots

screen shot 2018-08-27 at 11 09 04 am

screen shot 2018-08-27 at 11 09 26 am

screen shot 2018-08-27 at 1 06 03 pm


clairesunstudio:
can we leverage something like this --> https://www.npmjs.com/package/svg-to-jsx so that we can directly use svgs in the shared assets?

clairesunstudio:
screen shot 2018-08-30 at 10 04 08 am

clairesunstudio (approved):
Looks good! Like the icon screenshot! accessibility can happen in a follow up

react/nwb.config.js
```diff
@@ -11,5 +13,34 @@ module.exports = {
   },
   babel: {
     stage: 1
+  },
+  webpack: {
```
mrjmd, line 17: Since we know this will only work with the umd published version of Mayflower React, what are the implications at this point for consuming this in search?

mrjmd (approved):
Approved, with one big question about integration into search.

react/nwb.config.js
mrjmd: More broadly I'm wondering if this nwb.config.js is even necessary since we're not using the umd version anywhere.

PR #232, opened 2018-09-10
Refactor Color Variables 

avrilpearl:
## Description
This PR updates the naming of our global/base color variables. It focuses on aligning our documentation with the variable naming. Further, it creates and refactors a defined grayscale. Finally, it attempts to name variables based on usage.

color-swatches-pl-update


clairesunstudio (approved):
please add a changelog

PR #244, opened 2018-09-13
DP-8086: Mayflower React: Port General Teaser Molecule from Mass.Gov Search.

smurrayatwork:
## Description
This adds the following new components: GeneralTeaser, IconLink, Link. This also modifies the Image component to add a new prop for classnames: classes.

Backstop configuration has been updated to support creating a custom override for component stories.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-8086)

## Steps to Test

1. Pull the branch and run `npm install`.
2. Go to http://localhost:6006/. Click on Organisms -> GeneralTeaser. Verify that the GeneralTeaser's look/feel is okay and all knobs work.
3. Repeat Step two with IconLink and Link (under Molecules).

## Screenshots


react/backstop/backstop.js
```diff
@@ -27,10 +27,20 @@ let scenarios = storyBookBackstop.mapComponents(testComponents, debug);
...
-  if ((result.label.indexOf('Error') > -1) || (result.label.indexOf('NarrowTemplate') > -1)) {
-    result.delay = 5000;
-  }
+  const delays = [
+    'Error',
+    'NarrowTemplate',
+    'GeneralTeaser'
+  ];
+// eslint-disable-next-line prefer-const
+  let result = { ...item };
```
pmallett-mc: This does not need to be let or spread because we don't care if scenarios is mutated, it's immediately reassigned to the value returned by map (a new array).

react/backstop/backstop.js
```diff
@@ -27,10 +27,20 @@ let scenarios = storyBookBackstop.mapComponents(testComponents, debug);
...
+  delays.some((value) => {
```
pmallett-mc: This seems like a strange use of .some(), the result of the call is not being used. Clearer as:
```
if (delays.indexOf(item.label) != -1) {
  result.delay = 5000;
}
```
(which matches the code you wrote for overrides below: https://github.com/massgov/mayflower/pull/244/files#diff-05cb2343d70bab35e91a87095c1b419eR103)

or, if we really want to use some(), it would be:
```
if (delays.some((value) => item.label.indexOf(value) > -1)) {
  result.delay = 5000;
}
```

react/backstop/backstop.js
```diff
@@ -27,10 +27,20 @@ let scenarios = storyBookBackstop.mapComponents(testComponents, debug);
 // The Error and example Template pages need a delay to allow the background
 // animation to complete.
```
pmallett-mc, line 28: Comment should be updated for why GeneralTeaser was added

mark-casias (approved):
Tested locally successfully and code looks good. Approved

react/src/components/atoms/media/Image/index.js
```diff
@@ -1,28 +1,30 @@
...
-    className={`ma__image${shape && ` ${shape}`}`}
-    alt={alt}
-    src={src}
-    height={height}
-    width={width}
-  />
-);
+const Image = (props) => {
+  const { classes, shape, ...imgProps } = props;
+  if (shape && shape.length > 0) {
```
pmallett-mc, line 6: shape is removed from the story, but still used here

react/src/components/atoms/media/Image/index.js
```diff
@@ -1,28 +1,30 @@
...
 Image.propTypes = {
-  shape: PropTypes.string,
-  src: PropTypes.string,
+  classes: PropTypes.arrayOf(PropTypes.string),
+  src: PropTypes.string.isRequired,
   alt: PropTypes.string,
   width: PropTypes.number,
-  height: PropTypes.number
+  height: PropTypes.number,
+  shape: PropTypes.string
```
pmallett-mc, line 19: shape is removed from the story, but still used here

react/src/components/atoms/media/Image/index.js
```diff
@@ -1,28 +1,30 @@
...
+};
+
+Image.defaultProps = {
+  classes: [
+    'ma__image'
+  ],
+  alt: '',
+  shape: ''
```
pmallett-mc, line 27: shape is removed from the story, but still used here

react/src/components/molecules/CalloutLink/CalloutLink.stories.js
```diff
@@ -8,7 +8,7 @@ import CalloutLink from './index';
 import CalloutLinkDocs from './CalloutLink.md';
 import calloutLinkOptions from './CalloutLink.knobs.options';
 
-storiesOf('molecules/CalloutLink', module).addDecorator(withKnobs)
```
pmallett-mc: Was this move discussed? these were previously grouped as variations of the same molecule, now just bare in the molecules listing:
screen shot 2018-09-13 at 16 17 48

react/src/components/molecules/ImagePromo/ImagePromo.stories.js
```diff
@@ -34,7 +34,7 @@ const getCommonPropsWithKnobs = () => ({
   phone: null
 });
 
-storiesOf('molecules/ImagePromo', module).addDecorator(withKnobs)
```
pmallett-mc: Same as Q: for molecules/CalloutLink, these were grouped as two variations of the ImagePromo component

react/src/components/molecules/Link/index.js
```diff
@@ -0,0 +1,27 @@
...
+    title={props.info}
+  >{props.text}&nbsp;<SvgArrow />
+  </a>
+);
+
+Link.propTypes = {
+  href: PropTypes.string,
+  info: PropTypes.string,
+  text: PropTypes.string
+};
```
pmallett-mc, line 19: href & text should maybe be required instead of having defaults?

react/src/components/organisms/Footer/Footer.stories.js
```diff
@@ -13,7 +13,7 @@ import SocialLinksLiveData from '../../molecules/SocialLinks/SocialLinksLive.jso
 import stateSeal from '../../../../../assets/images/stateseal.png';
 
 
-storiesOf('organisms/Footer', module).addDecorator(withKnobs)
```
pmallett-mc: same as above, the Footer stories were previously grouped under organisms, now cluttering up the listing:
screen shot 2018-09-13 at 16 21 37

react/src/components/organisms/GeneralTeaser/index.js
```diff
@@ -0,0 +1,252 @@
...
+        if (typeof child.type === 'string') {
+          valid = child.type === 'IconLink';
+        }
+        valid = child.type.name && child.type.name === 'IconLink';
+        return valid;
+      });
+      return valid;
+    };
+    if (component && !isValid(component)) {
+      return new Error(`Invalid prop ${propName} supplied to ${componentName}. Got: ${component.type.name}. Validation failed.`);
```
pmallett-mc, line 240: consider `throw` for these (Error is intended as an exception object)
removes the need for the `eslint-disable-next-line consistent-return` disables, too

pmallett-mc (changes requested):
see comments

react/src/components/organisms/GeneralTeaser/index.js
smurrayatwork: I think doing it this way way recommended by React docs instead of `throw`

react/backstop/backstop.js
smurrayatwork: Not doing this gives another eslint error for changing function variable properties.

react/backstop/backstop.js
smurrayatwork: Removed; I had thought I needed the delay but I didn't for GeneralTeaser.

clairesunstudio (approved):

react/backstop/backstop.js
pmallett-mc: It's right down there on line 33 https://github.com/massgov/mayflower/pull/244/files#diff-8de69294a9eac4b07724e8185a481d62R33

PR #254, opened 2018-09-18
Update mf/react to work with shared assets

pmallett-mc:

## Description
Updates mf/react to use the shared assets in the project root.
During builds, assets are copied into src/assets so that they can be bundled in /es/assets, then they are cleaned up so that they only exist at /assets to avoid confusion.

The UMD build (unused) is removed, it was not working with path aliases

When `npm start` is run, a symbolic link is created instead because cleanup cannot be done reliably. This also obviates the need for a special assets-watch update during local dev.

## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1. check out code, run all `npm run` commands
1. everything should work as expected:
** npm run build should result in a /es/ build where paths resolve to /es/assets
** npm start should work because of the /src/assets symbolic link
** SharedAssets alias will work during npm start and storybook builds because of the resolver in react/.storybook/webpack.config.js 
** npm lint will work because of the path alias resolver in react/.eslintrc.json
** in case src/assets is accidentally left in place during dev, it is git ignored so duplicated assets will not be committed
** SharedAssets alias is handled during npm build by the setup in react/nwb.config.js

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


react/package.json
```diff
@@ -59,13 +60,17 @@
     "acorn-object-rest-spread": "^1.1.0",
     "acorn-static-class-property-initializer": "^1.0.0",
     "babel-core": "^6.26.0",
+    "babel-plugin-module-resolver": "^3.1.1",
+    "babel-plugin-transform-import-paths": "^1.0.0",
     "bourbon": "^5.1.0",
```
clairesunstudio: looks good! The only thing I would say is move `bourbon` `bourbon-neat` and `normalize-scss` into build dependencies, so when we compile css in search or EEC we don't need to install them separately

react/scripts/assets-copy.js
```diff
@@ -0,0 +1,13 @@
+const fse = require('fs-extra');
+
+fse.remove('src/assets', (err1) => {
```
clairesunstudio: and based on our convo sounds like we don't need the remove step here

smurrayatwork (approved):

clairesunstudio:
screen shot 2018-09-19 at 10 24 15 am
still getting this error ^ we can look into this later

clairesunstudio:
Symlink method with `--copy-files` is probably not going to work for windows system. We can verify that and follow up when necessary

clairesunstudio (approved):
Otherwise looks great! exciting stuff! thanks @petermallett

PR #260, opened 2018-09-21
remove view box values from header svgs to fix IE bug

clair0917:
## Description
Hot fix to fix IE page scroll

IE ignores the width and height elements of svgs that have viewbox values set and are not wrapped in  tags. By removing the viewbox values from the svgs used in the header (utility nav and search button) the extra page scrolling is fixed.
Other browsers respect the width and height values as set, so removing the viewbox attributes produces no visible changes.  

The overflow visible attribute on the utility nav content adds scrollbars in IE. This is actually expected behavior for the browser, but I think it's not what the designers had in mind, so I used the -ms attribute to remove the extra scrollbars there as well.


rbayliss:
I'm still seeing breakage:
1. I can't scroll (the overflow property is still on the body).
2. The space is still there when I disable that property.

I'm not saying you're not on the right track, but we haven't gotten there yet.

rbayliss:
Also, I should note that for testing, I'm visiting https://mayflower.digital.mass.gov/b/hotfix/header-svg-fix/patterns/05-pages-Homepage/05-pages-Homepage.html?1537551691869 in IE11.

PR #263, opened 2018-09-21
Loosen the child validation in AccordionWrapper so we can import in EEC

pmallett-mc:

## Description

Removes the element name check from AccordionWrapper child validation.

During the gatsby build for EEC, the component names are mangled and not checkable.

## Related Issue / Ticket
N/A

## Steps to Test

1. npm run build
1. in EEC npm link ../mayflower/react
1. npm run build && npm run serve
** the build should not fail due to AccordionWrapper prop validation.

## Screenshots
N/A

## Additional Notes:
N/A

* EEC AccordionWrapper build

#### @TODO
N/A

### DIFF
--- a/react/src/components/organisms/AccordionWrapper/index.js
+++ b/react/src/components/organisms/AccordionWrapper/index.js
@@ -7,7 +7,7 @@ const AccordionWrapper = (props) => {
   return(
     <div className="ma__accordion-wrapper">
       { React.Children.map(children, (child) => {
-        if (React.isValidElement(child) && child.type.name === 'AccordionItem') {
+        if (React.isValidElement(child)) {
           const clone = React.cloneElement(child, {
             secondary: props.secondary,
             emphasize: props.emphasize,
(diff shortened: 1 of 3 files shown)


mrjmd (approved):

react/src/components/organisms/AccordionWrapper/index.js
```diff
@@ -17,8 +17,8 @@ const AccordionWrapper = (props) => {
           return clone;
         }
         return(
-          /* eslint-disable no-console */
-          console.log(`Warning! You cannot pass a ${child.type.name} child to AccordionWrapper`)
+          /* eslint-disable-next-line no-console */
+          console.warn('Warning! You can only pass an AccordionItem child to AccordionWrapper.')
```
clairesunstudio, line 21: this is only checking if there's a child passed and but not actually checking if it's an AccordionItem, is that the intention?

clairesunstudio: @petermallett why do we need to pass non accordionItem into accordionWrapper in EEC?

pmallett-mc: We don't need to pass anything else, in the description of this PR it says why the change was made: 
> Removes the element name check from AccordionWrapper child validation.
> 
> During the gatsby build for EEC, the component names are mangled and not checkable.

trying to check for a component name fails during the EEC build because the name ends up being something like 'd' or similar.

clairesunstudio (approved):
tested in eec and this pr fixes the hard error, will need to follow up and add back the validation, see IconLink

PR #264, opened 2018-09-21
React/configureable site logo

clairesunstudio:


react/src/components/atoms/media/SiteLogo/index.js
```diff
@@ -17,13 +19,24 @@ SiteLogo.propTypes = {
...
 SiteLogo.defaultProps = {
   url: {
     domain: '/'
-  }
+  },
+  image: {
+    src: logo,
+    alt: 'Massachusetts state seal',
+    width: 45,
+    height: 45
```
pmallett-mc, line 36: @clairesunstudio the image w/h props are affecting the img tag attributes, but due to the `.ma__site-logo a img` properties, they do not have a visual affect; is that ok for now? (marking approved, otherwise).

pmallett-mc (approved):
approved w/ minor comment: https://github.com/massgov/mayflower/pull/264/files#r219858504

react/src/components/atoms/media/SiteLogo/index.js
clairesunstudio: thanks for noting that, will follow up on it. I'm trying get a quick on in without touching the patternlab side

avrilpearl (approved):

PR #275, opened 2018-09-27
React/accordion wrapper child validation

pmallett-mc:
## Description
Fix the react/AccordionWrapper child PropType to allow a single element, or array of elements.
This should remove the need for the Fragment in EEC.

## Related Issue / Ticket
N/A

## Steps to Test
1.  npm run start
1. load http://localhost:6006/
1. Navigate to organisms/AccordionWrapper
1. You should not see PropType warnings about the `children` prop

## Screenshots
N/A

## Additional Notes:
N/A

* mf/react, Search, EEC

#### @TODO
N/A

### DIFF
--- a/changelogs/AccordionWrapper-PropType.txt
+++ b/changelogs/AccordionWrapper-PropType.txt
@@ -1,0 +1,4 @@
+___DESCRIPTION___
+Fixed
+Minor
+- (React) Fixed AccordionWrapper child PropType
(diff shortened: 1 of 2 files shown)


smurrayatwork (approved):

clairesunstudio (approved):
This pr only fixes : `Failed prop type: Invalid prop `children` of type `array` supplied to `AccordionWrapper`, expected a single ReactElement.` and it's not only allowing accordionItem in the proptype check

PR #276, opened 2018-09-27
fix selectbox bug

avrilpearl:
Fixes bug relating to selectbox atom in mayflower react

### DIFF
--- a/react/src/components/atoms/forms/SelectBox/index.js
+++ b/react/src/components/atoms/forms/SelectBox/index.js
@@ -50,7 +50,8 @@ class SelectBox extends React.Component {
       const matchedValue = matchedItem && matchedItem.value;
       return matchedValue;
     };
-    const selectedValue = getValueByText(this.props.options, selected) ? getValueByText(this.props.options, selected) : this.props.options[0].value;
+    const valueInOptions = getValueByText(this.props.options, selected);
+    const selectedValue = valueInOptions || this.props.options[0].value;
     return(
       <section className={classNames}>
         <label htmlFor={this.props.id} className={labelClassNames}>{this.props.label}</label>
@@ -70,7 +71,7 @@ class SelectBox extends React.Component {
             ))}
           </select>
           <div className="ma__select-box__link">
-            <span className="js-dropdown-link">{selected}</span>
+            <span className="js-dropdown-link">{valueInOptions ? selected : this.props.options[0].text}</span>
             <span className="ma__select-box__icon" />
           </div>
         </div>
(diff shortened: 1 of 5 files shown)


react/src/components/atoms/forms/SelectBox/index.js
```diff
@@ -46,7 +46,7 @@ class SelectBox extends React.Component {
     const labelClassNames = stackLabel ? 'ma__select-box__label' : 'ma__label--inline ma__label--small';
     const selectBoxInline = stackLabel ? '' : 'ma__select-box__field--inline';
     const selectedIndex = this.props.options.findIndex((option) => option.text === selected);
```
clairesunstudio: this still needs work, if the value selected from props is not in this.props.options, selectedIndex is undefined and component will break

react/src/components/atoms/forms/SelectBox/index.js
```diff
@@ -45,8 +45,11 @@ class SelectBox extends React.Component {
     const { stackLabel } = this.props;
     const labelClassNames = stackLabel ? 'ma__select-box__label' : 'ma__label--inline ma__label--small';
     const selectBoxInline = stackLabel ? '' : 'ma__select-box__field--inline';
-    const selectedIndex = this.props.options.findIndex((option) => option.text === selected);
-
+    const selectedValue = this.props.options.map((option) => {
+      if (option.text === selected) {
+        return option.value;
+      } return this.props.options[0].value;
+    });
```
pmallett-mc: This assigns an array to `selectedValue` (one with as many elements as this.props.options has). Probably not what was intended.

pmallett-mc (dismissed):
This isn't quite right -- see the example here https://reactjs.org/docs/forms.html#the-select-tag; the component state should be tracking the select's value, not the text that is displayed. Then we don't have to look up into the options list to find ids or keys, only check whether the value exists or not for validation.

avrilpearl:
@pmallett-mc  --> for the selectbox, the text was the state we originally started tracking and are dependent on in eec and massgov-search. I get that at some point we should switch this over to value but that is not what this pr is trying to resolve so that can be a follow up pr that will require work in massgov-search and eec. this pr is just trying to fix a bug that is blocking eec work.

clairesunstudio:
this will fix situations when there's no match, the text rendered and value are kept in sync:
e.g.

screen shot 2018-09-28 at 3 01 18 pm
screen shot 2018-09-28 at 3 00 00 pm

clairesunstudio (approved):

PR #277, opened 2018-09-27
React/filter box refactor

clairesunstudio:
- (React) DP-1234: Refactor FilterBox to allow any form components 
Note: This will need to be followed up in search and other apps consuming FilterBox / SearchBanner

This is a step towards making searchBanner with postInputFilter responsive on mobile


react/src/components/organisms/FilterBox/index.js
```diff
@@ -25,27 +23,11 @@ const FilterBox = (props) => {
         <form className={`ma__filter-box__form js-filter-box ${isActive}`} action={action}>
           <div className="main-content--two">
             <div className="ma__filter-box__filters">
-              {organization && (
-                <div className="ma__filter-box__organizations">
-                  <InputTextTypeAhead {...organization} />
+              { fields.map((field) => (
+                <div className={field.class}>
```
pmallett-mc: these will need a `key` attribute now

pmallett-mc (changes requested):
This looks pretty good, but the FilterBox is failing to render in Storybook when I test this in conjunction w/ the latest develop:
filterbox-errors

Also, though SearchBanner functions, there are a lot of console warnings, mostly "Warning: Encountered two children with the same key"

I also think this should be tested out with Search & we should have a PR to go in with it, after all, if nothing can use this, there's no reason to rush into it.

clairesunstudio:
@petermallett good catch! updated and fixed filterbox. This is part of the search banner work I'm doing for mobile responsiveness for EEC. I separated this out so that it's easier to review. I created a pr in search just so we don't forget, but it won't work until https://github.com/massgov/massgov-search/pull/326 and https://github.com/massgov/massgov-search/pull/327 get in so that we can bring in the new mayflower. We should spend some time next week to get those resolved

clairesunstudio:
But for now, @petermallett can you review this again so that i can continue working on search banner styles for EEC.  created a test branch here -->https://github.com/massgov/eec-childcare-provider-search/pull/125 thanks!

react/src/components/utilities/componentPropTypeCheck.js
```diff
@@ -6,9 +9,73 @@ const componentPropTypeCheck = (props, propName, componentName, componentString)
...
+      return false;
+    }
+    return comp.every((child) => {
+      if (typeof child.class !== 'string') {
+        return false;
+      }
+      if (!child.component) {
+        return false;
+      }
+      if (typeof child.component.type === 'string') {
```
clairesunstudio: do you mean `child.component.type.name` here?

react/src/components/utilities/componentPropTypeCheck.js
```diff
@@ -6,9 +9,73 @@ const componentPropTypeCheck = (props, propName, componentName, componentString)
...
+        return componentTypes.indexOf(child.component.type) > -1;
+      }
+      return child.component.type.name &&
```
clairesunstudio, line 73: this seems repetitive to the one above

react/src/components/utilities/componentPropTypeCheck.js
```diff
@@ -6,9 +9,73 @@ const componentPropTypeCheck = (props, propName, componentName, componentString)
...
+          componentTypes.indexOf(child.component.type.name) > -1;
+    });
+  };
+  if (!component || (component && !isValid(component))) {
+    return new Error(`Invalid prop ${propName} supplied to ${componentName}. Got: ${component || 'undefined'}. Validation failed.`);
```
clairesunstudio, line 78: the error message doesn't return the component name correctly

clairesunstudio:
error message example:
screen shot 2018-09-28 at 11 19 55 am

PR #287, opened 2018-10-03
DP-10244: Mayflower React: Import Componentized SCSS (atoms)

mark-casias:
## Description
This PR moves all atom CSS into the component itself instead of depending on the mass import. Uses assets directory. Future PR's for this ticket will update other portions of mayflower.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-10244)

## Steps to Test

1.  There should be no visual regression in this.


clairesunstudio:
looks good so far, one thing I would say for testing purposes, we probably want to remove all the remaining atoms imports in index.scss

mark-casias:
@clairesunstudio Updated the PR as asked. The only atom scss I didn't remove was being used only in the contact group molecule. I didn't want to update the molecules in this PR so I left the import in place.

clairesunstudio:
tested in search: https://github.com/massgov/massgov-search/pull/333 looks good!

react/src/components/atoms/lists/OrderedList/style.scss
```diff
@@ -0,0 +1,2 @@
+@import 'global';
+@import '01-atoms/global/elements';
```
clairesunstudio: seems like elements should be treated different, if the project import only the components that don't have global/elements it will not have basic body html styles. I would say for now keep it in index.scss and into the base

clairesunstudio:
markie, i don't see forms.scss being imported anywhere, I wonder how is that working?

react/src/components/atoms/media/Image/style.scss
```diff
@@ -0,0 +1,3 @@
+@import 'global';
+@import '01-atoms/global/elements';
```
clairesunstudio: this can be removed

react/src/components/atoms/text/Paragraph/style.scss
```diff
@@ -0,0 +1,2 @@
+@import 'global';
+@import '01-atoms/global/elements';
```
clairesunstudio: this can be removed

react/src/components/atoms/headings/Heading/style.scss
```diff
@@ -0,0 +1,2 @@
+@import 'global';
+@import '01-atoms/global/elements';
```
clairesunstudio: this can be removed

clairesunstudio:
forms, elements and content links are left over in atoms because they are used in multiple other components and we need to componentize those separately. 
- content link is being used in molecule contact group 
- forms contains styles for input, fieldset, select...
- elements contains html, body styles as well as styles for Image, Paragraph, Heading,...

It's difficult to test each individual components as the styles are not scoped. we might run into issues where styles are missing but we will do it as it goes.

clairesunstudio (approved):

PR #288, opened 2018-10-03
React: Add suggestions as passable prop to header search

avrilpearl:
Enables consumers to pass a suggestion component to header search if the want to render suggestions.


clairesunstudio:
can you add a story for this?

clairesunstudio:
I think we should make the input passable with options InputText or InputTextTypeahead

clairesunstudio (approved):

PR #298, opened 2018-10-09
DP-10649:  TabContainer and Tab components.

smurrayatwork:

## Description
This adds two new components: TabContainer and Tab. See the TabContainer story in how to use it.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-10649)

## Steps to Test

1. Pull the branch, start storybook `npm run start`.
2. View the TabContainer story under Organisms/TabContainer.

## Screenshots
screencapture-localhost-6006-2018-10-09-16_33_25

## Additional Notes:
This probably needs better knobs, but right now this will have to do. The nested knob doesn't work dynamically; you need to click on the parent tab for things to refresh. This only happens in storybook.


react/src/components/organisms/TabContainer/style.scss
```diff
@@ -0,0 +1,98 @@
...
+    flex-wrap: nowrap;
+    overflow-x: auto;
+    -webkit-overflow-scrolling: touch;
+    -ms-overflow-style: -ms-autohiding-scrollbar;
+    .ma__tab-title, %ma__tab-title {
+      border-color: $c-gray-light;
+      outline-color: $c-gray-light;
+      cursor: pointer;
+      padding: 1.5rem 1.5rem 0.8rem 1.5rem;
+      font-size: 1.25rem;
```
clairesunstudio: I will recommend setting a `border-top: 5px solid transparent` here to prevents the small jump

react/src/components/organisms/TabContainer/style.scss
```diff
@@ -0,0 +1,98 @@
...
+  }
+  .ma__tab-container-body, %ma__tab-container-body {
+    padding: 1rem;
+    border: 1px solid $c-gray-light;
+    border-top-color: transparent;
+  }
+  .ma__tab-container--nested {
+    @extend %ma__tab-container;
+    .ma__tab-container-head {
+      height: 70px;
```
clairesunstudio: same here

react/src/components/organisms/TabContainer/index.js
```diff
@@ -0,0 +1,74 @@
...
+    this.state = {
+      activeTab: null,
+      activeContent: null,
+      // eslint-disable-next-line react/no-unused-state
+      setActiveTab: this.setActiveTab,
+      nested: this.props.nested
+    };
+  }
+
+  static getDerivedStateFromProps(props, state) {
```
clairesunstudio: storybook knob doesn't seem to be working, is the purpose of this function for making storybook work?

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -0,0 +1,47 @@
...
+    );
+  }
+}
+
+Tab.defaultProps = {
+  default: false
+};
+
+Tab.propTypes = {
+  default: PropTypes.bool,
```
clairesunstudio, line 56: can you add descriptions to the proptypes? same to the container?

react/src/components/organisms/TabContainer/style.scss
```diff
@@ -0,0 +1,98 @@
...
+      border-bottom: 1px solid $c-gray-light;
+      flex-wrap: nowrap;
+      overflow-x: auto;
+      -webkit-overflow-scrolling: touch;
+      -ms-overflow-style: -ms-autohiding-scrollbar;
+      .ma__tab-title {
+        @extend %ma__tab-title;
+        border: none;
```
clairesunstudio: this also needs some border to avoid jumps

react/src/components/organisms/TabContainer/index.js
smurrayatwork: The purpose of that is to keep this.state.nested updated if the nested prop changes, as it's only initialized in the constructor.

clairesunstudio: ya it's initialized with a prop value, which you are not setting the state anywhere else. i'm just wondering if it's necessary to check for prop updates

react/src/components/organisms/TabContainer/index.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  render() {
+    const classes = classNames({
+      'ma__tab-container': true,
+      'ma__tab-container--nested': (this.state.nested === true)
+    });
+    // eslint-disable-next-line react/prop-types
+    const { children, defaultTab } = this.props;
+    const active = !defaultTab ? 0 : defaultTab;
```
clairesunstudio: this check doesn't seem necessary since its set in default

react/src/components/organisms/TabContainer/style.scss
```diff
@@ -0,0 +1,98 @@
...
+    -ms-overflow-style: -ms-autohiding-scrollbar;
+    .ma__tab-title, %ma__tab-title {
+      border-color: $c-gray-light;
+      outline-color: $c-gray-light;
+      cursor: pointer;
+      padding: 1.5rem 1.5rem 0.8rem 1.5rem;
+      font-size: 1.25rem;
+    }
+    .ma__tab-title:focus, .ma__tab-title--active:focus {
+      outline: none !important;
```
clairesunstudio: all the !important overrides don't seem necessary, is there a reason for these?

react/src/components/organisms/TabContainer/style.scss
```diff
@@ -0,0 +1,98 @@
...
+      border-color: $c-gray-light !important;
+      border-top-color: $c-primary !important;
+      outline-color: $c-gray-light !important;
```
clairesunstudio: outline color set but outline never applied?

clairesunstudio: Another thought i have here is that instead of border we could use outline so that you don't have to set transparent borders

react/src/components/organisms/TabContainer/index.js
```diff
@@ -0,0 +1,66 @@
...
+      activeContent: null,
+      // eslint-disable-next-line react/no-unused-state
+      setActiveTab: this.setActiveTab
+    };
+  }
+
+  render() {
+    const classes = classNames({
+      'ma__tab-container': true,
+      'ma__tab-container--nested': (this.props.nested === true)
```
avrilpearl: this can be simplified to just `'ma__tab-container--nested': this.props.nested`

mark-casias (changes requested):
Seeing some responsive issues

image

mark-casias (approved):
responsive issues have been addressed. Approved.

PR #300, opened 2018-10-10
React/refactor classnames of atoms

avrilpearl:
Description:
This pr refactors how we handle the logic of classnames using the classnames library for atoms components in mayflower react.

Ticket: https://jira.mass.gov/browse/DP-10750


react/src/components/atoms/forms/InputText/index.js
```diff
@@ -24,30 +25,29 @@ class InputText extends React.Component {
...
-        inputLabelClasses.push('ma__label--hidden');
-      }
-    }
-    const inputClasses = ['ma__input'];
-    if (inputText.required) {
-      inputClasses.push('js-is-required');
-    }
+    const inputLabelClasses = classNames({
+      ma__label: true,
+      'ma__label--required': inputText.required,
```
smurrayatwork, line 30: This line should also depend on if inputText.labelText exists.

react/src/components/atoms/forms/InputText/index.js
```diff
@@ -24,30 +25,29 @@ class InputText extends React.Component {
...
+      'ma__label--optional': !inputText.required,
```
smurrayatwork, line 31: Same here and the next line on inputText.labelText exists.

react/src/components/atoms/forms/SelectBox/index.js
```diff
@@ -34,17 +35,27 @@ class SelectBox extends React.Component {
...
-      classNames = this.props.className;
-    } else {
-      classNames = !this.props.required ? 'ma__select-box js-dropdown ma__select-box--optional' : 'ma__select-box js-dropdown';
-    }
-    const selectClassNames = this.props.required ? 'ma__select-box__select js-dropdown-select js-required' : 'ma__select-box__select js-dropdown-select';
     const { selected } = this.state;
     const { stackLabel } = this.props;
-    const labelClassNames = stackLabel ? 'ma__select-box__label' : 'ma__label--inline ma__label--small';
-    const selectBoxInline = stackLabel ? '' : 'ma__select-box__field--inline';
+    const sectionClasses = classNames({
```
smurrayatwork, line 40: The old logic seems to be that if this.props.className exists, then only use that for the section classes.

This should probably be wrapped in an if/else so that if this.props.className exists, sectionClasses is assigned to be this.props.className.

mark-casias (approved):

smurrayatwork (changes requested):

PR #313, opened 2018-10-12
DP-10784: Mayflower-React: Port Listing Table in React from PL

mark-casias:
## Description
Creates ListingTable component to replicate PL.

## Related Issue / Ticket
- [JIRA issue](https://jira.mass.gov/browse/DP-10784)

## Steps to Test

1. Checkout branch and run `npm run start`
1. Component is in Molecules -> ListingTable

## Screenshots
image


changelogs/DP-10784.txt
```diff
@@ -0,0 +1,4 @@
+___DESCRIPTION___
+Added
+Major
```
avrilpearl: i would classify this as minor impact

assets/scss/02-molecules/_listing-table.scss
```diff
@@ -70,6 +70,9 @@
 
   &__extra {
     display: none;
+    &.expanded {
```
avrilpearl, line 92: you also need to add a transition animation the the `__extra` block in order for the collapse to animate.
you should be able to use what is in accordion-items scss

```
&.collapsed {
  visibility: hidden;
  height: 0px;
}

&.collapsing, &.expanding {
  position: relative;
  overflow: hidden;
  visibility: visible;
  -webkit-transition-property: height, visibility;
  transition-property: height, visibility;
  transition-duration: 0.35s;
  -webkit-transition-duration: 0.35s;
  transition-timing-function: ease;
  -webkit-transition-timing-function: ease;
}

&.expanded {
  height: auto;
  visibility: visible;
  display: block;
}
```

changelogs/DP-10784.txt
mark-casias: updated

assets/scss/02-molecules/_listing-table.scss
mark-casias: Updated

react/src/components/molecules/ListingTable/index.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  handleClick() {
+    this.setState({
+      open: !this.state.open
+    });
+  }
+
+  render() {
+    const row = this.props.row;
```
avrilpearl: can simplify this to `const {row} = this.props;`

react/src/components/molecules/ListingTable/index.js
```diff
@@ -0,0 +1,95 @@
...
+    const visibleItems = row.visibleItems || 2;
```
avrilpearl, line 25: can simplify this to `const { visibleItems } = row.visibleItems || 2;`

mark-casias: This doesn't work because sometimes row.visibleItems is not set, so it can't be pulled out in this manner.

react/src/components/molecules/ListingTable/index.js
mark-casias: 👍

react/src/components/molecules/ListingTable/index.js
```diff
@@ -0,0 +1,95 @@
...
+class ListingTableItem extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      open: false
+    };
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick() {
```
mrjmd: I'm a big fan of using arrow functions for things like this so you don't have to bind in the constructor.

react/src/components/molecules/ListingTable/index.js
```diff
@@ -0,0 +1,95 @@
...
+    const visibleItems = row.visibleItems || 2;
+    const inlineAccordion = (row.items.length > visibleItems);
+    const rowClasses = ClassNames({
+      'ma__rich-text': true,
+      'js-accordion': inlineAccordion
+    });
+    const shownItems = row.items.slice(0, visibleItems);
+    const shownItemsContent = shownItems.map((item) => (<span className="ma__listing-table__data-item">{item}</span>));
+    const invisibleItems = (inlineAccordion) ? row.items.slice(visibleItems) : [];
+    const invisibleItemsContent = (invisibleItems.length) ? (
```
mrjmd: Instead of the ternary where the else returns an empty string, I would use the common React pattern of `const invisibleItemsContent = (invisibleItems.length) && [...]`

react/src/components/molecules/ListingTable/index.js
```diff
@@ -0,0 +1,95 @@
...
+      <Collapse in={this.state.open} dimension="height">
+        <div className="ma__listing-table__extra collapsed">
+          {invisibleItems.map((item) => (<span className="ma__listing-table__data-item">{item}</span>))}
+        </div>
+      </Collapse>
+    ) : '';
+    const invisibleMore = (inlineAccordion) ? (
```
mrjmd: Same as above, let's not use ternaries where the else returns an empty string but instead use the `&&` pattern.

mrjmd (changes requested):
A couple of small improvements, looks good and works great.

react/src/components/molecules/ListingTable/index.js
mark-casias: 👍

react/src/components/molecules/ListingTable/index.js
mark-casias: 👍

react/src/components/molecules/ListingTable/index.js
mark-casias: Completed this, however it is throwing an error in the lint.

avrilpearl (approved):

mrjmd (approved):
Looks good!

PR #327, opened 2018-10-17
React/filterbox extension

clairesunstudio:
This pr allows filterbox to conditionally render the controls section

https://jira.mass.gov/browse/DP-11000

npm link to budget repo https://github.com/massgov/budget/pull/6
appropriation-filter-prototype

### DIFF
--- a/react/src/components/atoms/forms/InputTextTypeAhead/InputTextTypeAhead.stories.js
+++ b/react/src/components/atoms/forms/InputTextTypeAhead/InputTextTypeAhead.stories.js
@@ -24,7 +24,8 @@ storiesOf('atoms/forms', module).addDecorator(withKnobs)
           inputOptions.options.orgSelector.map((option) => option.text),
           ''
         ),
-        onChange: action('InputTextTypeAhead onChange')
+        onChange: action('InputTextTypeAhead onChange'),
+        disabled: boolean('InputTextTypeAhead.disabled', false)
       };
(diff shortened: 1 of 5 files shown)


react/src/components/atoms/forms/InputTextTypeAhead/InputTextTypeAhead.stories.js
```diff
@@ -24,7 +24,8 @@ storiesOf('atoms/forms', module).addDecorator(withKnobs)
           inputOptions.options.orgSelector.map((option) => option.text),
           ''
         ),
-        onChange: action('InputTextTypeAhead onChange')
+        onChange: action('InputTextTypeAhead onChange'),
+        disabled: boolean('InputTextTypeAhead.disabled', false)
```
mrjmd, line 28: It's a bit nit-picky and certainly not a merge blocker, but it might make more sense to use `enabled` as your boolean than `disabled`, when true === enabled and false === disabled, it's easier to not have to think about later.

mrjmd (approved):
This looks good, made one small comment about disabled vs enabled as your boolean here that may or may not be worth fixing.

react/src/components/atoms/forms/InputTextTypeAhead/InputTextTypeAhead.stories.js
avrilpearl: agreed

avrilpearl (approved):

PR #330, opened 2018-10-19
(React) DP-10895: Fix for pageflipper bug

avrilpearl:
React: This pr addresses a bug in the pageflipper organism that occurs if the intro props (label and dec link) are not passed to pageflipper.

### DIFF
--- a/react/src/components/organisms/PageFlipper/index.js
+++ b/react/src/components/organisms/PageFlipper/index.js
@@ -4,20 +4,18 @@ import componentPropTypeCheck from '../../utilities/componentPropTypeCheck';
 import './style.css';
 
 const PageFlipper = (props) => {
-  const introLabel = (props.intro.label) ? (<span className="ma__page-flipper__context-label">{props.intro.label}</span>) : '';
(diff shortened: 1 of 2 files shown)


react/src/components/organisms/PageFlipper/index.js
```diff
@@ -4,20 +4,18 @@ import componentPropTypeCheck from '../../utilities/componentPropTypeCheck';
...
   const blank = (<div className="ma__page-flipper__blank">&nbsp;</div>);
   const prev = props.previousLink || blank;
   const next = props.nextLink || blank;
   return(
     <React.Fragment>
-      {introLink}
+      { props.intro && (
+        <div className="ma__page-flipper__context">
+          { props.intro.label && <span className="ma__page-flipper__context-label">{props.intro.label}</span> }
+          { props.intro.introDecorativeLink && props.intro.introDecorativeLink }
```
mrjmd: I think the first check here should still be `props.intro.introDecorativeLink.props.text` as is in develop, so that if the text is removed the entire decorative link does not get rendered.

mrjmd (changes requested):
Need to update the introDecorativeLink check, otherwise works great.

mrjmd (approved):
Looks good!

PR #331, opened 2018-10-22
DP-10853: Port Table of Contents Overlay from PL in React.

smurrayatwork:
## Description
Instead of directly porting the Table of Contents Overlay from patternlab, this PR creates a new component, Overlay, which can handle placing any content into an overlay. The story then shows the Table of Contents being placed into the Overlay component. The Knob for Overlay.linkChoice near the bottom of the knobs toggles displaying the title as a Link component or plain text. When plain text, the Overlay.title displays for changing text.

There's a new technique being used here for the title prop: render props. Title is an arrow function which is then called within the Overlay component to render the title. This allows for the title to not have to be created until it's called by the Overlay component. If a component (such as Link) is within the arrow function, it won't take up any memory until the function is called, and said memory for the component will be freed up when Overlay is unmounted.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-10853)

## Steps to Test
1. Pull the branch, check storybook under Organisms -> Overlay.

### Closed Overlay:
screencapture-localhost-6006-2018-10-22-16_00_55
### Open Overlay:
screencapture-localhost-6006-2018-10-22-16_01_14

#### @TODO
* This has some general styling work that needs to be done for colors.


react/src/components/organisms/Overlay/index.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+  toggleContent = () => {
+    this.setState((currentState) => {
+      return {
+        showContent: !currentState.showContent
+      };
+    }, () => {
+      // Set focus onto overlay-title after state is set.
+      if (this.state.showContent) {
+        document.getElementsByClassName('js-inline-overlay-title')[0].focus();
```
mrjmd, line 18: Gatsby build shouldn't choke on this `document` ref since this is wrapped in a runtime method, but I'm wondering if this selector you're using is right, since it's always going to set focus to the first title?

mrjmd (approved):
This is really neat. Only tested in storybook but everything seems to work fine.

PR #353, opened 2018-11-07
React/DP-10868  a11y tabs container focus group

ygannett:

## Description

Set up keyboard navigation between the tabs and their associated tab content with arrow keys.

In a tab group:
1. LEFT/RIGHT ARROW keys to navigate in the tabs within either the parent group or the child (= nested) group.
2. DOWN ARROW or ENTER key to navigate from the active tab to its tab content. (note: the focus is on the tab content container, not the first focusable element in the container.)

In a tab content container:
3. UP ARROW key or shift+TAB to move the focus from the tab content container to the active tab.
In tab content container, TAB key moves the keyboard focus to the closest focusable element in the container. 

Arrow keys are used only to navigate between the tabs and the active tab and its content container.

## Related Issue / Ticket

- [DP-10868](https://jira.mass.gov/browse/DP-10868)

## Steps to Test

1. Go to tab container in organisms.
2. Use LEFT/RIGHT ARROW keys to navigate between the tabs.
     - No errors at the first and last tabs by trying to navigate to non-existing tabs. It stays at the current tab.
     - The current keyboard focus matches the active tab except for when the component is rendered first time.
3. Use DOWN ARROW or ENTER key to navigate to the active tab's content area.
4. Use UP ARROW or Shift+Tab to navigate from the content area to its associate tab (= current active tab).

- The keyboard focus stays with the active tab when navigating in the tab group.
- When navigating to the tab content container from the active tab, the keyboard focus moves to the tab content container.  You see it as blue outline on the container as it gets a keyboard focus.
- The visual indicator of the active tab remains while its content area or contents of the content area has the keyboard focus. 

Navigation patterns between the active tab and its tab content:
1. active tab -> tab container -> element in tab container -> ... -> tab container -> active tab
2. parent active tab -> parent tab container -> child(= nested) active tab -> child tab container -> ... -> child(= nested) active tab -> parent tab container -> parent active tab

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


react/src/components/organisms/TabContainer/index.js
```diff
@@ -34,16 +63,36 @@ class TabContainer extends React.Component {
...
       return child;
     });
+
     return(
       <TabContext.Provider value={this.state}>
         <div className={classes}>
-          <div className="ma__tab-container-head">{this.childrenWithProps}
-          </div>
-          {this.state.activeTab && <div className="ma__tab-container-body">{this.state.activeContent}</div>}
+          <span id={this.props.a11yID} className="ma__visually-hidden">Use left and right arrows to navigate between tabs, up and down arrows to navigate between active tab and its content.</span>
```
smurrayatwork, line 93: This may be better placed in the Tab component render vs here, so each tab has the ability to define its description.

react/src/components/organisms/TabContainer/index.js
```diff
@@ -34,16 +63,36 @@ class TabContainer extends React.Component {
...
+
+
+
+// ADD onKeyDown to move the focus back to its paired tab. --> onKeyDown={this.handleTabContainerKeyDown}
+
+          {this.state.activeTab && <div className="ma__tab-container-body"
+                                    aria-labelledby={this.state.activeTab}
+                                    id={`tabpanel-${this.state.activeTab}`}
+                                    tabIndex="-1"
+                                    role="tabpanel"
```
smurrayatwork, line 102: I don't think I agree with this being a tab panel. The container body may or may not be another listing of tabs, but it depends on how the container is being used. Maybe a better role that's more generic fits here?

react/src/components/organisms/TabContainer/index.js
```diff
@@ -34,16 +63,36 @@ class TabContainer extends React.Component {
...
+          </ul>
+
+
+
+// ADD onKeyDown to move the focus back to its paired tab. --> onKeyDown={this.handleTabContainerKeyDown}
+
+          {this.state.activeTab && <div className="ma__tab-container-body"
+                                    aria-labelledby={this.state.activeTab}
+                                    id={`tabpanel-${this.state.activeTab}`}
+                                    tabIndex="-1"
```
smurrayatwork, line 101: I don't think this is needed on the container body.

react/src/components/organisms/TabContainer/index.js
```diff
@@ -34,16 +63,36 @@ class TabContainer extends React.Component {
...
+                                    role="tabpanel"
+
+
+                                    onKeyDown={this.handleTabContainerKeyDown}
```
smurrayatwork: This may take over keypresses for things within the tab container body...not sure this is a good place for setting this.

react/src/components/organisms/TabContainer/index.js
```diff
@@ -34,16 +63,36 @@ class TabContainer extends React.Component {
     const active = defaultTab;
     this.childrenWithProps = React.Children.map(children, (child, index) => {
```
smurrayatwork, line 71: The index here can be used to set the correct tabindex per tab. If you make a new property below on the cloneElement line, you could pass the index and then pull it out of tab's props.

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -31,15 +35,53 @@ class Tab extends React.Component {
...
+            <li role="presentation" className={tabClasses}>
+              <button
+                role="tab"
+                id={this.tabIdent}
+                onClick={(e) => {
+                  this.scrollIntoView(e.target);
+                  setActiveTab(this.tabIdent, this.props.children)
+                }}
+                aria-selected={active}
+                tabIndex={!active ? -1 : 0}
```
smurrayatwork: Instead of 0, if the index from TabContainer's cloneElement loop can be passed, the 0 can instead be set to that index value.

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -31,15 +35,53 @@ class Tab extends React.Component {
...
+                aria-describedby={this.props.a11yID}
+
+
+    // THIS EVENT IS ONLY AVAILABLE TO THE ONFOCUS TAB.
+                onKeyDown = {(e) => {
```
smurrayatwork:
```suggestion
                onKeyDown = (e) => {
```

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -31,15 +35,53 @@ class Tab extends React.Component {
...
+                  }
+                  if (e.keyCode === 39) {// RIGHT
+                    console.log('RIGHT');
+                  }
+                  if (e.keyCode === 40) {// DOWN
+                    console.log('DOWN');
+                  }
+                  console.log('button: ' + this.tabIdent);
+                  console.log("focus at: " + document.activeElement.id);
+                }}
```
smurrayatwork:
```suggestion
                }
```

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -31,15 +35,53 @@ class Tab extends React.Component {
...
+                aria-describedby={this.props.a11yID}
+
+
+    // THIS EVENT IS ONLY AVAILABLE TO THE ONFOCUS TAB.
+                onKeyDown = {(e) => {
+                  if (e.keyCode === 37 || e.keyCode === 39) {
+                    e.preventDefault();
+                    e.stopPropagation();
+
+                    if (this.tabIdent !== document.activeElement.id) {
```
smurrayatwork: You can actually use context for this. This has also already been set to the variable active above:
```suggestion
                    if (!active) {
```

react/src/components/organisms/TabContainer/tab.js
smurrayatwork: For functions of this size, it's better to define an arrow function which you could then set this to. Something like:
`onKeyDown = {this.handleKeyDown}`

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -31,15 +35,53 @@ class Tab extends React.Component {
...
+                      console.log('Set the active tab!');
+                      setActiveTab(document.activeElement.id, this.props.children);
```
smurrayatwork: this.tabIdent is already set to the same value of document.activeElement.id:
```suggestion
                      setActiveTab(this.tabIdent, this.props.children);
```

react/src/components/organisms/TabContainer/index.js
ygannett: This is a common practice to set up tabs:  https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/tabs/tabs-2/tabs.html

aria roles are pre-defined:  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques
So, even it sounds unfitting, we cannot change it.

> ARIA tabs consist of a role="tablist" element, which has some some role="tab" elements. These tabs are then linked to role="tabpanel" elements that contain the corresponding content for each tab.
https://simplyaccessible.com/article/danger-aria-tabs/

react/src/components/organisms/TabContainer/index.js
ygannett: I haven't set this method yet, but it's for moving the focus to its paired tab (= active tab).
As long as users land on the tab body container (with focus), they are sure they are on the tab body and can decide where they go from there.  That's why I set up this way. You might be seeing what I'm not aware of...
 
I'm looking a flow like this:
1. On [active tab]: DOWN ARROW key -> move the focus to its paired tab body container.
2. On [tab body container]: TAB/Shift + TAB and other keys to navigate the content within.
3. On [tab body container]: UP ARROW key -> move the focus to the active tab

react/src/components/organisms/TabContainer/tab.js
ygannett: I tried to remove the curly braces, but got this error:
storybook

react/src/components/organisms/TabContainer/tab.js
ygannett: With `!active`, the active tab doesn't always match the tab with focus :-(
banners_and_alerts_and_storybook
Am I missing something?

The reason I set it up this way is that I noticed an active tab in the tab component wouldn't have focus after navigating the tabs with the arrow keys, which is managed by focus-group module.

react/src/components/organisms/TabContainer/tab.js
ygannett: Changed to `this.props.tabIdent`.

react/src/components/organisms/TabContainer/tab.js
ygannett: The function is defined outside.

PR #367, opened 2018-11-29
Patternlab/DP-10447  ie11 excess page scrolling

clair0917:
## Description

In IE11 the site had extra space to the right and bottom of the view window. For all pages, SVGs were updated to keep them from overflowing their containers, on the homepage the styles were adjusted to keep the banner credits from pushing the page down.

## Related Issue / Ticket

- [DP-10447](https://jira.mass.gov/browse/DP-10447)

## Steps to Test

1. Visit the page in IE11 on Win10. Test that none of the pages have extra space to the right or bottom


assets/scss/03-organisms/_search-banner.scss
```diff
@@ -194,20 +203,21 @@
 
   .ma__banner-credit {
     @include ma-container;
-    position: absolute;
-      bottom: 20px;
-      left: 50%;
-    transform: translateX(-50%);
-    width: 100%;
+    // position: absolute;
```
tnathanjames: Shouldn't these lines just be removed instead of commented?

clair0917: yes, yes they should.

tnathanjames (approved):
LGTM

PR #386, opened 2018-12-26
DP-6375: A11y - incorrect use of aria expand/collapse tagging on topic pages content cards

clair0917:
## Description
Removes 'aria-expanded' from section link titles in desktop on topic pages

## Related Issue / Ticket

- [https://jira.mass.gov/browse/DP-6375](DP-6375)

## Steps to Test

1. Visit https://mayflower.digital.mass.gov/b/patternlab/DP-6375--correct-use-of-aria-expand-collapse/patterns/05-pages-topic/05-pages-topic.html and inspect the section titles in the cards. 
Test that the aria-expanded attribute is present and working in mobile, but not present when the accordion disappears. 

## Screenshots
screen_shot_2018-12-26_at_2_05_58_pm


ygannett (changes requested):
In addition to **aria-expanded**,  **aria-controls** also needs to be removed from the desktop view.

PR #393, opened 2019-01-03
[React] DP-11682: a11y Decorative Link Atom File Icon

avrilpearl:
## Description
Fixes issues related to the decorative link icon where there is no semantic connection between the PDF icon and the link. By including the icon with the link, in the a tag, screenreader users now know that is a pdf, doc, etc..

## Related Issue / Ticket
- [JIRA issue DP-11682](https://jira.mass.gov/browse/DP-11682)


ygannett (approved):
Looks great!  The markup is generated as expected:  The file icons are included in the link and their titles tell each file type.

smurrayatwork (approved):

PR #401, opened 2019-01-10
DP-11437 MF Location search by city/zip doesn't work if you click a map marker first

samseide:
## Description
Updating JS to fix autocomplete error when searching city/zip.

## Related Issue / Ticket

- [https://jira.mass.gov/browse/DP-11437]

## Steps to Test
See the steps outlined for recreating this issue in: https://jira.mass.gov/browse/DP-11437
Note: I was only able to get this error to be reproduced for me by using IE11 inside browserstack

### DIFF
--- a/changelogs/DP-11437.txt
+++ b/changelogs/DP-11437.txt
@@ -1,0 +1,4 @@
+___DESCRIPTION___
+Fixed
+Patch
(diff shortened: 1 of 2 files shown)


clair0917 (approved):
LGTM

PR #407, opened 2019-01-15
React/DP-11683: [a11y] Title not rendered in Decorative Link Atom

avrilpearl:
## Description
Set a condition in the decorative link atom to only add title when its value is available. Hides arrow in decorative link so not read by the screen reader.

## Related Issue / Ticket

- [DP-11683](https://jira.mass.gov/browse/DP-11683)
- [DP-11684](https://jira.mass.gov/browse/DP-11684)

### DIFF
--- a/changelogs/DP-11683.txt
+++ b/changelogs/DP-11683.txt
@@ -1,0 +1,4 @@
+___DESCRIPTION___
+Fixed
+Patch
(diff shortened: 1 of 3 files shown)


ygannett (approved):
Looks great.
title is rendered only when it has value.

smurrayatwork (approved):

PR #423, opened 2019-01-23
[React] Allow tooltip to render inline

clairesunstudio:


clairesunstudio:
We are switching to use DTA Connect helptip styles for handling inline tooltips

PR #432, opened 2019-01-25
Dp-12338:  Add inline help tip organism.

avrilpearl:
___DESCRIPTION___
Added (see below)
Minor (see below)
- (React) DP-12338: Added a help tip organism. #432
- (React) DP-12338: Added a close icon. #432

helptip-desktop
helptip-mobile


clairesunstudio (commented):
Can you make helpText the default of no children that have options to change its background-color: e.g. white, tint($c-primary, 90%)... so that we can render it in a call out alert.

react/src/components/organisms/HelpTip/index.js
```diff
@@ -0,0 +1,142 @@
...
+    const helpTextClasses = classNames({
+      'ma__help-tip__text': !children,
+      'ma__help-tip__text--children': children
+    });
+
+    return(
+      <span className={baseClass} id={id}>
+        <span className="ma__help-tip__label" id={labelId}>
+          {this.buildDangerouslyIfHasMarkup(this.props.textBefore, hasMarkup)}
+          {/* can't use a button b/c disrupts fieldsets when used inside of legend */}
```
clairesunstudio, line 75: what do you mean by legend?

clairesunstudio:
hey @avrilpearl, looks like theres a gap on mobile:
screen shot 2019-01-28 at 5 22 28 pm

instead of making the background solid colors, can you keep the mobile tray the same background as the on desktop? e.g.
desktop:
screen shot 2019-01-28 at 5 21 33 pm
mobile:
screen shot 2019-01-28 at 5 21 25 pm
If you could align the button to the right and be able to pass a close text would be great, sorry...

and for the purpose of accessibility, can you keep the text color the same --> #141414
screen shot 2019-01-28 at 5 16 17 pm

clairesunstudio (approved):

PR #436, opened 2019-01-29
Dp 12413 add input percentage

clairesunstudio:


react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -85,19 +86,24 @@ const NumberInput = (props) => (
           return(
             <div className="ma__input-number">
               <input {...inputAttr} />
+              {
+                <span className="ma__input-number-unit">{props.unit}</span>
```
```suggestion
                props.unit && <span className="ma__input-number-unit">{props.unit}</span>
```

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -85,19 +86,24 @@ const NumberInput = (props) => (
           return(
             <div className="ma__input-number">
               <input {...inputAttr} />
+              {
+                props.unit && <span className="ma__input-number-unit">{props.unit}</span>
```
avrilpearl: i would add a condition to not render this if the value is null, i.e. when the placeholder is being rendered

react/src/components/utilities/componentPropTypeCheck.js
```diff
@@ -80,3 +80,10 @@ export const validateFilters = (
 };
 
 export default componentPropTypeCheck;
+
+export const singleCharacterPropTypeCheck = (props, propName) => {
+  const prop = props[propName];
+  if (typeof prop !== 'string' || prop.length > 1) {
+    return new Error(`${propName} needs to be a single character string.`);
```
avrilpearl: i would make this more flexible, at least 2 length, e.g. if someone wanted to use other units such as kg, cm, km, ft, yd

avrilpearl (changes requested):
see comments above

avrilpearl (approved):
LGTM

PR #441, opened 2019-01-30
Dp 12451/input responsiveness

clairesunstudio:
(React) DP-12451: Added an inline option for Input and fixed responsiveness. Allow Input labelText to render elements other than a string. #441 

## example of line style and responsiveness:
screen shot 2019-01-30 at 4 29 33 pm
screen shot 2019-01-30 at 4 29 22 pm

## example passing in component as input label:
screen shot 2019-01-30 at 4 28 07 pm


react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -177,6 +180,8 @@ InputNumber.propTypes = {
   min: PropTypes.number,
   /** Using the up/down arrow keys will increment/decrement the input value by this number. */
   step: PropTypes.number,
+  /** Inline label and input field */
+  inline: PropTypes.bool
```
```suggestion
  inline: PropTypes.bool,
```

avrilpearl (approved):

PR #449, opened 2019-01-31
update state on defaultValue prop change

clairesunstudio:
This pr addresses the following issues:
- allow InputNumber to update on defaultValue prop change
- fix disabled styling of InputNumber with unit on disabled
- make up/down buttons optional
- add option to skip the slider

### DIFF
--- a/react/src/components/atoms/forms/InputNumber/index.js
+++ b/react/src/components/atoms/forms/InputNumber/index.js
@@ -19,7 +19,7 @@ const NumberInput = (props) => (
     <InputContext.Consumer>
       {
         (context) => {
-          const hasValue = context.value && context.value !== 'NaN';
+          const hasValue = context.value || context.value === 0;
 
           const inputClasses = classNames({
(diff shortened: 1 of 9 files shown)


react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -19,9 +19,18 @@ const NumberInput = (props) => (
     <InputContext.Consumer>
       {
         (context) => {
+          const hasValue = context.value && context.value !== 'NaN';
```
smurrayatwork: You can't do what's currently there because context.value can be a number.
```suggestion
              let stringValue;
              if (typeof context.value !== 'string') {
                stringValue = String(context.value);
              } else {
                stringValue = context.value;
              }
              const numberValue = Number(stringValue);
              const hasValue = (!Number.isNaN(numberValue) && (stringValue.length > 0));
```

smurrayatwork (approved):

PR #455, opened 2019-01-31
React/add option to display value

clairesunstudio:
screen shot 2019-01-31 at 4 20 32 pm


react/src/components/atoms/forms/CompoundSlider/index.js
```diff
@@ -34,14 +34,15 @@ const Handle = (props) => {
...
-    divProps.tabIndex = -1;
-  }
   return(
     <button className="ma__slider-handle" {...divProps}>
-      <div className="ma__slider-handle-value">
-        {roundedValue}
-      </div>
+      { props.displayValueFormat && (
+        <div className="ma__slider-handle-value">
+          { props.displayValueFormat === 'percentage' ? `${roundedValue * 100}%` : roundedValue }
```
smurrayatwork: The pre-rounded value should be multiplied by 100, then it should be rounded to show a percentage. Otherwise you'll end up with a giant float number here.

react/src/components/atoms/forms/CompoundSlider/index.js
```diff
@@ -34,14 +40,15 @@ const Handle = (props) => {
...
-    divProps.tabIndex = -1;
-  }
   return(
     <button className="ma__slider-handle" {...divProps}>
-      <div className="ma__slider-handle-value">
-        {roundedValue}
-      </div>
+      { props.displayValueFormat && (
+        <div className="ma__slider-handle-value">
+          { props.displayValueFormat === 'percentage' ? `${roundValue(step, value * 100, 0)}%` : roundedValue }
```
smurrayatwork: Isn't value a string here? Might be easier to use numbro, as it handles percents and rounding. Example:
`numbro(value).format({
    output: "percent",
    mantissa: 0
});`
http://numbrojs.com/format.html#percentages

avrilpearl (approved):

react/src/components/atoms/forms/CompoundSlider/index.js
```diff
@@ -1,20 +1,27 @@
...
     if (Math.floor(x) === x) return 0;
     return String(x).split('.')[1].length || 0;
   };
   const decimalPlaces = countDecimals(step);
-  const roundedValue = (Number.isInteger(step)) ? value : Number.parseFloat(value).toFixed(decimalPlaces);
+  const roundValue = (stp, val, dp) => {
+    if (Number.isInteger(stp)) {
+      return val;
+    }
+    return Number.parseFloat(val).toFixed(dp);
```
smurrayatwork, line 22:
```suggestion
    return numbro(val).format({ mantissa: dp });
```

smurrayatwork (approved):

PR #482, opened 2019-02-19
[React] hotfix: Input Number Callback Value

clairesunstudio:
Using context.value in callback causing a delay in the value update:
- fixed value delay in callback
- fixed synthetic event returning nulls
- set default value to 0 for up/down button with null default value

before:
input-number-value

after:
input-number-value

You can tested in storybook or npm link in https://github.com/massgov/calculators/pull/30


react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -59,27 +60,31 @@ const NumberInput = (props) => (
...
 
           const handleChange = (e) => {
+            e.persist();
             const { value } = e.target;
             const floatValue = Number(Number.parseFloat(value).toFixed(decimalPlaces));
             const updateError = displayErrorMessage(value, props.min, props.max, props.required);
             context.updateState({ value: floatValue, ...updateError }, () => {
               if (typeof props.onChange === 'function') {
-                props.onChange(e, context.value, props.id);
+                props.onChange(e, value, props.id);
```
smurrayatwork: Shouldn't this be floatValue?

```suggestion
                props.onChange(e, floatValue, props.id);
```

smurrayatwork (approved):

PR #484, opened 2019-02-21
React/dp 12807 hotfix

clairesunstudio:
Fixed
Patch
- (React) DP-12807: Prevent InputCurrency returning NaN when default value is set to null #484
screen shot 2019-02-21 at 4 09 45 pm

- (React) DP-12806: Fix error message inline styling #484
Before:
screen shot 2019-02-21 at 4 07 48 pm
after:
screen shot 2019-02-21 at 4 07 51 pm

### DIFF
--- a/react/src/components/atoms/forms/InputCurrency/index.js
+++ b/react/src/components/atoms/forms/InputCurrency/index.js
@@ -82,12 +82,14 @@ const Currency = (props) => (
               stringValue = context.value;
             }
             const numberValue = numbro.unformat(stringValue);
+            // default to 0 if defaultValue is NaN
+            const baseValue = numberValue || 0;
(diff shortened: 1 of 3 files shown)


react/src/components/atoms/forms/InputCurrency/index.js
```diff
@@ -82,12 +82,14 @@ const Currency = (props) => (
               stringValue = context.value;
             }
             const numberValue = numbro.unformat(stringValue);
+            // default to 0 if defaultValue is NaN
+            const baseValue = numberValue || 0;
```
smurrayatwork:
```suggestion
            const baseValue = Number.isNaN(numberValue) ? 0 : numberValue;
```

smurrayatwork (approved):

PR #490, opened 2019-02-26
Patternlab/DP-6354  A11y - fix pagination to use links rather than buttons and access accessibility features

clair0917:
## Description
Changed the pagination buttons to links

## Related Issue / Ticket

- [DP-6354](https://jira.mass.gov/browse/DP-6354)

## Steps to Test

1. The new links can be viewed in the inspector 
in Mayflower: https://mayflower.digital.mass.gov/b/patternlab/DP-6354--a11y-pagination-buttons-to-links/patterns/05-pages-location-listing/05-pages-location-listing.html?_page=1

in Drupal: http://mass.local/orgs/massachusetts-registry-of-motor-vehicles/locations?_page=1

## Screenshots
screen shot 2019-02-26 at 12 04 58 pm


ygannett:
I like the `<button>`s are replaced with `<a>`s and adding some context with hidden text.

A couple of things I noticed and my suggestion:
1. Since these links are for navigation, I suggest to wrap them with `<nav>`.  So, it’s like `<nav class="ma__pagination js-pagination">`. If this would causes any conflicts, `div.ma__pagination__container` can be replaced with `<nav>`.  This way, assistive technology devices can recognize and treat as navigation elements.

2. There is nothing to tell they are pagination links to screen reader users.  A heading or something equivalent is necessary for the container `nav.ma__pagination` changed above. There are 2 options:
- add `aria-label` 
- add visually hidden heading
Using `aria-label` is recommended since it's easier not dealing with heading#.

3. Markup difference in pages:  I see 2 `<div class="ma__pagination js-pagination">`s in the Location Listing page.  It seems typo.
pattern_lab_-_pages-location-listing
In other pages, they have only 1 `<div class="ma__pagination js-pagination">`.
pattern_lab_-_pages-press-listing

FYI:  Here is a good sample of accessible pagination links:
https://www.a11ymatters.com/pattern/pagination/
In their sample, it has `<nav role="navigation" aria-label="Pagination Navigation”>`, but you don’t need `role=“navigation”` since `<nav>` element already has the role.

ygannett (commented):
I like the `<button>`s are replaced with `<a>`s and adding some context with hidden text.

A couple of things I noticed and my suggestion:
1. Since these links are for navigation, I suggest to wrap them with `<nav>`.  So, it’s like `<nav class="ma__pagination js-pagination">`. If this would causes any conflicts, `div.ma__pagination__container` can be replaced with `<nav>`.  This way, assistive technology devices can recognize and treat these links as navigation elements.

2. There is nothing to tell they are pagination links (= what this area is) to screen reader users.  A heading or something equivalent is necessary for the container `.ma__pagination`. There are 2 options:
- add `aria-label` 
- add visually hidden heading
Using `aria-label` is recommended since it's easier not dealing with heading#.

3. Markup difference in pages:  I see 2 `<div class="ma__pagination js-pagination">`s in the Location Listing page.  It seems typo.
pattern_lab_-_pages-location-listing
In other pages, they have only 1 `<div class="ma__pagination js-pagination">`.
pattern_lab_-_pages-press-listing

FYI:  Here is a good sample of accessible pagination links:
https://www.a11ymatters.com/pattern/pagination/
In their sample, it has `<nav role="navigation" aria-label="Pagination Navigation”>`, but you don’t need `role=“navigation”` since `<nav>` element already has the role.

Let me know for any questions.

======
I'm sorry that I accidentally hit the wrong button and closed the PR.  I meant to post my comment.

clair0917:
I've updated the pagination navigation to use a nav tag and aria attributes. Weirdly, I couldn't find the extra wrapper in the markup and when I tested by adding a new class and then removing it, the extra wrapper magically disappeared.

ygannett:
I missed something earlier and I'm sorry about that.
In the pagenation links, there is no "current" indicator for screenreader users. Since we have the "is-active" class for the current page, we could give the link a different aria-label value instead of "Go to Page #":
Pattern_Lab_-_molecules-pagination

The extra wrapper is still there😢
Pattern_Lab_-_pages-location-listing
I played around with the `pagenation.twig` locally and found something funky.  I don't think it's nothing to do with the template itself. 

If you put some text as a marker in the template and go to the event list page.  You see your marker. When you try the same thing in the location listing page, you need to go to the page first before updating your marker. Save the change and wait the page to get refreshed.  You see your marker for a few seconds, then the page gets refreshed again and your marker is gone and the extra wrapper shows up. So, the extra wrapper is coming from somewhere else. It seems the refresh is triggered by `@molecules/google-map.twig`.

ygannett:
LGTM.

ygannett (approved):
LGTM

assets/scss/02-molecules/_pagination.scss
```diff
@@ -29,8 +30,8 @@
       margin: 0 .1em;
     }
 
-    &:disabled {
-      cursor: default;
+    &.disabled {
```
clairesunstudio, line 36: why change disable from using attribution to class?

patternlab/styleguide/source/_patterns/02-molecules/pagination.twig
```diff
@@ -1,23 +1,32 @@
...
-      type="button"
-      {% if pagination.prev.disabled %}
-        disabled
-      {% endif %}>{{ pagination.prev.text }}</button>
+  <nav class="ma__pagination__container" aria-label="Pagination Navigation">
+    <a 
+      class="ma__pagination__prev js-pagination-prev {% if pagination.prev.disabled %}disabled{% endif %}"
+      {% if pagination.prev.disabled %}aria-disabled="true" {% else %}
+      aria-label="Go to {{ pagination.prev.text }} page"{% endif %}
+      href="#">{{ pagination.prev.text }}</a>
```
clairesunstudio, line 8: `<a>` should be using href, is it currently attaching onClick to `js-pagination-prev`?

clairesunstudio: should be adding `role="button"`

clairesunstudio (approved):

assets/scss/02-molecules/_pagination.scss
ygannett: I missed that in my review.
I believe it is related to the change from `<button>` to `<a>`.

> The disabled attribute is allowed only for form controls. Using it with an anchor tag (an <a> link) will have no effect.

https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/disabled

PR #502, opened 2019-03-08
DP-8334: Use NPM instead of Bower to pull in front end dependencies

rbayliss:
## Description

This PR removes switches from using Bower to NPM for dependencies in Mayflower Patternlab.  The motivation for doing this is that Bower is deprecated (see https://bower.io/), and Github does not do scanning of Bower dependencies for security vulnerabilities.  I tried to keep versions as close as I could, but in some cases I wasn't able to maintain the version or we needed to switch to a completely different package.  Things that are different:

* I have no clue where we got our basic polyfills from, but we're now using a combination of `core-js` (for ES level polyfills), and `mdn-polyfills` (for browser level polyfills).  These should be comparable, because of the nature of polyfills. 
* Specialized polyfills (fetch, UrlSearchParams, and picturefill) - have also been switched to what are probably new sources, but the new sources we're using are the most canonical ones I could find.
*  TwigJS - went from a claimed version of `0.10.5` to `1.13.2`, although this isn't actually as extreme as it seems, since `twigjs-bower` used Twig `1.10.4`.
* Fitvids - it was much easier to switch to the vanilla JS version of fitvids than to figure out how to import the jQuery fitvids plugin.  It seems to work the same though, so I think we should be fine here.

Additionally, at the end of this work, I ran `npm audit` to fix known vulnerabilities.  I believe the only things that were updated were build related (lots of `minimatch` updates).

The overall effect of this PR is that it increases our vendor JS size from 332K to 396K, which I consider an acceptable tradeoff for now.  We could reduce this further in the future by fully removing Handlebars (~100k, which we're barely using), and by splitting off the polyfills into a file that's only loaded for older browsers (50k).  

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-8334)
- [Github issue]()

## Steps to Test

1. Using IE 11, test the following pages in the [Mayflower preview](https://mayflower.digital.mass.gov/b/feature/DP-8334-remove-bower/index.html), interacting with the UI and just generally trying to break things:
    * Event Listing
    * Location Listing
    * Rules of Court 
    * Press Listing

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


ygannett (approved):
LGTM.
Tested in:
**Windows 7**
- IE 11
- FF 65

**Mac OS 10.14.3**
- Chrome72

Comparing to the current version (V9.2.1), the found display issues are not related to this change.
They are either global or IE specific css issues, which should be handled separately.

No behavior issues are found in any pages in all browsers listed above even after interacting with the UI.

PR #503, opened 2019-03-11
React/DP-12921: Prevents min/max from being past on InputNumber and InputCurrency.

smurrayatwork:
## Description
* This changes InputNumber and InputCurrency so that changes made can't go beyond the passed in min/max values.
* This also adds the[ npm package "is"](https://www.npmjs.com/package/is) to the repo. We should use this going forward for type checking.
* Focusing InputCurrency will now remove the placeholder text if the input has no value, and place it back on blur if there is still no value.
* Finally, InputCurrency has been changed so that focusing no longer switches the currency formatting to numbers.

## Related Issue / Ticket

- [DP-12921](https://jira.mass.gov/browse/DP-12921)
- [DP-12734](https://jira.mass.gov/browse/DP-12734)

## Steps to Test
1. Pull the branch, npm install, run storybook, and browse to atoms -> forms -> Input(Number or Currency). Ensure that either input can no longer go beyond their set min or max values.


clairesunstudio:
Since min/max only limits the inputs based on submission, and we are submitting values using onBlur. I think we should do a check in onBlur so that it's preventing out-of-bound values, both in inputCurrency and inputNumber.

I would suggest in handleChange/handleBlur we do a check and if input exceeds max, value will be set to max and if it exceeds min, it will be set to min. Let me know if that doesn't make sense.

clairesunstudio:
If the value onChange is updated to a value exceeding the limits. the up/down keys and buttons stop work, because the check of newValue against min/max
inputCurrency
This could be fixed if we do what i mentioned above to prevent value to be out of bound

clairesunstudio:
For InputCurrency formatting, is there a way to keep track on the actual number value with user interaction and have the display be a layer on top of that. 

e.g. when user backspace it will remove digits and will prevent the displayed `$` and `,` to be removed. 

User should not be able to modify the displayed `$` `,` `.`
inputCurrency-format


See these examples:
https://codepen.io/tutsplus/pen/PzpoWK
http://uipatterns.io/currency-input

clairesunstudio:
Changes work well except for InputNumber min is not setting to min onBlur:

Screen Shot 2019-03-15 at 9 44 14 AM
Screen Shot 2019-03-15 at 9 44 21 AM

react/src/components/atoms/forms/InputNumber/InputNumber.stories.js
```diff
@@ -14,8 +14,16 @@ storiesOf('atoms/forms', module)
     'InputNumber', (() => {
       const inputTextOptionsWithKnobs = Object.assign(...Object.entries(InputNumberOptions).map(([k, v]) => (
         { [k]: v() })));
+      const storyProps = {
+        style: (inputTextOptionsWithKnobs.inline) ? { width: '400px' } : { width: '200px' }
```
clairesunstudio, line 18: Is this only for testing purposes? do you want to you remove it to fix the visual regression?

clairesunstudio:
The other thing is in Form:
For some reason when input0 is set to 100 (max is 100) input1 becomes 90 instead of 0
form

Also onBlur only updates its own value doesn't rerender the other linked input, not sure if this is going to be an issue in the calculaters. i will npm link and test it out

react/src/components/atoms/forms/InputNumber/InputNumber.stories.js
smurrayatwork: This is only for the story, but I should update the backstop image.

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -78,10 +78,15 @@ const NumberInput = (props) => {
...
-            if (is.number(newValue)) {
-              newValue = Number(numbro(inputEl.value)
-                .format({ mantissa: countDecimals(props.step) }));
+            let newValue;
+            if (is.empty(inputEl.value)) {
+              newValue = inputEl.value;
+            } else {
+              newValue = Number(inputEl.value);
+              if (is.number(newValue)) {
+                newValue = Number(numbro(inputEl.value)
```
clairesunstudio, line 90: Number() doesn't seem necessary here

clairesunstudio: nvm i'm wrong

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -52,43 +55,74 @@ const NumberInput = (props) => (
...
+                  .format({ mantissa: countDecimals(props.step) }));
+              }
             }
-            const updateError = displayErrorMessage(newValue, props.min, props.max, props.required);
+            const updateError = displayErrorMessage(newValue);
```
clairesunstudio, line 94: props.required still needs to be passed to updateError

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -29,16 +33,15 @@ const NumberInput = (props) => (
             'ma__input-number-unit--showButtons': props.showButtons
           });
 
-          const decimalPlaces = countDecimals(props.step);
-
-          const displayErrorMessage = (val, min, max, isRequired) => {
-            if (isRequired && String(val).length === 0) {
+          const displayErrorMessage = (val) => {
+            const { min, max, isRequired } = props;
```
```suggestion
            const { min, max, required } = props;
```

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -29,16 +33,15 @@ const NumberInput = (props) => (
...
+            if (isRequired && is.empty(val)) {
```
```suggestion
            if (required && is.empty(val)) {
```

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -52,43 +55,74 @@ const NumberInput = (props) => (
...
+                newValue = props.max;
+              }
+              if (newValue < props.min) {
+                newValue = props.min;
+              }
+              inputEl.value = Number(numbro(newValue)
+                .format({ mantissa: countDecimals(props.step) }));
+              const updateError = displayErrorMessage(newValue);
+              context.updateState({ value: inputEl.value, ...updateError }, () => {
+                if (typeof props.onBlur === 'function') {
```
```suggestion
                if (is.fn(props.onBlur)) {
```

react/src/components/atoms/forms/InputNumber/index.js
```diff
@@ -52,43 +55,74 @@ const NumberInput = (props) => (
...
+              newValue = Number(inputEl.value);
+              if (is.number(newValue)) {
+                newValue = Number(numbro(inputEl.value)
+                  .format({ mantissa: countDecimals(props.step) }));
+              }
             }
-            const updateError = displayErrorMessage(newValue, props.min, props.max, props.required);
+            const updateError = displayErrorMessage(newValue);
             context.updateState({ value: newValue, ...updateError }, () => {
               if (typeof props.onChange === 'function') {
```
```suggestion
              if (is.fn(props.onChange)) {
```

clairesunstudio (approved):

react/src/components/atoms/forms/InputCurrency/index.js
```diff
@@ -3,211 +3,236 @@ import PropTypes from 'prop-types';
...
-                context.updateState({ showError, errorMsg, value: toCurrency(numberValue, 2) }, () => {
-                  // invokes custom function if passed in the component
-                  if (typeof props.onBlur === 'function') {
-                    // context.value won't be immediately changed, so pass new value over.
-                    props.onBlur(numberValue);
-                  }
-                });
-              }
-            },
-            onFocus: (e) => {
```
avrilpearl, line 137: Can you put this unformatting logic back in this pr?

changelogs/DP-12921.md
```diff
@@ -0,0 +1,7 @@
+___EXAMPLE___
+Minor
+Changed
+- (React) [InputNumber] DP-12921: Limits the component from changing value between the min and max passed.
+- (React) [InputCurrency] DP-12921: Limits the component from changing value between the min and max passed.
+- (React) [InputCurrency] DP-12734: Prevents changing currency to numbers from strings on focus.
```
avrilpearl: can you update the changelog here - removing this per comment below.

PR #504, opened 2019-03-11
Patternlab/Add class for GTM

clairesunstudio:
Patch
Added
(Patternlab) [OrganizationNavigation] DP-12928: Add link list specific classes: `ma__org-nav-i-want-to__findService`, `ma__org-nav-i-want-to__learnAbout`,`ma__org-nav-i-want-to__login` to sections for GTM. #504

[Jira](https://jira.mass.gov/browse/DP-12928)

To track organization subnav menu use, we need unique class identifiers such that we can differentiate between "Find a service or task", "Learn more about", and "Log in to" in Google Tag Manager.

Example org pages with different subnav menu configurations: 
https://www.mass.gov/orgs/civil-service
https://www.mass.gov/orgs/massachusetts-supreme-judicial-court

Right now, the three sections in org-sub-nav.png all have ma__org-nav-i-want-to-section as a class. This ticket is done when these three sections have different classes.

### DIFF
--- a/changelogs/DP-12928.md
+++ b/changelogs/DP-12928.md
@@ -1,1 +1,3 @@
+Patch
+Added
 (Patternlab) [OrganizationNavigation] DP-12928: Add Subnav specific class to sections for GTA. #504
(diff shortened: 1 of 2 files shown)


mrossi113 (approved):
👍 LGTM! This will need a Drupal PR to be implemented into production.

PR #529, opened 2019-04-09
[React] DP-13409: Fix onclick function not being called in Tabs Container

avrilpearl:
Patch
Fixed
- (React) [TabContainer, Tabs] DP-13409: Fixes bug causing the handleClick callback function on Tabs in TabContainer to not be called. #529

### DIFF
--- a/react/src/components/organisms/TabContainer/tab.js
+++ b/react/src/components/organisms/TabContainer/tab.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import is from 'is';
 import { TabContext } from './context';
 
 const Tab = React.forwardRef((props, ref) => (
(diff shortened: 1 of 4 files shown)


clairesunstudio:
tab and key up/down also needs to trigger handleClick

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -76,9 +84,9 @@ const Tab = React.forwardRef((props, ref) => (
             e.target.scrollIntoView({ behavior: 'smooth', block: 'nearest', inline: 'center' });
             if (activeTab !== tabIdent) {
```
clairesunstudio, line 76: i think this can be removed, it's condition is never true?

clairesunstudio (approved):

clairesunstudio (changes requested):
sorry wait, i think we need to expose the tab text to be able to use on the buget side. Currently, `onTabChange` is returning content. i think we need to create a ref for the project side to be able to select the inner text of the tab

clairesunstudio (approved):

clairesunstudio:
added necessary in budget in this pr --> https://github.com/massgov/budget/pull/207

PR #544, opened 2019-04-26
React/DP-13718: Fuzzy search default suggestions.

clairesunstudio:
Minor
Added
- (React) [InputTextFuzzy] DP-13688: Added option to render all options as suggestions by default, via prop `renderDefaultSuggestion`
- (React) [Storybook Config] DP-13688: Added storybook console addon

Removed
- (React) [InputTextTypeAhead] DP-13688: Deprecate InputTextTypeAhead.

To test:
yarn link-mayflower in budget: https://github.com/massgov/budget/pull/242


react/src/components/atoms/forms/InputTextFuzzy/index.js
```diff
@@ -73,6 +75,7 @@ class InputTextFuzzy extends React.Component {
           switch (event.key) {
             case 'ArrowDown':
             case 'ArrowUp':
+              event.persist();
```
clairesunstudio: don't think this is necessary unless we want to pass event to a callback, right?

react/src/components/atoms/forms/InputTextFuzzy/index.js
```diff
@@ -21,14 +21,16 @@ class InputTextFuzzy extends React.Component {
     this.fuse = new Fuse(this.props.options, fuseOptions);
   }
   handleChange = (e) => {
+    e.persist();
```
clairesunstudio: don't think this is necessary will check locally

react/src/components/atoms/forms/InputTextFuzzy/index.js
```diff
@@ -42,17 +50,23 @@ class InputTextFuzzy extends React.Component {
...
         const suggestions = this.optionsToSuggestions(this.props.options);
         this.setState({
           suggestions,
           value: ''
+        }, () => {
+          if (typeof this.props.onChange === 'function') {
+            this.props.onChange({ event: e, value, suggestions });
+          }
         });
+      } else {
```
clairesunstudio: is this change on purpose? moving the callback into setState in the if block will only callback if value is empty?

react/src/components/atoms/forms/InputTextFuzzy/index.js
```diff
@@ -20,16 +20,70 @@ class InputTextFuzzy extends React.Component {
...
-    const suggestions = this.fuse.search(e.target.value);
+  clearSuggestions = (event) => {
     this.setState({
-      value: e.target.value,
-      suggestions
+      suggestions: [],
+      highlightedItemIndex: null
+    }, () => {
+      if (is.fn(this.props.onSuggestionClick)) {
+        this.props.onSuggestionClick(event, {
```
clairesunstudio, line 29: i don't think we should pass this callback in blur

react/src/components/atoms/forms/InputTextFuzzy/index.js
```diff
@@ -69,13 +123,19 @@ class InputTextFuzzy extends React.Component {
         value: this.state.value,
         disabled: this.props.disabled,
         id: this.props.inputId,
-        onKeyDown: (event, { newHighlightedSectionIndex, newHighlightedItemIndex }) => {
+        onFocus: this.handleFocus,
+        onBlur: (event) => {
+          event.persist();
```
clairesunstudio: don't think we need to persist this event unless we want to pass a custom callback

react/src/components/atoms/forms/InputTextFuzzy/index.js
clairesunstudio: since we are not using state in props.onChange function, what's the difference putting it in callback vs outside of setState? if we pass it as callback we will have to do it in the setState below also

react/src/components/atoms/forms/InputTextFuzzy/index.js
clairesunstudio: onSuggestionClick is when you select on a suggestion, e.g. we use it to navigate to a url and that shouldn't be triggered onBlur

avrilpearl (approved):

PR #542, opened 2019-04-22
Feature/dp 13100  topic card links

broleomo:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](https://github.com/massgov/mayflower/blob/dev/docs/change-log-instructions.md)

## Description

- Styling the topic cards to match the Invision mockups provided in DP-13100. 

## Related Issue / Ticket

- Jira Ticket DP-13100

## Steps to Test

*Go here*
http://localhost:3000/?p=pages-topic

*Do This*
- Scroll to the topic cards.

*You should See*
- The topic cards matching the mockups outlined in this Invision link: https://eotss.invisionapp.com/share/XQR3BDXJ869#/screens

1. 

## Screenshots
N/A

## Additional Notes:

Anything else to add?

#### Impacted Areas in Application

* Section links and topic cards

PR #505, opened 2019-03-12
DP-12475 Remove the indentation feature part from richText.js

ygannett:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](https://github.com/massgov/mayflower/blob/dev/docs/change-log-instructions.md)

## Description

**TO-DO:  Set up something to maintain the same visual after removing richText.js.**

- Remove `richText.js` from the file system and `index.js` to replace the feature in Drupal.
- Remove js hook classes for indentation and anchor from `03-organisms/by-author/rich-text.twig`.
- Adjust rules-of-court.json's rich text output in sections to maintain the same visual presentation with the change.

## Related Issue / Ticket

- [\[BE + FE? Dev\] Convert Mayflower Patternlab's richtext.js into a Drupal text filter 2](https://jira.mass.gov/browse/DP-12475)
- [DP-12475 Convert Mayflower Patternlab's richtext.js into a Drupal text filter 2 #3040](https://github.com/massgov/mass/pull/3040)

## Steps to Test

1. In the file structure, check **richText.js** is removed from `patternlab/styleguide/source/assets/js/modules/` and in `patternlab/styleguide/source/assets/js/index.js`.

1. Go to the [Rules of court page](http://localhost:3000/?p=pages-rules-of-court).

1. Compare the section's indentations with the [current implementation](https://mayflower.digital.mass.gov/?p=pages-rules-of-court) for any difference.

## Screenshots
Use something like [licecap](http://www.cockos.com/licecap/) to capture gifs to demonstrate behaviors.

## Additional Notes:

Anything else to add?

#### Impacted Areas in Application

* 

#### @TODO

*

PR #547, opened 2019-05-01
Release 9.6.0

clairesunstudio:
### Added
- (Patternlab) [SectionLinks] DP-13100: Changed topic card heading styles, and footer CTA styles #542
- (React) [InputTextFuzzy] DP-13688: Added option to render all options as suggestions by default, via prop `renderDefaultSuggestion` #544
- (React) [Storybook Config] DP-13688: Added storybook console addon #544

### Removed
- (Patternlab) [assets] DP-12475: Remove richText.js. #505
- (Patternlab) [RichText] Removed js flag classes for richText.js. #505

### Fixed
- (React) [FilterBox] Fixed warning caused by using `start` instead of `flex-start` within stying. #543


mrossi113 (approved):
👍 LGTM!

assets/scss/00-base/_colors.scss
```diff
@@ -25,6 +25,7 @@ $c-granite-gray: #535353;
 $c-overcast-gray: #DCDCDC;
 $c-platinum-gray: #F2F2F2;
 $c-atlantic-gray: #707070;
+$c-very-light-gray: #CCCCCC;
```
avrilpearl, line 28: we should look at this - it should be added to the gray variable color scale in this file as well.

avrilpearl (approved):
approved - just want to follow up on that color variable addition/change

assets/scss/00-base/_colors.scss
clairesunstudio: good catch, so the structure of base color variables are not meant to be used direction, we want to pass these to a greyscale variable. Let's do follow up later

PR #550, opened 2019-05-02
add showButton prop to inputCurrency

clairesunstudio:
Minor
Added
- (React) [InputCurrency] DP-12875: Add `showButtons` prop to control whether the up/down buttons are shown. #550


avrilpearl (approved):
lgtm - just need circle tests to pass before merging it in

PR #559, opened 2019-05-15
React/DP-13924: Input fuzzy bug

smurrayatwork:
## Description
This PR addresses the following:
* Upgrades Storybook to its latest minor version.
* Upgrades Fuse.js to 3.4.4.
* Increases the number of characters that the Fuse library will process to 300. This was causing the issues reported in the ticket.
* Adds another fuse object to InputTextFuzzy that supports searching empty strings. This is used for displaying the entire result set when the input is empty.

## Related Issue / Ticket

- [Github issue](https://jira.mass.gov/browse/DP-13924)

## Steps to Test
1. Pull the branch, then run `nvm use`, `npm install`, `npm run start`. Browse to the InputTextFuzzy component in storybook and verify that none of its functionality has changed and the bug reported no longer occurs.

## Additional Notes:
Text highlighting/matching within the suggestions is matching exactly what the Fuse.js library returns. It appears like a bug, but this can be changed by using different configuration settings.


avrilpearl (approved):

clairesunstudio (approved):
Autowhatever is being switched out with autosuggest in https://github.com/massgov/mayflower/pull/566/

PR #564, opened 2019-05-23
DP-13970:  MassGov Search Jest Related Fixes

smurrayatwork:
## Description
This PR handles the following:
- (React) [Icon] DP-13970: Removed usage of webpack's require.context in favor of dynamic ES6 imports.
- (React) [DecorativeLink] DP-13970: Fixes warnings appearing related to the title prop when it's not set or passed.
- (React) [Pagination] DP-13970: Reworked props to use destructoring.
- (React) [ListingTable] DP-13970: Fixes the prop type definition for items to remove console warnings.
- (React) [HeaderSearch] DP-13970: Fixed console warning from using postInputFilter when it's not passed.
- (React) [ButtonWithIcon] DP-13970: Changed the storybook story to create Icons within each story instead of outside of them.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-13970)

## Steps to Test
1. Pull the branch and ensure that all previously existing functionality still works the same way.


react/src/components/molecules/HeaderSearch/index.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { componentWithName } from 'airbnb-prop-types';
```
clairesunstudio, line 3: nice! we should slowing replacing the `componentPropTypeCheck` utility function. Do they have a proptype with multiple components?

avrilpearl (approved):

react/src/components/molecules/HeaderSearch/index.js
smurrayatwork: They do!

clairesunstudio: cool! we should convert the other proptype checks dependent on the custom functions and convert them to use airbnb-prop-types

clairesunstudio (approved):

PR #600, opened 2019-05-31
remove yarn lock

clairesunstudio:
remove yarn.lock and just use npm


avrilpearl (approved):
you added an nvmrc to patternlab too! yay!

PR #616, opened 2019-06-07
React/HotFix: fix tab context export

avrilpearl:
- (React) [TabContext] hotfix: Fixes export of TabContext in react package build.

### DIFF
--- a/changelogs/react-hotfix-tabcontext-export.md
+++ b/changelogs/react-hotfix-tabcontext-export.md
@@ -1,0 +1,3 @@
+Patch
+Fixed
+- (React) [TabContext] hotfix: Fixes export of TabContext in react package build.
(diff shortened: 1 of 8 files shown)


react/src/index.js
```diff
@@ -113,7 +113,7 @@ export AccordionWrapper from './components/organisms/AccordionWrapper';
 export TableofContents from './components/organisms/TableofContents';
 export Tab from './components/organisms/TabContainer/tab';
 export TabContainer from './components/organisms/TabContainer';
-export TabContext from './components/organisms/TabContainer/context';
+export { TabContext } from './components/organisms/TabContainer/context';
```
clairesunstudio: i think instead of doing this, we should export default from the context file and that will also resolve the eslint error

changelogs/react-hotfix-tabcontext-export.md
```diff
@@ -0,0 +1,3 @@
+Patch
+Fixed
+- (React) [TabContext] hotfix: Fixes export of TabContext in react package build. #616
```
clairesunstudio: i would directly just update the changelog since its a hotfix so it can go into master

clairesunstudio:
just two small comments above otherwise tested and works great

clairesunstudio (approved):

PR #621, opened 2019-06-11
React/DP-14340:  emergency alert molecule.

smurrayatwork:
## Description
This PR addresses the following:
- (React) [EmergencyAlert] DP-14340: Adds the EmergencyAlert molecule from pattern lab.
- (React) [Link] DP-14340: Adds support for using props.children with Link.
- (React) [Link] DP-14340: Adds support for passing custom classes to Link component.
- (React) [Link] DP-14340: Sets default text to empty string.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-14340)

## Steps to Test

1. Pull the branch. Browse to molecules -> EmergencyAlert and verify that the component works as intended.

## Screenshots
Screen Shot 2019-06-11 at 4 06 51 PM


clairesunstudio:
I think in order to use it in budget, we also need the organism, which has the wrapper container --> https://mayflower.digital.mass.gov/?p=organisms-emergency-alerts

clairesunstudio:
themes:
Screen Shot 2019-06-12 at 1 03 07 PM

bg color vars:
`$c-duckling-yellow-lightest`
`$c-bay-blue-lightest`
`$c-berkshires-green-lightest`
`$c-gray-lightest`
`$c-error-red-lightest`

react/src/components/molecules/EmergencyAlert/style.scss
```diff
@@ -0,0 +1,82 @@
+@import "global";
+@import '00-base/mixins';
+@import '02-molecules/emergency-alert';
+@import '02-molecules/emergency-header';
```
clairesunstudio:
```suggestion
```

react/src/components/molecules/EmergencyAlert/style.scss
```diff
@@ -0,0 +1,82 @@
...
+
+
+.ma__emergency-alert--c-primary {
+  @include ma-button("normal","solid",$c-primary);
```
clairesunstudio: maybe we should create a callout mixin instead of using the button one which includes hover style etc.

react/src/components/molecules/Link/index.js
```diff
@@ -6,12 +6,13 @@ import { Icon } from '../../../index';
 
 const Link = (props) => {
   const icon = props.icon ? (<Icon name={props.icon} svgWidth={13.2} svgHeight={13.2} />) : '';
+  const classes = (props.classes) ? `js-clickable-link ${props.classes}` : 'js-clickable-link';
```
clairesunstudio:
```suggestion
```

react/src/components/molecules/Link/index.js
```diff
@@ -6,12 +6,13 @@ import { Icon } from '../../../index';
...
   return(
     <a
       href={props.href}
-      className="js-clickable-link"
+      className={classes}
```
clairesunstudio, line 17:
```suggestion
      className={props.classes}
```
or maybe we use the classname lib here

react/src/components/molecules/Link/index.js
clairesunstudio: js-clickable-link is not used

react/src/components/molecules/EmergencyAlert/EmergencyAlert.stories.js
```diff
@@ -0,0 +1,29 @@
...
+        theme: select('theme', {
+          'c-warning (default)': 'c-warning',
+          'no theme': '',
+          'c-primary': 'c-primary',
+          'c-primary-alt': 'c-primary-alt',
+          'c-gray-dark': 'c-gray-dark'
+        }, 'c-warning'),
+        message: text('message', 'MassPort is operating with limited flights due to road closures.'),
+        timeStamp: text('timeStamp', '02.15.16, 5:00 p.m'),
+        link: ({ theme, linkClasses, TextElement }) => <Link icon={null} classes={linkClasses} href="https://www.mass.gov"><TextElement>Read full alert</TextElement></Link>
```
clairesunstudio:
```suggestion
        link: ({ linkClasses, TextElement }) => <Link icon={null} classes={linkClasses} href="https://www.mass.gov"><TextElement>Read full alert</TextElement></Link>
```
theme is not consumed in Link, so don't think we need to pass here. Should also take out from linkArgs?

smurrayatwork: It shouldn't be taken out. This is to give the ability to directly pass the theme to the link function, in case somehow it wasn't known to whatever is using EmergencyAlert. I put it in the story so that people know it exists.

react/src/components/molecules/Link/index.js
smurrayatwork: I don't think this should be done. There's no defaultProp set for `classes`, so just assuming it always exists when it's not required is bad.

react/src/components/molecules/EmergencyAlert/EmergencyAlert.stories.js
clairesunstudio: gotcha, ok!

react/src/components/molecules/Link/index.js
clairesunstudio: I think having an unused class there is also not a great solution. can we set it to null? or conditionally passing className? can we use the classnames lib>

react/src/components/molecules/EmergencyAlert/style.scss
```diff
@@ -0,0 +1,43 @@
+@import "global";
+@import '00-base/mixins';
```
avrilpearl: this is in global so you don't need the sep import here.
```suggestion
```

react/src/components/molecules/EmergencyAlert/style.scss
```diff
@@ -0,0 +1,43 @@
...
+@import '02-molecules/emergency-alert';
+
+@mixin ma-emergency-alert ($color) {
```
avrilpearl: can you move lines 5-43 into the assets directory file that you are importing above (line 3) i.e. `@import '02-molecules/emergency-alert';`

react/src/components/molecules/EmergencyAlert/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+EmergencyAlert.propTypes = {
+  /** A string that controls different color themes for the component. */
+  theme: PropTypes.oneOf(['', 'c-highlight', 'c-primary-alt', 'c-primary', 'c-gray', 'c-error']),
+  /** A message describing the event. */
+  message: PropTypes.string.isRequired,
+  /** A string representing the time of the event. */
+  timeStamp: PropTypes.string,
+  /** An optional function whose return value is a link to take the user to page with more information. */
+  link: PropTypes.func
```
avrilpearl: same as on the other molecule, can we make it possible to pass a string here or function?

react/src/components/molecules/EmergencyAlert/index.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import is from 'is';
+import './style.css';
+
+const EmergencyAlert = (props) => {
+  const { message, timeStamp, link } = props;
+  const theme = props.theme.length === 0 ? 'c-highlight' : props.theme;
```
avrilpearl: should we switch to using classnames here?

clairesunstudio: I think we should, e.g.
```suggestion
  const alertClasses = classNames({
    'ma__emergency-alert': true,
    [`ma__emergency-alert--${props.theme}`]: props.theme
  });
```

react/src/components/molecules/EmergencyAlert/index.js
```diff
@@ -0,0 +1,44 @@
...
+          &nbsp;{link({ ...linkArgs })}
+        </span>
+      )}
+    </p>
+  );
+};
+
+EmergencyAlert.propTypes = {
+  /** A string that controls different color themes for the component. */
+  theme: PropTypes.oneOf(['', 'c-highlight', 'c-primary-alt', 'c-primary', 'c-gray', 'c-error']),
```
clairesunstudio: since there's a default prop for theme, i think we don't need an empty default
```suggestion
  theme: PropTypes.oneOf(['c-highlight', 'c-primary-alt', 'c-primary', 'c-gray', 'c-error']),
```

react/src/components/molecules/EmergencyAlert/EmergencyAlert.stories.js
```diff
@@ -0,0 +1,30 @@
...
+import Link from '../Link';
+
+storiesOf('molecules', module)
+  .addDecorator(withKnobs({ escapeHTML: false }))
+  .add(
+    'EmergencyAlert', (() => {
+      const props = {
+        theme: select('theme', {
+          'c-highlight (default)': 'c-highlight',
+          'no theme': '',
```
clairesunstudio: default prop is set to 'c-highlight'
```suggestion
```

react/src/components/molecules/EmergencyAlert/style.scss
```diff
@@ -0,0 +1,43 @@
...
+      svg {
+        fill: tint($c-font-base, 50%);
+      }
+      &:hover {
+        border-bottom-color: transparent;
+      }
+    }
+  }
+}
+$colors: (
```
clairesunstudio: 💯👏 great solution!

react/src/components/molecules/EmergencyAlert/style.scss
```diff
@@ -0,0 +1,43 @@
+@import "global";
+@import '00-base/mixins';
+@import '02-molecules/emergency-alert';
+
+@mixin ma-emergency-alert ($color) {
+  color: $c-font-base;
+  background-color: tint($color, 90%);
+  border-color: tint($color, 50%);
```
clairesunstudio: I think in patternlab (shared scss) only the subsequent border top is set 
```suggestion
    & + & {
    border-top-color: $color;
  }
```

react/src/components/molecules/Link/index.js
```diff
@@ -1,17 +1,22 @@
...
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
 import './style.css';
 
 import { Icon } from '../../../index';
 
 const Link = (props) => {
   const icon = props.icon ? (<Icon name={props.icon} svgWidth={13.2} svgHeight={13.2} />) : '';
+  const classes = classNames({
+    'js-clickable-link': true,
```
clairesunstudio, line 11: i still think we should remove this, in general we should clean up unused class and also, currently Link classes is not the a knob in the story we should add that. But don't have to be in this pr

clairesunstudio (approved):

avrilpearl (approved):

PR #637, opened 2019-06-20
React/DP-14424: TabContainer, Tab Refactor

smurrayatwork:
## Description
TabContainer and Tab needed a refactor so that they could better handle server side rendering, and not track Tab content within TabContainer. To that end, [React Portals](https://reactjs.org/docs/portals.html) are being used. To support server side rendering, the [react-portal](https://github.com/tajo/react-portal) package was added. This is simply a portals helper and does not provide the actual portals API, as that's directly from React.

The TabBody component was created for rendering Tab's children directly into TabContainer's tab body div. This is accomplished using [MutationObserver,](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) allowing the TabBody component to watch for DOM changes to a specific DOM node. When TabContainer's tab body div is rendered and added to the document, TabBody's state is updated with that DOM node, which is then used to render Tab's children.

This PR should be used [to test a similar PR on the budget repo](https://github.com/massgov/budget/pull/306).
## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-14424)

## Steps to Test
1. Pull the branch, then: `nvm use`, `npm install`, `npm run start`.
2. Browse to Organisms -> TabContainer and verify that none of the visible functionality has changed from before this branch was made.


clairesunstudio:
Noticed this one bug, otherwise work like a charm
tabs
go to a subtab, use keyboard arrows to navigate, and going beyond the first or the last tab break

react/src/components/organisms/TabContainer/tab.js
```diff
@@ -3,109 +3,113 @@ import PropTypes from 'prop-types';
...
+      tabIndex: active ? 0 : -1
+    };
+    return(
+      <React.Fragment>
+        <li role="presentation" className={tabClasses}>
+          <button {...buttonProps}>{this.props.title}</button>
+        </li>
+        {global.window && (
+          <TabBody active={activeTab === tabIdent} tabContainerBodyId={this.context.tabContainerBodyId}>
+            {this.props.children}
```
clairesunstudio, line 98: I'm curious why you think this should use portal, as the react tree structure is the same as the DOM tree. In the doc portal is used specifically if we want the structure to be different

clairesunstudio (approved):

clairesunstudio:
confirmed bug fixed in budget (develop)

avrilpearl (approved):

PR #649, opened 2019-06-27
Patternlab/DP-14350  new navigation

clair0917:
## Description
Updates header and relationship indicators in Mayflower so they are ready to launch when testing is complete

## Related Issue / Ticket

- [DP-14350](https://jira.mass.gov/browse/DP-14350)

## Steps to Test

1. The new menu and relationship indicators can be tested here: 
https://mayflower.digital.mass.gov/b/patternlab/DP-14250--new-navigation/patterns/05-pages-topic/05-pages-topic.html

2. I couldn't find a set of header tags in a page, so you can test their updates here:
https://mayflower.digital.mass.gov/b/patternlab/DP-14250--new-navigation/patterns/02-molecules-header-tags/02-molecules-header-tags.html

## Screenshots
Screen Shot 2019-06-12 at 9 24 23 AM


patternlab/styleguide/source/assets/js/modules/mainNav.js
```diff
@@ -1,232 +1,301 @@
-export default (function (window, document, $, undefined) {
+const body = document.querySelector("body");
+let width = body.clientWidth;
+const menuButton = document.querySelector(".js-header-menu-button");
+let menuButtonText = menuButton.querySelector('.ma__header__menu-text');
```
ygannett: We use a unique prefix for selectors for JS, 'js-' to separate one for css 'ma__' like `<div class="ma__header__search js-header-search-menu">`. 

If the existing markup doesn't have one for JS, feel free to add one.  
This way, we can manage selectors independently between JS and styles without risking breaking either styles or JS as we refactor templates, styles or JS.

patternlab/styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
```diff
@@ -1,5 +1,5 @@
 <div class="ma__site-logo">
-  <a href="/" title="Mass.gov home page">
+  <a href="/" id="ma__site-logo-link" title="Mass.gov home page">
```
ygannett, line 2: What is this id used for?

ygannett (changes requested):
### Menu:

- [Chrome, Safari] In the desktop display, the focus outline for the Menu overwraps on the "Select Language" .
- [FF] The focus outline for the Menu is not visible.
- [Keyboard in all browsers] In the menu dropdown, as closing sub menu area with ESC key, focus remains in the sub menu area.  It looks like users lost focus. Its top menu has "x" to indicate the submenu is open instead of "+". As a sub menu container closes, focus moves to its parent top menu item and the top menu item's symbol changes from "x" to "+".
- [Safari + VoiceOver] Unable to access sub menu. As a top menu is on focus and "ctl + option + space", which is equivalent to "enter" in VoiceOver, is used, the sub menu opens then closes right away. "Enter" doesn't work in this case.

Note: I didn't test in IE 11 since @clair0917 mentioned she'd been working on the related issues.

### Utility nav:

- if users don't close the drop-down content container of the utility nav as they tab through elements, the container stays open. As the focus moves out of the container, it is not relevant to users anymore. If any elements in the container including the container itself has no focus, the container shouldn't be displayed. In other words, as the focus moves out of the container, the container should close.
- in the drop-down content container of the utility nav, as tabbing through elements in the container, after the last focusable item, the focus moves out of the container to the next item in its parent's siblings and the content container remains open. To sighted users, it looks like they lost focus.
   
   Users should always decide what they do next. In this case, whether they stay in the tab content container or leave.

   After a quick conversation with Dan, it'd more make sense that:
      
    1. users tab through the focusable items in the content container, and when they get to the last focusable item, hitting TAB key moves the focus to the first focusable item in the container. In other words, focus makes a circle in the focusable items in the container as they hit TAB key.
    1. setting the close button focus as the last item in the container seems to more make sense to indicate users has gone through all focusable items in the container, and to suggest next possible action. In this case, visual change is not necessary since it is relatively small area to be able to recognize all content in the container. Just focus order needs to be adjusted.
    1.  users leave the container by hitting ESC key or CLOSE button in the container. 
    1. as users leave the container, the container close and focus goes back to its parent link, NOT to next or previous link. In other words, focus follows focusable items in the structure of the markup.




### Relationship indicator:

**02-molecules/relationship-indicators.twig**

- The component is not semantically marked up. Suggested markup based on the sample page:
  ```
    <div class="ma__relationship-indicators">
      <div class="ma__relationship-indicators--section primary tags-open">
        <span class="ma__relationship-indicators--label" id="LABEL1">
          <span class="ma__visually-hidden">This page, PAGE TITLE, is a </span>
          Part of
        </span>
        <ul class="ma__relationship-indicators--terms" aria-labelledby="LABEL1">
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Department of Conservation and Recreation</a>
          </li>
          <li class="wrapped-link ma__relationship-indicators--term" style="">
            <a href="#">Other agency or organization</a>
          </li>

          <li>
            <button type="button" class="js-relationship-indicator-button is-open" style="display: inline-block;">
              show <span class="tag-count">1</span> <span class="tag-state">more</span>
              <span class="ma__visually-hidden"> associated pages with this page, PAGE TITLE</span>
            </button>
          </li>
        </ul>
      </div>

      <div class="ma__relationship-indicators--section secondary">
        <span class="ma__relationship-indicators--label" id="LABEL2">
          <span class="ma__visually-hidden">This page, PAGE TITLE, is </span>
          offered by
        </span>

        <ul class="ma__relationship-indicators--terms" aria-labelledby="LABEL2">
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Snowmobile in the Parks</a>
          </li>
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Cross-Country Skiing</a>
          </li>
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Accessible Hiking</a>
          </li>
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Scenic Viewing Areas</a>
          </li>
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Hunting</a>
          </li>
          <li class="wrapped-link ma__relationship-indicators--term">
            <a href="#">Biking</a>
          </li>

          <li>
            <button type="button" class="js-relationship-indicator-button" style="display: inline-block;">
                show <span class="tag-count">5</span> <span class="tag-state">more</span>
                <span class="ma__visually-hidden"> organizations related to this page, PAGE TITLE.</span>
            </button>
          </li>
        </ul>
      </div>
    </div>
  ```

     - Add context info to `.ma__relationship-indicators--label` for screen reader users.
     - Add `id` to `.ma__relationship-indicators--label` for `aria-labelledby` in `.ma__relationship-indicators--terms`.  The label is not necessary to be `<h#>` in this case because of the association with the `aria-labelledby`.
     - Group links as a list to clarify their association.
     - Use `aria-labelledby` to semantically label the list for screen reader users. 
     - `<button>` needs context info for screen reader users.

- Don't pass any values for `alt` and `title` to **01-atoms/11-text/wrapped-link.twig**. Maybe, its json file needs to be changed.

**01-atoms/11-text/wrapped-link.twig**

- Completely remove `alt` which is invalid in `<a>`.  Any `<a>` is NOT supposed to have `alt`.
- `title` is not supposed to be added in this case. You might set a condition for this one just in case there are some valid use cases somewhere else.

patternlab/styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
clair0917: This was added so I could catch the focus to close the menu. The rest of the code for it was waiting for the utility menu. That code has now been added to use this link ID

ygannett: Sounds good.

patternlab/styleguide/source/_patterns/02-molecules/relationship-indicators.twig
```diff
@@ -33,10 +35,10 @@
         
       {% endfor %}
       {% endblock %}
-      <button type="button" class="js-relationship-indicator-button">
+      <button type="button" class="js-relationship-indicator-button" role="button" aria-pressed="false">
```
ygannett, line 38: The button is supposed to be wrapped with `<li>` for itself since it's inside the `<ul>`.

clair0917:
New changes to this PR will now be tracked in ticket https://jira.mass.gov/browse/DP-15110

ygannett (changes requested):
For the relationship indicator, now its focus order is:
1. 1st item
2. button
--- click the button to display the hidden items ---
3. un-hidden 1st item
...

Although the focus order is correct, as the button is clicked, focus moves to the third item by itself, which is not supposed to happen.  Moving focus should be done by user's will.  This is especially important for screen reader users, which wouldn't see the focus change and they would get confused with their whereabout.

The expected behavior is;

- as users click the button to display the hidden items, the hidden items are displayed and focus stays on the button,
- as users tab when the current focus is on the button, focus moves to the first unhidden item.

This focus order might to be achieved more easily and simply with CSS flexbox than JS.

ygannett:
I also noticed that the **titles** in the _links_ are still showing, whose values are duplicates to their link labels and unnecessary.
```
<ul class="ma__relationship-indicators--terms" aria-labelledby="primary">
  <li class="wrapped-link js-wrapped-link ma__relationship-indicators--term">
    <a href="#" title="Department of Conservation and Recreation">Department of Conservation and Recreation</a>
  </li>

  <li class="wrapped-link js-wrapped-link ma__relationship-indicators--term" style="display: list-item;">
    <a href="#" title="Other agency or organization">Other agency or organization</a>
  </li>

  <button type="button" class="js-relationship-indicator-button is-open" role="button" aria-pressed="true" style="display: inline-block;">
  show <span class="tag-count">1</span> <span class="tag-state">fewer</span>
  </button>
</ul>
```

PR #657, opened 2019-07-03
react/DP-14730: Address all lint errors/warnings on atoms

avrilpearl:
Description: Would like to be running lint checks on mayflower-react in circle ci. However, in order to do so we need to address the existing lint errors/warnings. This pr fixes all warnings/errors on the atoms in mayflower-react.


clairesunstudio:
looks good, found this bug to follow up on --> https://jira.mass.gov/browse/DP-14753

clairesunstudio:
Also as a follow up, we should address all the console errors in the atoms stories, e.g.
Screen Shot 2019-07-08 at 12 55 24 PM

clairesunstudio (approved):

PR #662, opened 2019-07-05
[react] emergency alerts overlay fix

clairesunstudio:
Minor
Changed
- (Patternlab, React) [EmergencyAlerts] DP-14741: Allow EmergencyAlerts to render everywhere not just right below header, specify mobile top positioning with `.ma__ajax-pattern` wrapper and fix z-index to not overlay main nav dropdown.  #662

----

Currently, the z-index for emergency alerts is higher than the main nav drop down, resulting:

Screen Shot 2019-07-05 at 9 48 08 AM

Things to check on the patternlab side:
1. Will this change have any negative impact on emergency alerts on mass.gov?
2. Is there currently an overwrite of the z-index on mass.gov handling this issue?
3. Another style we need to overwrite in projects is the top 40px on mobile, what is the reason for it?
```
  @media($bp-header-toggle-max) {
    top: 40px;
    // z-index:1;
  }
```

In mayflower react alert use, we have to overwrite both the z-index and the top position to 0px, e.g. budget, calculators, etc.


ygannett (approved):
The change looks good to me.

> 1. Will this change have any negative impact on emergency alerts on mass.gov?

In Mass.gov, the alerts below the header using a different template. Thus, different styles, which are assigned to `div.ma__header-alerts`, are applied.  `.ma__emergency-alerts` is only used at the top of page above the header, so technically nothing gets over-wrapped.  It makes me wonder the user of the z-index setting.

The bottom line is that no impact to worry about. 

> 2. Is there currently an overwrite of the z-index on mass.gov handling this issue?

As noted above, alerts in the location has different template, so no issue seen in react.  The alert container, `.ma__header-alert`, has `z-index:1;`, which is much lower setting than the dropdown's, `$z-dropdown: 90;`.

You can see it at L. 5 in **assets/scss/03-organisms/_header-alert.scss**.

> 3. Another style we need to overwrite in projects is the top 40px on mobile, what is the reason for it?

That is to visually push down the alert container below the `.ma__header__button-container`, which is inside `<header>`.  Without the setting, the top part of the alert is covered by the `.ma__header__button-container`.
mobile

Update review on 7/8/2019:
Tested in mobile in Drupal.  No overwrap with the top menu bar. 

Looks good.  Ready to merge.

PR #677, opened 2019-07-18
DP-15056: Update contact-row template

merauluka:
## Description
Added blocks around contact fields to allow overriding each on a field-by-field basis. This prevents having to duplicate markup in Drupal in order to inject the contact information.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-15056)
- [Related Drupal issue](https://jira.mass.gov/browse/DP-11107)

## Steps to Test

This should have no affect on the output of the markup in patternlab. So the test is to verify that markup on the contact-row template is outputted as before.

1. /?p=organisms-contact-row
2. /?p=pages-contact-row-examples

### DIFF
--- a/changelogs/DP-15056.txt
+++ b/changelogs/DP-15056.txt
@@ -1,0 +1,3 @@
+Minor
+Added
+- (Patternlab) [contact-row] DP-15056: Added blocks around contact groups to allow overriding on a field-by-field basis.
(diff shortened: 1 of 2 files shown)


lannyheidbreder-current:
If there are any differences after this change, I can't see them.

lannyheidbreder-current (approved):

PR #683, opened 2019-07-22
DP-9274 MF [a11y] Add more context to more link text

lannyheidbreder-current:
## Description

Prepared the `link` component to accept the labelContext variable for accessible, lengthy descriptions. Updated the `press-listing-as-grid` component's style guide example to use this new variable.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-9274)

#### Go here and do this

1. `npm run gulp` the style guide
2. Visit http://localhost:3000/?p=atoms-link
3. Inspect the link in the browser developer tools

#### What you should see

A previously unseen visually-hidden span with " to show as an example" following the visible link text

#### Go here and do this

1. Go to http://localhost:3000/?p=organisms-press-listing-as-grid
2. Scroll to the bottom and inspect the "See all news and announcements" link in the browser developer tools

#### What you should see

A visually-hidden span with " for the blah blah location" after the visible link text (this is the same as the previous title text)

## Additional Notes:

To fully achieve the goal of DP-9274, some back-end changes will be required. A ticket for that part of the work is forthcoming.

#### @TODO

* Create ticket for back-end portion of the work
    * Work needed is to change the method that prepares the news section "More" links to output the accessibility text after the link text instead of a `title` attribute on the link

### DIFF
--- a/changelogs/DP-9274.md
+++ b/changelogs/DP-9274.md
@@ -1,0 +1,6 @@
+___DESCRIPTION___
+Patch
+Added
+- (Patternlab) [Link] DP-9274: Added `labelContext` variable for an optional visually-hidden suffix #TK
+Changed
+- (Patternlab) [PressListingAsGrid] DP-9274: Changed JSON to use the `link` component’s new `labelContext` variable in the style guide #TK
--- a/changelogs/DP-9274.md
+++ b/changelogs/DP-9274.md
@@ -1,6 +1,6 @@
 ___DESCRIPTION___
 Patch
 Added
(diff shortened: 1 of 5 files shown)


ygannett (changes requested):
Looks great!  Thank you for making this change.

One little thing.  There is another more link at L.1021 in patternlab/styleguide/source/_patterns/05-pages/organization-elected-official.json.  This set uses the same template as the one you updated above at L. 997, the "more" set.  To maintain consistency of the sample page, would you update the "pastMore" set as well?

ygannett (approved):
Looks good.  Thank you for making the change.

PR #700, opened 2019-08-12
Patternlab/DP-14912  mobile sticky toc fix

clair0917:
## Description
Repositions the sticky TOC on mobile screens

## Related Issue / Ticket

- [DP-14912](https://jira.mass.gov/browse/DP-14912)

## Steps to Test

1. On a mobile phone, open the link and scroll the page until the sticky TOC gets stuck to the top.
2. Click to open the link
https://mayflower.digital.mass.gov/b/patternlab/DP-14912--mobile-sticky-TOC-fix/patterns/05-pages-regulation/05-pages-regulation.html
3. You should see that the stick TOC now sits below the menu, rather than being stuck underneath it.

## Screenshots
File

### DIFF
--- a/patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig
+++ b/patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig
@@ -8,7 +8,7 @@
     <div class="ma__sticky-toc__container">
       <h2 class="ma__sticky-toc__title" id="page-contents">{{ stickyTOC.title }}</h2>
       <nav class="ma__sticky-toc__links" aria-labelledby="page-contents">
-        <button class="ma__sticky-toc__toggle-link">+</button>
+        <button class="ma__sticky-toc__toggle-link"></button>
         <div class="ma__sticky-toc__column">
           <div class="ma__sticky-toc__secondary-label">{{ stickyTOC.title }} <span class="secondary-label-close"></span></div>
           {% for link in stickyTOC.links %}
--- a/assets/scss/03-organisms/_sticky-toc.scss
+++ b/assets/scss/03-organisms/_sticky-toc.scss
@@ -90,15 +90,27 @@
     line-height: .65em;
     overflow: hidden;
     position: absolute;
-      right: 10px;
-      top: 9px;
-    transition: transform .2s ease;
-    transform-origin: center center;
+    right: 10px;
+    width: 100%;
+    height: 100%;
+    max-height: 45px;
+   
     z-index: 1;
 
-    .is-open & {
-      transform: rotate(135deg);
+    &::after {
+      content: "+";
+      position: absolute;
+      top: 10px;
+      right: 10px;
+      transition: transform .2s ease;
+      transform-origin: center center;
+
+      .is-open & {
+        transform: rotate(135deg);
+      }
     }
+
+    
 
     @media ($bp-x-small-min) {
       display: none;
@@ -498,6 +510,8 @@
       top:0;
       right: 0;
       transition: right 0.3s ease-out;
+      position: fixed;
+      z-index: 9999;
     }
   }
 }
--- a/assets/scss/03-organisms/_sticky-toc.scss
+++ b/assets/scss/03-organisms/_sticky-toc.scss
@@ -447,7 +447,7 @@
   
   overflow: auto;
 
-  @media($bp-small-min) {
+  @media($bp-medium-min) {
    padding-top: 35px;
   }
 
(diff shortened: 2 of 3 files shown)


patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig
```diff
@@ -8,7 +8,7 @@
     <div class="ma__sticky-toc__container">
       <h2 class="ma__sticky-toc__title" id="page-contents">{{ stickyTOC.title }}</h2>
       <nav class="ma__sticky-toc__links" aria-labelledby="page-contents">
-        <button class="ma__sticky-toc__toggle-link">+</button>
+        <button class="ma__sticky-toc__toggle-link"></button>
```
ygannett, line 11: \<button\> cannot be empty (= no label text). Screen reader users wouldn't know what's the button is for.  The original label "+" was not helpful to them either.  So, this is a good opportunity to make it right.

1. Please remove the pseudo content "+" from **.ma__sticky-toc__toggle-link::after**.  Screen readers read it, but doesn't deliver helpful information to users.

1. Put "+" back into the `<button>` wrapping with `<span aria-hidden="true">`, so screen readers don't announce it.

1. Please add visually hidden text to `<button class="ma__sticky-toc__toggle-link">`:
```<span aria-hidden="true">+</span><span class="ma__visually-hidden">Show table of contents</span>```


---

L. 38
Mass_Gov
```
<div class="ma__sticky-toc__sticky-toggle">
          <button class="stickyTOC-open">{{ stickyTOC.title }}</button>
</div>
```
Please add the visually hidden text to describe the purpose of the button:
```
<button class="stickyTOC-open"><span class="ma__visually-hidden">Show </span>{{ stickyTOC.title }}</button>
```

Once the button is clicked, the focus stays on the button, which is generally an expected behavior.  As tabbing through from there, focus moves onto the followed content instead of the displayed overlay.  Thus, users don't have access to the overlay content. 

So, in this case, as the overlay content is displayed, the focus needs to move the overlay container, **div.ma__sticky-toc__column**.  Note that focus wouldn't be set on a focusable element at this point.  As the container has focus and users tab, then focus moves to the first focusable element in the container.

File: patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig [hunk 2]
```diff
@@ -8,7 +8,7 @@
...
         <div class="ma__sticky-toc__column">
           <div class="ma__sticky-toc__secondary-label">{{ stickyTOC.title }} <span class="secondary-label-close"></span></div>
```
ygannett, line 13: Mass_Gov_1

The close button doesn't seem to get focus via keyboard and screen readers since its `<span>`. 
This is an assumption since I cannot test this part.  Here are the expected presentation and behaviors.

1. the close button needs to be focusable and recognized as a button by AT devices.
    - use `<button>` instead of `<span>` if no issues by doing so.

1. it needs its label. So, it needs the similar treatment as **.ma__sticky-toc__toggle-link** above.
   - remove the pseudo content "x" from **.ma__sticky-toc__secondary-label .secondary-label-close::after**. Screen readers read it, but doesn't deliver helpful information to users.
   - put "x" in the button wrapping with `<span aria-hidden="true">`, so screen readers don't announce it.
   - add visually hidden text to the button:
```<span aria-hidden="true">x</span><span class="ma__visually-hidden">Hide the table of contents</span>```



1. As the close button is clicked, the focus moves back on the  `<button class="stickyTOC-open">`.

ygannett (changes requested):
Please see the following comments.

patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig
ygannett: Move this issue to DP-15605.

patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig
ygannett: Move this issue to DP-15605.

patternlab/styleguide/source/_patterns/03-organisms/by-template/sticky-toc.twig
ygannett: Move this issue to DP-15605.

ygannett (approved):
LGTM.
The accessibility issues are moved to DP-15605.

PR #701, opened 2019-08-13
DP-9085 MF problem with desktop address display on event item

lannyheidbreder-current:
## Description
Edited several components' example JSON to use line breaks in addresses.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-9085)

#### Go here and do this

Go to the following Mayflower pages:

- ?p=organisms-contact-row
- ?p=templates-event
- ?p=pages-event
- ?p=pages-event-end-date-with-time
- ?p=pages-event-end-date
- ?p=pages-event-public-meeting
- ?p=pages-event-recurring

#### What you should see

In each case, the address lines should be separated by `<br>` elements, not commas.

Note that sometimes, one *functional* address line will have a comma inside it, and the address may wrap after the comma. For example, in the last two pages listed above, “Division of Fisheries and Wildlife Field Headquarters, Richard Cronin Building” is one functional line, since it is the entire name of the location.

## Screenshots

Before:

9085-before

After:

9085-after


ygannett (approved):
LGTM.

PR #709, opened 2019-08-16
DP-15599: Prevent SVGs in SocialLink from receiving keyboard focus

apalmer0:
## Description

We're building an application that uses the Mayflower React footer, and recently ran through an accessibility audit. One of the recommendations from the company performing the audit was to provide alternative text for images, which called out the SVGs in the social links. However, because those images are nested inside anchor tags which do have the "textual equivalent", it was recommended instead to:

> hide the svg element from assistive technologies by assigning it an aria-hidden="true" attribute and focusable="no" so that it does not receive keyboard focus.
>
> [Compliant Code Example]
> <a...>\<svg focusable="no" aria-hidden="true">..\</a>

## Related Issue / Ticket

- [DP-15599](https://jira.mass.gov/browse/DP-15599) (I think this is the link...)

## Steps to Test

1. Sorry - looks like the link to the deploy docs isn't working either, but I don't believe there's much to test here... just ensuring that those attributes are present on the SVG elements.

## Screenshots

No behavior or UI is changed here, so I don't believe these are necessary.

## Additional Notes:

There might be a better solution to this issue than what I've implemented, or it might not even be necessary... I don't have much experience with accessibility improvements so I'm just going off what the accessibility consultants recommended. 

* Anywhere the `SocialLink` component is used.  

#### @TODO

* N/A

### DIFF
--- a/react/src/components/molecules/SocialLinks/index.js
+++ b/react/src/components/molecules/SocialLinks/index.js
@@ -26,7 +26,7 @@ const icons = {
     width="11"
     height="20"
     viewBox="0 0 11 20"
-    focusable="no"
+    focusable="false"
     aria-hidden="true"
   >
     <path
@@ -43,7 +43,7 @@ const icons = {
     width="21"
     height="20"
     viewBox="0 0 21 20"
-    focusable="no"
+    focusable="false"
     aria-hidden="true"
   >
     <path
@@ -59,7 +59,7 @@ const icons = {
     width="20"
     height="20"
     viewBox="0 0 20 20"
-    focusable="no"
+    focusable="false"
     aria-hidden="true"
   >
     <path
@@ -76,7 +76,7 @@ const icons = {
     width="21"
     height="17"
     viewBox="0 0 21 17"
-    focusable="no"
+    focusable="false"
     aria-hidden="true"
   >
     <path
@@ -93,7 +93,7 @@ const icons = {
     width="21"
     height="16"
     viewBox="0 0 21 16"
-    focusable="no"
+    focusable="false"
     aria-hidden="true"
   >
     <path
--- a/react/src/components/molecules/SocialLinks/index.js
+++ b/react/src/components/molecules/SocialLinks/index.js
@@ -26,6 +26,8 @@ const icons = {
     width="11"
     height="20"
     viewBox="0 0 11 20"
+    focusable="no"
+    aria-hidden="true"
   >
     <path
       id="SvgjsPath1007"
@@ -41,6 +43,8 @@ const icons = {
     width="21"
     height="20"
     viewBox="0 0 21 20"
+    focusable="no"
+    aria-hidden="true"
   >
     <path
       id="SvgjsPath1022"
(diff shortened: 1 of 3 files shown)


avrilpearl:
@apalmer0 --> This is awesome - so happy to see someone outside of Digital Services make a PR into the react side! Quick note: Can you add a changelog to your PR? We use that for release notes.

Add the changelog to the changlelog folder in the root of this repo. It should be a markdown file. Your changelog should look something like this:

Patch
Fixed
- (React) [SocialLinks]: Prevent SVGs in SocialLink from receiving keyboard focus. #709

apalmer0:
@avrilpearl Sure thing! What should the changelog title be? Looks like other changes use their ticket number but I don't have one of those yet.

avrilpearl:
@apalmer0 --> I have made a ticket in our internal system to track this --> DP-15599
Thanks!
Sienna

avrilpearl:
great thanks @apalmer0 --> sending it over the our a11y expert @ygannett for initial review

react/src/components/molecules/SocialLinks/index.js
```diff
@@ -70,6 +76,8 @@ const icons = {
     width="21"
     height="17"
     viewBox="0 0 21 17"
+    focusable="no"
```
ygannett: Please use the value, **false**, instead of **no**.  This applies to the all changes.
`focusable="no"` still gets focus depending on the combination of the OS and IE versions.

ygannett (changes requested):
Please see below comment.

react/src/components/molecules/SocialLinks/index.js
apalmer0: Cool thanks! Will do

clairesunstudio (approved):
Looks great, will merge it in. This is exciting! Thanks, @apalmer0 for making the pr! 
FYI, We do weekly releases at noon on Wednesdays. Please join the [Mayflower Slack](https://mayflowerdesignsystem.slack.com/join/shared_invite/enQtNTc0MjMwMjcwMzU0LWRkNGEzZjNiNjEwNDZlOWYxZWRmYTY0NWIxOTNkZTA3NzNhN2U5OTk0OGUzYjE4NjhhNDRiNDhhOGJmZDBjNTM) to get updates. 

Out of curiosity, what project are you working on? Would love to learn more about it and your experience using Mayflower. Let's chat in Slack?

clairesunstudio:
We might convert these social media icons to use our updated [Icon component](https://mayflower-react.digital.mass.gov/?path=/story/atoms-icons--icon) as a followup to leverage shadow DOM. Will keep you updated but this should not affect your implementation.

ygannett (approved):
@apalmer0 Thank you for making the change.
It's ready to merge.

avrilpearl:
@apalmer0 --> this has been merged and can be included in our standard Wednesday releases if that works on your end. We can notify you when the new package is published with this fix.

apalmer0:
Oh awesome, thanks for jumping on this - cool to see it get merged so fast! I just joined your slack channel, I'll chat with you there!

PR #711, opened 2019-08-19
DP-15613: Only use single nav element in footer

apalmer0:
## Description

A second issue raised in the accessibility audit referenced in #709 was that:

> Assistive technology users such as screen reader users have the ability to navigate the page via landmarks or bring up the list of landmarks. They may be confused if they see several \<nav> regions without a unique identifier and may be confused as to what they contain.

The recommendation was to:

> use a single \<nav> element to house all the footer links and give it a aria-label of footer for instance or any relevant identifying name instead of breaking the footer into different \<nav> regions.

To be fair the consultant also added that this:

> is potentially not an accessibility issue but I would still keep it or downgrade it to an advisory.

So this might not be worth incorporating, but I wanted to try to address it.

So to that end, this PR moves each of the semantic HTML tags up a level:

component|old HTML element|new HTML element
---|---|---
`FooterLinksNav`|nav|div
`FooterLinks`|section|nav
`Footer`|div|section

The `<nav>` tag that was previously on the `FooterLinksNav` component has been replaced by a `<div>`, and the `<nav>` is now on `FooterLinks` component. The `<section>` tag that was on the `FooterLinks` component is now on the `Footer` component that renders the `FooterLinks`. 

This also removes the `aria-labelledby` attribute that was previously on the `FooterLinksNav` component, 1) because this is now a `div` instead of a `nav`, which 

> may not work as expected as the \<div> without any roles is a native presentation element.

and 2) because the value provided by the FooterLinks Live data is just the `id` of that data, so the consultant recommended:

> changing the off-screen header text to something more meaningful providing context to the three different footer link sections

One option for this would be the `heading` prop of that data.

## Related Issue / Ticket

- [DP-15613](https://jira.mass.gov/browse/DP-15613) (I think this is the link...)

## Steps to Test

1. I believe since this is just a change in which semantic HTML tags are used, there shouldn't be any behavior or UI change.

## Screenshots
Screen Shot 2019-08-19 at 10 52 05 AM

## Additional Notes:

Anything else to add?

As before, this accessibility stuff is new to me so I'm kinda just parroting what the accessibility guys told me. Happy to defer to someone else with more experience here.

* The `Footer` component, specifically the Footer links.

#### @TODO

* NA.

#### Today I learned...

This accessibility stuff is new to me so pretty much all of it is stuff that I'm learning. This might be old news for others.


avrilpearl:
@apalmer0 --> here is a ticket number for ya DP-15613

react/src/components/molecules/FooterLinks/index.js
```diff
@@ -4,19 +4,19 @@ import './style.css';
 import { SidebarHeading } from '../../../index';
 
 const FooterLinks = (footerLinks) => (
-  <section className="ma__footer-links">
```
ygannett, line 7: Please add the heading above the `nav.ma__footer-links` as 
```<h2 class="ma__visually-hidden" id="footerMenu">Site Navigation Links</h2>```
-  A heading is required for `<section>`.
- The heading tells what the section/navigation is for screen reader users. 
- The `id` is for `aria-labelledby` in the followed `<nav>` to create the association.
- The heading text is not necessary to be hardcoded as shown.  It can be a prop.

File: react/src/components/molecules/FooterLinks/index.js [hunk 2]
```diff
@@ -4,19 +4,19 @@ import './style.css';
...
+  <nav className="ma__footer-links" aria-label="footer links">
```
ygannett: Please remove the `aria-label` and use `aria-labelledby="footerMenu"` instead. The `aria-labelledby`'s value matches the id of the `<h2>` above.

react/src/components/molecules/FooterLinks/index.js
```diff
@@ -4,19 +4,19 @@ import './style.css';
...
       ))
     }
-  </section>
+  </nav>
 );
 
 
 const FooterLinksNav = (footerLinksNav) => (
-  <nav aria-labelledby={footerLinksNav.id} className={`ma__footer-links__items${footerLinksNav.showNavHeading ? ' ma__footer-links__items--heading' : ''}`}>
+  <div className={`ma__footer-links__items${footerLinksNav.showNavHeading ? ' ma__footer-links__items--heading' : ''}`}>
```
ygannett, line 22: The `<div>` is not necessary. Please remove it.  The `<ul>`s don't have a wrapper.

react/src/components/molecules/FooterLinks/index.js
```diff
@@ -4,19 +4,19 @@ import './style.css';
...
     {
       footerLinksNav.showNavHeading ? <SidebarHeading title={footerLinksNav.heading} level={2} /> : <h2 className="visually-hidden" id={footerLinksNav.id}>{footerLinksNav.heading}</h2>
```
ygannett: Please use `<h3>` instead of `<h2>` since we adds `<h2>` for the `section.ma__footer__nav`.
The heading doesn't need `id` since no need for the association with the removed `<nav>` anymore with `aria-labelledby`.  Please remove the `id`.

react/src/components/molecules/FooterLinks/index.js
```diff
@@ -28,7 +28,7 @@ const FooterLinksNav = (footerLinksNav) => (
         ))
       }
     </ul>
-  </nav>
+  </div>
```
ygannett, line 34: Since we don't wrap `<ul>`,  please remove this `</div>`.

ygannett (changes requested):
@apalmer0, please see the following comments. Please let me know for any questions.

react/src/components/molecules/FooterLinks/index.js
apalmer0: Hi @ygannett - could you clarify here? There's a classname on this wrapper, so if I remove the `<div>` I don't know where I'd put that classname. I assume I'd wrap the `<SidebarHeading>`/`<h3>` and the `<ul>` with a `<React.Fragment>`, but I don't believe I can put a classname on that.

react/src/components/molecules/FooterLinks/index.js
apalmer0: Hi @ygannett - regarding your last comment about the heading text being a prop, would that be a new prop on the `Footer` component, or would we just reuse a different one?

ygannett: Hi @apalmer0, thank you for doing all improvement.
Currently, no equivalent heading and its value exist.  So, it's new. I figured, it's more beneficial to give flexibility to name the heading to better meet their own purpose.

I assume you're thinking about `h2#FooterLinks1/2/3.visually-hidden` as the first child of `nav.ma__footer-links__items`. This heading is for the following `<ul>` even its associated with the `nav` via `aria-labelledby`.  This section is not marked up properly, so it looks confusing to explain this.

Currently, the section has 3 units of the navigation as you see:
- `div.ma__footer__nav` > `section.ma__footer-links` >
   - Unit 1: `nav.ma__footer-links__items` + `h2.#FooterLinks1` + `ul`
   - Unit 2: `nav.ma__footer-links__items` + `h2.#FooterLinks2` + `ul`
   - Unit 3: `nav.ma__footer-links__items` + `h2.#FooterLinks3` + `ul`

We're going to change them to 1 unit of navigation with 3 units of link groups(= `ul`):
- `section.ma__footer__nav` > `h2#footerMenu.ma__visually-hidden` > `nav` >
       - Unit 1:  `h3.ma__visually-hidden` + `ul.ma__footer-links__items`
       - Unit 2:  `h3.ma__visually-hidden` + `ul.ma__footer-links__items`
       - Unit 3:  `h3.ma__visually-hidden` + `ul.ma__footer-links__items`
In the new markup, the `h3`s don't need `id` since they are not associated with `nav`.

Let me know if you have any questions.

react/src/components/molecules/FooterLinks/index.js
ygannett: You see I put the `class` on `nav` on `ul` in above comment. Since `h3` is not visible, this doesn't cause any display issue.  `ul.ma__footer-links__items` would maintain the current positioning. You can say that the `ul` turns the wrapper used to be `nav`. FYI, the `ul`s don't have `aria-labelledby`.

I'm looking at output markup and not really looking from React side.  If this would cause any troubles in React, please let me know. I'm not so familiar with React, so I might not be aware of certain things from its aspect. I apologize for that, but please don't hesitate to point them out.

ygannett:
@apalmer0 Hi, just checking in.  Please let me know if you have any questions and concerns for my comments.

apalmer0:
Hi @ygannett! Sorry for the delay - this is still on my radar, just had some stuff come up that needed to take priority. I'll have a feedback commit pushed up asap!

ygannett:
@apalmer0 No worry. Just wanted to make sure if my comments are blocking you and see anything I can do.

PR #712, opened 2019-08-19
React/Update socialLinks to use Icon and followup accessibility for SR

clairesunstudio:
This pr 
1. Removes hard coded svgs in the SocialLinks component and added social media icons into Icon component options. 

2. Added option to set icon fill color
Screen Shot 2019-08-19 at 5 14 47 PM

3. Added `arial-label` to SocialLinks `<a>` to follow best practices for Screen Readers


react/src/components/molecules/SocialLinks/index.js
```diff
@@ -17,116 +18,27 @@ const SocialLinks = (socialLinks) => (
...
-  </svg>
-};
-
 const SocialLink = (socialLink) => (
   <li className="ma__social-links__item">
     <a
       href={socialLink.href}
       className="ma__social-links__link js-social-share"
       data-social-share={socialLink.linkType}
       title={socialLink.altText}
```
ygannett: Please remove `title` which is replaced with `aria-label`.

react/src/components/atoms/icons/Icon/index.js
```diff
@@ -9,7 +9,8 @@ export default class Icon extends React.Component {
     svgWidth: PropTypes.number,
     svgHeight: PropTypes.number,
     classes: PropTypes.arrayOf(PropTypes.string),
-    ariaHidden: PropTypes.bool
+    ariaHidden: PropTypes.bool,
+    fill: PropTypes.string
```
ygannett, line 13: Any reason to use `fill` directly instead of via css?

ygannett (changes requested):
@clairesunstudio Please see the following comments.

react/src/components/molecules/SocialLinks/index.js
clairesunstudio: `title` shows more information to the non-SR users, e.g. when mouse over the icon, it explains what the icon is, which is not completely replaceable by aria-label. I wonder here if there's any harm to SR users to have both

react/src/components/atoms/icons/Icon/index.js
clairesunstudio: So `fill` is fully ecapsulate into the svg while css classes change the style from outside. If an svg has a fill attribute already, it css doesn't work. This is will make icon styling more componentized.

react/src/components/molecules/SocialLinks/index.js
ygannett: If they are in a condition to be able to see the title value, users would benefit from that.
Here is what I mentioned as the inconsistent implementation of `title`...
The [W3C HTML5 Recommendation](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute) warns against use of the title attribute:

> Relying on the title attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

When both, `title` and `aria-label` are added, screen reader could read both values depending on combinations of OS and devices.  Screen reader users would listen the same content twice, which is not so good user experience.

react/src/components/atoms/icons/Icon/index.js
ygannett: Right.  By adding `fill` directly in `svg` here, you don't want the icons to set different colors from outside (= css) even they are used somewhere else?

react/src/components/molecules/SocialLinks/index.js
clairesunstudio: I see. Are you good with removing `title`? The only thing would be there will not be a tooltip to explain the icons

react/src/components/atoms/icons/Icon/index.js
clairesunstudio: it should be able to override --> https://codepen.io/team/css-tricks/pen/yebOYp

react/src/components/molecules/SocialLinks/index.js
ygannett: It doesn't seem any reasonable alternative is available right now, but we'll continue research for non-screen reader users.

react/src/components/atoms/icons/Icon/index.js
ygannett: Ok, let's go with this change.

ygannett (approved):

PR #722, opened 2019-08-26
add circle auto changelog

clairesunstudio:
As we are going to automate our release process. Auto consolidate changelogs depends on all prs have a changelog that follows certain standards. This pr adds a circle task to automate adding changelog to dependabot prs
Screen Shot 2019-08-27 at 5 48 52 PM


avrilpearl:
(1) Can we label these [Dependency]
(2) Can you separate Security ones in type: Security and straight-up non-security related fixes in type: Changed.

avrilpearl (changes requested):
see comment above re:type and component label

.circleci/config.yml
```diff
@@ -219,12 +219,30 @@ jobs:
...
+      - checkout
+      - *no_host_check
+      - add_ssh_keys:
+          fingerprints:
+            - "a7:d5:55:42:fe:37:52:b6:71:7c:a6:d8:05:78:fd:25"
+      - *configure_git
+      - run:
+          name: 'Auto add changelog for dependabot'
+          command: |
+            chmod +x ./.circleci/dependabot.sh
```
mrossi113: @clairesunstudio you can `chmod +x ./.circleci/dependabot.sh` and then commit to the PR instead.

.circleci/config.yml
```diff
@@ -219,12 +219,30 @@ jobs:
...
           fi
 
+  auto_changelog:
+    <<: *react_defaults
+    steps:
+      - checkout
+      - *no_host_check
+      - add_ssh_keys:
+          fingerprints:
+            - "a7:d5:55:42:fe:37:52:b6:71:7c:a6:d8:05:78:fd:25"
```
mrossi113: This could be an environment variable in CircleCI instead adding directly in  the config.yml file.

.circleci/dependabot.sh
```diff
@@ -0,0 +1,22 @@
...
+if [[ $lastCommit =~ "/react" ]]; then
+  lastCommit="(React) $lastCommit"
+fi
+if [[ $lastCommit =~ "/patternlab" ]]; then
+  lastCommit="(Patternlab) $lastCommit"
+fi
+
+if [ "$lastCommit" = "$lastDependabot" ]
+then
+  echo "- [Dependencies] $lastCommit" >> changelogs/dependabot.md
```
mrossi113: Is all dependbot PRs using only one changelog?

mrossi113 (commented):
I just have a few comments.

.circleci/config.yml
clairesunstudio: good call will add that to circle

.circleci/config.yml
clairesunstudio: the push to pr part is in the shell script

.circleci/dependabot.sh
clairesunstudio: yes, they will all be in one file. do you want them to be separate? based on our conversation last week I thought there would be no difference on your end.

mrossi113: No, that is fine I was just trying to understand this PR.

.circleci/config.yml
mrossi113: Right, but you want to make the shell script executable correct?

clairesunstudio: yep, this is the line to make it executable

avrilpearl (approved):

mrossi113 (approved):
👍 LGTM!

PR #730, opened 2019-09-05
React/DP-15689:  Extend ButtonWithIcon & Add LinkDropdown

avrilpearl:
Minor

Changed
- (React) [ButtonWithIcon] DP-15689: Extended the ButtonWithIcon component to pass multiple theme (c-primary, c-primary-alt, c-highlight, c-gray-dark) and display for multiple usages per base button atom (primary, secondary, tertiary, quaternary, quaternary-simple). #730
- (React) [DecorativeLink] DP-15689: Enabled classnames to be passed to DecorativeLink. #730
- (React) [ListingTable] DP-15689: Updated ListingTable to consume ButtonWithIcon for expand functionality. #730
- (React) [ButtonAlert] DP-15689: Made buttonalert a variant of buttonwithicon. #730

Added
- (React) [LinkDropdown] DP-15689: Created LinkDropdown molecule. #730

Fixed
- (React) [Button, ButtonWithIcon] DP-15689: Fixed color contrast a11y issue of c-highlight theme. #730


assets/scss/00-base/mixins/_ma-button.scss
```diff
@@ -33,7 +33,13 @@
     @if($type == "solid") {
       background-color: $color;
       border-color: transparent;
-      color: $c-font-inverse;
+
+      @if $color == $c-highlight {
```
clairesunstudio, line 37: thanks for fixing this! I just realized that the secondary highlight button is also not accessible. but I don't think it looks good with black text either.
Screen Shot 2019-09-06 at 3 33 06 PM
Screen Shot 2019-09-06 at 3 31 17 PM

we could add a darker yellow color into the design system. e.g.
Screen Shot 2019-09-06 at 3 36 22 PM

my recommendation would be to remove that option from secondary. WIll leave up to you and @ygannett to decide

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -45,24 +89,149 @@
     vertical-align: middle;
     fill: $c-gray-dark;
```
clairesunstudio:
```suggestion
```
currently this class is interfering iconColor setting

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -1,10 +1,54 @@
...
 
+$colors: (
+  "c-highlight": $c-highlight,
+  "c-primary": $c-primary,
+  "c-primary-alt": $c-primary-alt,
+  "c-black": $c-black,
+  "c-gray-dark": $c-gray-dark,
+  "c-white": $c-white
+);
+@mixin ma-button-icon-secondary ($color: $c-primary-alt) {
```
clairesunstudio: can you move these @mixins into mixins folder? just so it stays consistent with _buttons.scss

maybe see if these make sense to be somehow combined in the _ma-button mixin (just a thought)

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -1,10 +1,54 @@
...
+  border-bottom-color: rgba($color,0.75);
+  color: $color;
+  svg {
+    fill: rgba($color,0.75);
+  }
+  &:hover {
+    border-bottom-color: $color;
+  }
+}
+@mixin ma-button-icon-quaternary-simple ($color: $c-primary-alt) {
```
clairesunstudio: In regular button component, quaternary is hard coded to only render theme `$c-primary` (all the other theme colors don't work) 
I think we should extend the buttons to allow button quarternary to use more themes except for hightlight, as currently they don't work in all quarternary usage for accessiblity
Screen Shot 2019-09-06 at 3 55 15 PM
Screen Shot 2019-09-06 at 3 55 18 PM

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -45,24 +89,149 @@
     vertical-align: middle;
     fill: $c-gray-dark;
     height: 20px;
-    transition: fill .4s;
+    transition: transform .5s, fill .4s;
   }
 
-  &.ma__icon-green > svg {
-    fill: $c-primary-alt;
+  @each $class, $c in $colors {
```
clairesunstudio: i'm wondering if setting the icon svg color separately, as I feel like the icon should be aligned with the color of the button text and set by the theme of the button in the mixin as a whole (including hover state). I can't think of a case it's should be different on the top of my head. e.g. tertiary button if themes can't be changed, icon color shouldn't be able to change either

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -45,24 +89,149 @@
...
+  }
+
+  &--quaternary, &--expanded.ma__button-icon--quaternary, &:hover.ma__button-icon--quaternary {
+    min-height: 30px;
+    background-color: transparent;
+    box-shadow: none;
+    border: none;
+    padding: 0;
+    text-transform: none;
+    border-bottom: 2px solid rgba($c-primary-alt,0.75);
```
clairesunstudio: setting the default here is not neccessary, as in the mixin you already set default by passing `$color: $c-primary-alt`

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
clairesunstudio: can we set the default to $c-primary instead
```suggestion
@mixin ma-button-icon-quaternary-simple ($color: $c-primary) {
```

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -1,10 +1,54 @@
...
+    border-bottom-color: $color;
+  }
+}
+@mixin ma-button-icon-quaternary-simple ($color: $c-primary-alt) {
+  color: $color;
+  svg {
+    fill: rgba($color,0.5);
+  }
+}
+@mixin ma-button-icon-alert ($color: $c-primary-alt) {
```
clairesunstudio: can we set the default to $c-black instead
```suggestion
@mixin ma-button-icon-alert ($color: $c-black) {
```

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
```diff
@@ -1,10 +1,54 @@
...
+  svg {
+    fill: $color;
+  }
+  &:hover {
+    svg {
+      fill: $c-white;
+    }
+  }
+}
+@mixin ma-button-icon-quaternary ($color: $c-primary-alt) {
```
clairesunstudio: can we set the default to $c-primary instead
```suggestion
@mixin ma-button-icon-quaternary ($color: $c-primary) {
```

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
clairesunstudio: also applied to all usage, don't need to set default color as its already set in mixin

react/src/components/molecules/LinkDropdown/index.js
```diff
@@ -0,0 +1,118 @@
...
+    this.setState((prevState) => ({ buttonExpand: !prevState.buttonExpand }));
+  }
+  handleSelect(event) {
+    const { target } = event;
+    this.setState({ buttonExpand: false });
+    if (is.fn(this.props.onItemSelect)) {
+      this.props.onItemSelect({ target });
+    }
+  }
+  handleClickOutside(event) {
```
clairesunstudio: should escape key also close dropdown?

react/src/components/molecules/LinkDropdown/index.js
```diff
@@ -0,0 +1,118 @@
...
+    if ((node && !node.contains(event.target))) {
+      if (this.state.buttonExpand) {
+        this.setState({ buttonExpand: false });
+      }
+    }
+  }
+  render() {
+    const { dropdownItems, dropdownButton } = this.props;
+    const dropdownButtonProps = {
+      ...dropdownButton,
```
clairesunstudio, line 67: should `'aria-haspopup': ` in this component also be true, without the option to overwrite?

react/src/components/molecules/LinkDropdown/LinkDropdown.stories.js
```diff
@@ -0,0 +1,33 @@
...
+storiesOf('molecules', module)
+  .addDecorator(withKnobs({ escapeHTML: false }))
+  .add('LinkDropdown', (() => {
+    const props = {
+      dropdownButton: {
+        text: text('text', 'Add to calendar', 'dropdownButton'),
+        theme: select('theme', buttonWithIconOptions.theme, 'c-primary', 'dropdownButton'),
+        usage: select('usage', buttonWithIconOptions.usage, 'quaternary-simple', 'dropdownButton'),
+        id: text('id', 'dropdownbutton-simple', 'dropdownButton'),
+        'aria-haspopup': boolean('aria-haspopup', true, 'dropdownButton'),
```
clairesunstudio: shouldn't this always be true?

react/src/components/molecules/LinkDropdown/index.js
```diff
@@ -0,0 +1,118 @@
...
+      'ma__link-dropdown-item': true,
+      'ma__link-dropdown-item--expanded': this.state.buttonExpand
+    });
+
+    return(
+      <div ref={this.wrapperRef} className={dropdownClasses}>
+        <ButtonWithIcon {...dropdownButtonProps} />
+        {this.state.buttonExpand && (
+          <div className={dropdownMenuClasses} aria-labelledby={dropdownButtonProps.id}>
+            {dropdownItems.map((item) => <DecorativeLink {...item} className={dropdownItemClasses} onClick={this.handleSelect} />)}
```
clairesunstudio: this handleSelect function will never get triggered because onClick is not a prop for DecorativeLink. If we want to pass a custom click function here we should probably use Button

clairesunstudio:
linkDropDown hover style is always primary alt, even a different theme is set. the text color on hover should probably match the theme selected
Screen Shot 2019-09-06 at 5 13 00 PM

should we move buttonWithIcon styles into assets folder?

clairesunstudio:
Great work! Left quite a few comments above don't hate me 😛You refactored a bunch of things 👍  !

react/src/components/molecules/LinkDropdown/index.js
avrilpearl: always set to true

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: Please see the response to mixin comment above, do to this I cannot change the default theme away from primary-alt

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: @clairesunstudio --> not true here since default theme is c-primary-alt that theme class is not appended so I need to set the default with no theme class to c-primary-alt. Unless we want to change this but it is the same way the base button is doing it (buttons.scss), where with no class it is styled in the theme c-primary.

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: see the response to mixin comment above re: default color.

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: see the response to mixin comment above re: default color.

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: ?

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: removed mixins here.

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: removed option for quaternary to use c-highlight theme

react/src/components/atoms/buttons/ButtonWithIcon/style.scss
avrilpearl: removed option to style icon with icon color prop

assets/scss/00-base/mixins/_ma-button.scss
avrilpearl: updated to black - will wait to hear from yukiko to finish this

ygannett: I think it's better to remove the option.  Yellow is a tricky color. I looked through the other options.  Their pattern is dark text on a lighter version of the theme colors at normal state, then white text on the theme colors. Even with a darker yellow for the text, we cannot follow this pattern. It would look off from the pattern.

ygannett (commented):
Looks good!
Please see my comment for assets/scss/00-base/mixins/_ma-button.scss.

react/src/components/molecules/LinkDropdown/index.js
avrilpearl: added this.

ygannett (approved):
Looks good.

PR #739, opened 2019-09-13
DP-15731 MF create list styles for the "campaign-full-width-header-not-top" organism v 0.1

lannyheidbreder-current:
## Description
Added ability for WYSIWYG unordered lists to be given classes. Implemented .ma__horizontal-list class. Used vanilla JS to prevent widows.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-15731)

#### Go here and do this

Go to `?p=organisms-key-message-not-top` in Mayflower

#### What you should see

A horizontal list of items, styled to match the spec. The last row should never have a single item, even after resizing the window. (It may take half a second after resizing for the items to reflow.)

## Screenshots

Screen Shot 2019-09-13 at 4 21 59 PM

- unordered-list atom
- key-message-not-top organism


merauluka (approved):
These changes look good to me!

clair0917 (approved):
LGTM

PR #745, opened 2019-09-18
DP-15856 address filtering not accurate

sumeetpareek:
## Description

- When user's entered city names in location listing pages, the map did not zoom to expected area, even if there were actual location items in the entered city area.
- After a lot of debugging on the drupal and mayflower side, we found the root cause was that google maps API geocoded entered city in a global fashion (for example: it thought `cambridge` was the `cambridge, UK` city).
- We tried to correct with my using `bounds` as we do with autocomplete, but that didn't work every time either. So we have a combined solution now.

## Related Issue / Ticket

- https://jira.mass.gov/browse/DP-15856

### === 😢😭 BEFORE  😢😭 ===

Follow numbered steps from below image

see_ma_cities_in_maps_not_working

1. On `develop` branch ... goto `mayflower/patternlab/styleguide/` ... run `gulp` ... and then open this URL in your browser = http://localhost:3000/?p=pages-location-listing (refresh the page)
1. From chrome dev tools open the sources tab
1. Click on the three-dots-menu and open the file `locationListing.js`
1. Set a breakpoint in the `sortDataAroundPlace()` function towards the end by which point sorting is already done.
1. On the webpage enter the cityname `cambridge` and dismiss any autocomplete suggestions that come up by pressing ESC
1. Click submit button to filter
1. Notice that the breakpoint is hit and google geocodes your input as an unexpected city 😢😭 

NOTE: Also see the order of listing on the webpage after you release the breakpoint hold.

### === 😃😃 AFTER  😃😃 ===

Follow numbered steps from below image

ensure_ma_cities_in_maps_work

1. Checkout this branch that is `DP-15856-address-filtering-not-accurate` branch ... goto `mayflower/patternlab/styleguide/` ... run `gulp` ... and then open this URL in your browser = http://localhost:3000/?p=pages-location-listing (refresh the page)
1. From chrome dev tools open the sources tab
1. Click on the three-dots-menu and open the file `locationListing.js`
1. Set a breakpoint in the `sortDataAroundPlace()` function towards the end by which point sorting is already done.
1. On the webpage enter the cityname `cambridge` and dismiss any autocomplete suggestions that come up by pressing ESC
1. Click submit button to filter
1. Notice that the breakpoint is hit things work as you would expect 😃😃

NOTE: Also see the order of listing on the webpage after you release the breakpoint hold.

## Screenshots

## Additional Notes:

Anything else to add?

* 

#### @TODO

*

### DIFF
--- a/patternlab/styleguide/source/assets/js/modules/eventListingInteractive.js
+++ b/patternlab/styleguide/source/assets/js/modules/eventListingInteractive.js
@@ -240,8 +240,14 @@ export default (function (window,document,$,undefined) {
       }
       // If place argument was populated from locationFilter (zipcode text input) but not from Place autocomplete.
       else {
+        // When users just enter a cityname, google's `geocoder` inside the `geocodeAddressString` is not precise
+        // even with 'bounds' parameter set, so we add " MA, USA" to acheive desired accuracy.
+        // NOTE: This is NOT done when autocomplete result is selected by user, in which case the above `if` situation is executed.
+        var user_entered_location_string = place;
+        var user_entered_location_parts = user_entered_location_string.split(",");
+        var reasonable_formatted_location = user_entered_location_parts.pop() + " MA, USA";
         // This is an asynchronous function
-        listings.geocodeAddressString(place, function(result) {
+        listings.geocodeAddressString(reasonable_formatted_location, function(result) {
           transformReturn.data = sortDataAroundPlace(result, filteredData);
           transformReturn.geocode = result;
           // Return the data sorted by location and the geocoded place object
--- a/patternlab/styleguide/source/assets/js/modules/locationListing.js
+++ b/patternlab/styleguide/source/assets/js/modules/locationListing.js
@@ -337,8 +337,14 @@ export default (function (window, document, $) {
       }
       // If place argument was populated from locationFilter (zipcode text input) but not from Place autocomplete.
       else {
+        // When users just enter a cityname, google's `geocoder` inside the `geocodeAddressString` is not precise
+        // even with 'bounds' parameter set, so we add " MA, USA" to acheive desired accuracy.
+        // NOTE: This is NOT done when autocomplete result is selected by user, in which case the above `if` situation is executed.
+        var user_entered_location_string = place;
+        var user_entered_location_parts = user_entered_location_string.split(",");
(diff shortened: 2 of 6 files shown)


patternlab/styleguide/source/assets/js/modules/locationListing.js
```diff
@@ -323,8 +337,14 @@ export default (function (window, document, $) {
       }
       // If place argument was populated from locationFilter (zipcode text input) but not from Place autocomplete.
       else {
+        // When users just enter a cityname, google's `geocoder` inside the `geocodeAddressString` is not precise
+        // even with 'bounds' parameter set, so we add " MA, USA" to acheive desired accuracy.
+        // NOTE: This is NOT done when autocomplete result is selected by user, in which case the above `if` situation is executed.
+        var user_entered_location_string = place;
+        var user_entered_location_parts = user_entered_location_string.split(",");
+        var reasonable_formatted_location = user_entered_location_parts.pop() + " MA, USA";
```
merauluka, line 345: @sumeetpareek If we are open sourcing this, then are we concerned about forcing the map to be within Massachusetts? Should we not allow this to be configurable in some way?

patternlab/styleguide/source/assets/js/modules/eventListingInteractive.js
```diff
@@ -240,8 +240,14 @@ export default (function (window,document,$,undefined) {
       }
       // If place argument was populated from locationFilter (zipcode text input) but not from Place autocomplete.
       else {
+        // When users just enter a cityname, google's `geocoder` inside the `geocodeAddressString` is not precise
+        // even with 'bounds' parameter set, so we add " MA, USA" to acheive desired accuracy.
```
merauluka, line 244: @sumeetpareek Mispelled "achieve" here.

And same question as below. I think we should attempt to make the " MA, USA" configurable in some way, or surface it to the end user. Otherwise it will always defer to returning addresses from within Massachusetts even if the open source user does not wish this behavior.

merauluka (approved):
Approved now that the follow up task to determine how to handle region restrictions to prevent this open-source project from becoming Mass. centric.
https://jira.mass.gov/browse/DP-15968

mrossi113 (approved):
👍 LGTM! External reviewed this PR.

PR #797, opened 2019-10-15
DP-16106: Fix gen teaser bug

avrilpearl:
Patch
Fixed
(React) [GenTeaser] DP-16106: Fix import statement bug resulting in jest suite failure on consumer side of mayflower-react package. #797


clairesunstudio (approved):
looks good!

PR #828, opened 2019-11-13
react/DP-14222: Makes it possible to render render with/without UtilityNav and/or MainNav content

avrilpearl:
Minor
Changed
- (React) [Header] DP-14222: Enable Header to render with and without UtilityNav and MainNav content.


clairesunstudio:
Works well, the only thing i noticed is that if we remove both mainNav and utilNav the mobile hamburger menu is still rendering:
Screen Shot 2019-11-15 at 3 37 48 PM
Screen Shot 2019-11-15 at 3 37 53 PM
 
It'd be nice also if mainNav and utilNav are both not passed, to remove/decrease `padding-top` on `ma__header`

clairesunstudio (approved):

PR #829, opened 2019-11-15
React/DP-12195: Address AccordionItem a11y issue

avrilpearl:
Minor
Fixed
- (React) [AccordionItem] DP-12195: Adds `id` prop to accordion item component to address a11y issues. #829

### DIFF
--- a/react/src/components/molecules/AccordionItem/index.js
+++ b/react/src/components/molecules/AccordionItem/index.js
@@ -16,7 +16,7 @@ class AccordionItem extends React.Component {
   }
 
   handleClick() {
-    const open = this.state;
+    const { open } = this.state;
     this.setState({
       open: !open
     });
--- a/react/src/components/molecules/AccordionItem/index.js
+++ b/react/src/components/molecules/AccordionItem/index.js
@@ -24,7 +24,7 @@ class AccordionItem extends React.Component {
 
   render() {
     const {
-      buttonId, secondary, border, emphasize, icon, title, headerLevel, children, expandableSectionId, info
+      secondary, border, emphasize, icon, title, headerLevel, children, id, info
     } = this.props;
     const { open } = this.state;
     const accordionClasses = classNames({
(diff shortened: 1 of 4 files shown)


react/src/components/molecules/AccordionItem/index.js
```diff
@@ -16,54 +16,61 @@ class AccordionItem extends React.Component {
...
               </div>
             )}
-            { this.props.secondary && (
+            { secondary && (
               <div className="ma__accordion-header__icon--secondary">
                 <Icon name="chevron" svgHeight={20} svgWidth={20} />
               </div>
             )}
-            <Heading class={headingClasses} text={this.props.title} level={this.props.headerLevel} />
+            <Heading class={headingClasses} text={title} level={headerLevel} />
```
clairesunstudio, line 68: I wonder if the `aria-labelledby` id should be on the heading instead of the button which contains the label text. I'm not sure SR can read `aria-label` on an element

clairesunstudio: i haven't found an example that the referenced id is on the outter div with aria-label

react/src/components/molecules/AccordionItem/index.js
```diff
@@ -88,7 +95,11 @@ AccordionItem.propTypes = {
   /** Whether to style the accordion as secondary or not. */
   secondary: PropTypes.bool,
   /** Heading level for accordion title */
-  headerLevel: PropTypes.number
+  headerLevel: PropTypes.number,
+  /** Expand button id, used for a11y */
+  buttonId: PropTypes.string,
```
clairesunstudio: can we only ask for one ID prop and inside the component form the two different ids? 
e.g.  something like:
```
prop: accordionID
expandableSectionId = `expandable${accordionID}`
labelId = `label${accordionID}`
```

clairesunstudio:
good catch! just the two comments above

clairesunstudio (dismissed):
`id` and `aria-labelledby` are not passed into the html div of `Transition` in `Collapse`
Screen Shot 2019-11-15 at 1 19 57 PM

react/src/components/molecules/AccordionItem/index.js
avrilpearl: based on w3 --> https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html it should be on the button level event if nested title

avrilpearl:
> `id` and `aria-labelledby` are not passed into the html div of `Transition` in `Collapse`
> Screen Shot 2019-11-15 at 1 19 57 PM

@clairesunstudio --> this is because the story was not updated

clairesunstudio (approved):

PR #836, opened 2019-11-25
(React) DP-11777: Enable focus in a tab's content

avrilpearl:
Minor
Fixed
- (React) [TabContainer] DP-11777: Enable focus in tab container by removing arrowdown and arrowup mgmt at the tab level. (#836)


clairesunstudio:
I think we should remove the handleKeyDown event in TabContainer/index.js instead

clairesunstudio (approved):

PR #839, opened 2019-12-03
React / change footnotelink ref id to classname

clairesunstudio:
[DP-11765](https://jira.mass.gov/browse/DP-11765)

To test: 
In Mayflower
- [ ] pull down this branch and run `npm i && npm run build` inside of `react/`
- [ ] run yarn link

in Budget
- [ ] in budget repo root, run `yarn link-mayflower`
- [ ] run `yarn develop:govbudget`

### DIFF
--- a/react/src/components/atoms/links/FootNoteLink/index.js
+++ b/react/src/components/atoms/links/FootNoteLink/index.js
@@ -13,7 +13,7 @@ class FootNoteLink extends React.Component {
       <button
         className={`ma__footnote-link footnoteref${index}`}
         aria-describedby={`footnotemsg${index}`}
-        aria-label="see footnote"
+        aria-label={`see footnote ${index}`}
         onClick={() => this.handleScroll()}
         tabIndex="0"
         role="link"
--- a/react/src/components/atoms/links/FootNoteLink/index.js
+++ b/react/src/components/atoms/links/FootNoteLink/index.js
@@ -11,8 +11,7 @@ class FootNoteLink extends React.Component {
     const { index } = this.props;
     return(
       <button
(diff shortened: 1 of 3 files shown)


react/src/components/atoms/links/FootNoteLink/index.js
```diff
@@ -11,8 +11,7 @@ class FootNoteLink extends React.Component {
     const { index } = this.props;
     return(
       <button
-        id={`footnoteref${index}`}
-        className="ma__footnote-link"
+        className={`ma__footnote-link footnoteref${index}`}
         aria-describedby={`footnotemsg${index}`}
         aria-label="see footnote"
```
ygannett: Since aria-label overrides the button label, screen reader users wouldn't know which footnote they'd look at with the button with the current setup.

I suggest its value as "see footnote {index}" to distinguish each in multiple footnotes.  Then, screen reader users would know which footnote they would look at.
```aria-label="see footnote {index}"```

**Review by ygannett — COMMENTED** [review]
The initial change looks good. Below is something noticed during the review.

File: react/src/components/atoms/links/FootNoteLink/index.js (hunk 1 above)
Comment (clairesunstudio) [inline react/src/components/atoms/links/FootNoteLink/index.js]:
```suggestion
        aria-label={`see footnote ${index}`}
```

clairesunstudio: good catch, addressed in commit below

ygannett: Thank you for the quick change.

ygannett (approved):
It looks good!

PR #840, opened 2019-12-03
Dp 16378 release automation test

clairesunstudio:


scripts/release-branch.js
```diff
@@ -0,0 +1,146 @@
+// Added shelljs to use maybe in the script "npm install shelljs"
+// https://github.com/shelljs/shelljs
+const shell = require('shelljs');
+
+
+const Octokit = require("@octokit/rest");
+const octokit = new Octokit();
```
rbayliss: Need to use Auth here: 
```
const octokit = new Octokit({auth: process.env.DANGER_GITHUB_API_TOKEN});
```
I'm not actually sure what format the token is in right now - it should be `token XXX` (note the use of the word token).  See this page for auth details: https://octokit.github.io/rest.js/#authentication

scripts/release-branch.js
```diff
@@ -0,0 +1,146 @@
...
+
+console.log(process.env.DANGER_GITHUB_API_TOKEN)
+
+const owner = 'massgov';
+const repo = 'mayflower';
+const title = 'Release/test';
+const head = 'release/9.34.0';
+const base = 'master';
+
+octokit.pulls.create({
```
rbayliss: This returns a promise. You can either deal with this using `.then()`, or wrap it inside of an `async` function.  For simplicity, let's just go with this:
```
octokit.pulls
  .create({ ... })
  .catch(function() {
    console.error(`There was an error creating the Github PR: ${e.toString()`);
    process.exit(1);
  })
```
Because this is a promise, anything written after this line may be executed _before_ the PR finishes creating, which is a bit confusing.  So either make this the last thing that happens in the script, or put the rest of the stuff that needs to happen inside a `.then()` callback chained off the same function.

.circleci/config.yml
```diff
@@ -309,11 +309,12 @@ workflows:
           requires: [build]
     triggers:
       - schedule:
-          cron: "00 18 * * 2"
+          cron: "55 21 * * 2"
```
```suggestion
          cron: "00 18 * * 2"
```

PR #845, opened 2019-12-09
React/dp 16677 filterbox invalid proptype

clairesunstudio:
https://jira.mass.gov/browse/DP-16677

Fix Filter and TableOfContents prototype checks and remove obsolete util functions for proptype checks.

- (React) [FilterBox, IconLink, PageFlipper, TableofContents] DP-16677: Fix component prop type checks by replacing custom util function `componentPropTypeCheck` with `componentWithName` from 'airbnb-prop-types'. #845


rbayliss (approved):
This looks ok to me. I'm a little unsure of how or where to test, but the fact that it passed tests without any input changes suggests it won't break anything downstream.  I visited the Filterbox and IconLink component, and they both look good.

PR #856, opened 2019-12-11
Fix changelog duplicated key

clairesunstudio:
Each changelog file can only have one change type key. Duplicated key currently will fail. In the follow up we should make change type a list item to allow multiple


clairesunstudio:
Each changelog file can only have one change type key. Duplicated key currently will fail. In the follow up we should make change type a list item to allow multiple

.circleci/config.yml
```diff
@@ -304,7 +304,7 @@ workflows:
           requires: [build]
     triggers:
       - schedule:
-          cron: "00 18 * * 3"
+          cron: "45 18 * * 3"
```
```suggestion
          cron: "00 19 * * 3"
```

.circleci/config.yml
```diff
@@ -304,7 +304,7 @@ workflows:
           requires: [build]
     triggers:
       - schedule:
-          cron: "00 18 * * 3"
+          cron: "00 19 * * 3"
```
```suggestion
          cron: "25 19 * * 3"
```

PR #858, opened 2019-12-11
Sync Master into Develop

clairesunstudio:
9.34.0


clairesunstudio:
@rbayliss Since the last commit on master is a hotfix and the release tag was cut by running the shell script locally due to the failure in the first place. 

Am I understanding this correctly:
Whenever there are hotfixes directly going into master (other than release prs) `ci/circleci: build_github_tag/hold` will run in circle and we should just ignore it

PR #860, opened 2019-12-12
React/dp16678 fix tab ref proptype

clairesunstudio:
https://jira.mass.gov/browse/DP-16678

To test:

In mayflower react
- npm start
- go to http://localhost:6006/?path=/story/organisms-tabcontainer--tabcontainer and inspect


ygannett (approved):
It looks good.  No error message found with the change.

ygannett (approved):
It looks good.  No error found with the change.

PR #900, opened 2020-01-21
Automation/dp 17062 automate pr tag description

clairesunstudio:
- Refactored auto-release scripts to share vars
- Automated adding PR and Tag description
- Convert github-tag shell script with node script

Before: Nothing in the PR body
After:
Screen Shot 2020-01-22 at 9 52 51 AM

Before & After (release description):
Screen Shot 2020-01-22 at 10 46 24 AM


clairesunstudio:
Screen Shot 2020-01-21 at 5 09 09 PM

await release branch push not ready??

scripts/consolidate-changelogs.js
```diff
@@ -0,0 +1,32 @@
+const path = require('path');
+const fs = require('fs');
+
+const directoryPath = path.resolve(__dirname, '../changelogs');
+const changelogPath = `${path.resolve(__dirname, '../')}/CHANGELOG.md`;
+const tempLogsPath = `${path.resolve(__dirname, '../')}/tempLogs.json`;
```
rbayliss: It bugs me that this script has an unseen dependency on the `compile-changelogs.js` script.  I mean, it should work, but only as long as you call these scripts in exactly the right order.  For example, looking at the `github_tag` Circle job, it looks like we run `node scripts/release-tag.js` alone.  `scripts/release-tag.js` is actually dependent on that `tempLogs.json` file, but it won't exist in that workflow, because we don't call `scripts/compile-changelogs.js` first.  So that one will fail.  

In general, I like the direction this PR is headed (breaking apart the monolothic scripts into composable parts). But I think the way it would make more sense to me if we avoid using a temp file to pass data between the scripts.  It seems like `consolidate-changelogs.js` should just do its own changelog file parsing (or call `compile-changelogs.js`, where `compile-changelogs.js` returns the data instead of writing to a temp file).  Then, `release-branch.js` and `release-tag.js` should just parse the `CHANGELOG.md` file to grab the changelog data for the release (it should already know the version # to look for).

scripts/consolidate-changelogs.js
```diff
@@ -0,0 +1,32 @@
...
+  if (err) throw err;
+})
+
+
+// Remove the changelog files
+for (var i=0; i<changelogs.length; i++) {
+  var changeLogFilePath = directoryPath + "/" + changelogs[i];
+  fs.unlink(changeLogFilePath, (err) => {
+    if (err) {
+        console.log(`failed to delete changelog: ${err.toString()}`);
```
rbayliss: Should we also throw this error so the script doesn't finish thinking everything went ok?

rbayliss (changes requested):

scripts/consolidate-changelogs.js
clairesunstudio: So I started with exporting those data in the tempLogs.json, to share between jobs. If I don't create the temp file e.g. changelogs at the time of running release-tag will be removed already so the compiled changelogs will have no content. The only solution I could think of was to create a file and cache it and remove it after the release-tag job gets the data. Is there another I could do this?

clairesunstudio: e.g. https://github.com/massgov/mayflower/pull/904 This was when I used imports/exports

scripts/consolidate-changelogs.js
clairesunstudio: Do you mean `throw err` to exit it out?

clairesunstudio:
```suggestion
        console.log(`failed to delete changelog: ${err.toString()}`);
        throw err;
```
like this ^?

clairesunstudio: or 
```suggestion
        console.log(`failed to delete changelog: ${err.toString()}`);
        process.exit(1);
```

rbayliss: I think just throwing the error will do it.

.circleci/config.yml
```diff
@@ -260,6 +260,8 @@ jobs:
       - run: git config --global user.email "circleci@example.com"
       - run: git config --global user.name "CircleCI Deployment Bot"
       - run:
+          command: 'node scripts/compile-changelogs.js'
+          command: 'node scripts/consolidate-changelogs.js'
           command: 'node scripts/release-branch.js'
```
rbayliss: Sorry I didn't notice this earlier - this syntax isn't valid. You'll need 3 separate `run` steps like:
```
- run: node scripts/compile-changelogs.js
- run: node scripts/consolidate-changelogs.js
- run: node scripts/release-branch.js
```

PR #911, opened 2020-01-27
DP-17150 Relationship Indicator

ygannett:

## Description

- Separate the relationship indicator changes from the global nav feature branch `patternlab/DP-17138_globalnav-relationship-indicator` as a new branch from the latest `develop`.
- Change the label for the secondary set to **OFFERED BY**.
- Change the widths to:
         - Primary - 60%
         - Secondary - 40%
- Remove the 15 px padding at the bottom from the primary set container to match the secondary set's style.
- Fix the keyboard trap on the show more/less button WHEN THE HIDDEN ITEMS ARE DISPLAYED.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-17150)

## Steps to Test

1. Check the following changes in http://localhost:3000/?p=pages-service.

- "Offered by" and Organizations should be on the right in green.  Switch the actual text shown so that "Department of Conservation and Recreation" and the "Other agency..." text are on the right with the header "Offered by".
- "Related to" and the 6 items shown should be on the left and in blue.  Change "Related to" to "Part of"
- Tabbing is trapped when "show more" in the relationship indicator is expanded.  User should be able to tab out of it.
- There should be less space between label ("Offered by" and "Part of") and data. See design linked below.
- Left side containing "Part of" should take up roughly 60% of the total screen width as shown rather than the current width which appears 50%.
- This div `<div class="ma__relationship-indicators--section primary">` has 15px of padding in desktop.  It should not - this is needed only when the relationship indicators are stacked up on a mobile screen.  The padding should match the secondary on a desktop and should be 1 px.
- No keyboard trap around the show more/less button .

## Screenshots
Pattern_Lab_-_pages-service

Pattern_Lab_-_pages-service_1

## Additional Notes:

Anything else to add?

* 

#### @TODO

*


patternlab/styleguide/source/assets/js/modules/headerTags.js
```diff
@@ -59,6 +63,17 @@ export default (function (window,document,$,undefined) {
 
       // Change button text.
       $tagState.text($tagStateText === 'fewer' ? 'more' : 'fewer');
+
+      $button.attr('aria-pressed', function(_, attr) { return !(attr == 'true') });
+
+    });
+
+    $button.on('keydown', function(e) {
+      if (e.keyCode === 9 && $tagWrapper.parent().hasClass('tags-open') && !$button.hasClass('last-item')) {
```
danstafford: This is interesting how the tab focus jumps to the first previously hidden link after opening the toggle. I see how this would ensure that you didn't miss any links as you tab through this list. But just in my opinion I find that a bit confusing to have the focus jump around like that. It seems clearer to me to open the toggle and see additional links and simple have to tab backwards to get to them if you want to click them.

danstafford (commented):
This looks good @ygannett ! Just a few things I noticed...

I ran through the Steps to Test in the description and this looks good but I think the items in the list are mixed up. So for example, I think "Department of Conservation and Recreation" should appear under "Offered by" and be green and remain on the right. And "PART OF" should be on the left and have "Snowmobile in the Parks" as a list item. Does that sound right?

Also this test item looks like it got cut off so I am not sure what div to look for:
> This div "
has 15px of padding in desktop. It should not - this is needed only when the relationship indicators are stacked up on a mobile screen. The padding should match the secondary on a desktop and should be 1 px.

Also I put a comment next to `headertags.js` with some thoughts.

Finally it looks like there is 230 files in this PR which makes it a bit difficult to review. Does `develop` have to be merged back into this or something?

patternlab/styleguide/source/assets/js/modules/headerTags.js
ygannett: This feature is more likely for screen reader users. When they click the button, the hidden content gets accessible.  However, they wouldn't know the revealed content is prier to the button and they need to go backward to access them.
To them, the current set up seems to more make sense as I thought.  If you have better ideas, please share.  I'd love to have your input.

danstafford: Oh OK thanks, that makes sense.

danstafford: Although @ygannett I don't think you typically navigate a webpage using the tab key if you are using a screen reader, at least not if you are using VoiceOver.
https://www.apple.com/voiceover/info/guide/_1131.html

ygannett: Thank you for catching the content mix-up.  I fixed it. 

I'm sorry about the test script.  I hit the wrong key to display the html tag.  
> - This div `<div class="ma__relationship-indicators--section primary">` has 15px of padding in desktop.  It should not - this is needed only when the relationship indicators are stacked up on a mobile screen.  The padding should match the secondary on a desktop and should be 1 px.

It is too many files changed in this branch.  What happened is that another developer worked on this component and another one in a branch while ago, and left.  Since it doesn't make sense to have those two components in the same branch, I took out this component change from the original branch. Since this component is used in many sample pages, the content change in JSON files and reference screenshots for backstop added up the numbers of the changed files.  So, don't get overwhelmed.  The initial changes related to the feature is in these files:

- assets/scss/02-molecules/_relationship-indicators.scss
- assets/scss/02-molecules/_sticky-nav.scss
- assets/scss/03-organisms/_sticky-toc.scss
- patternlab/styleguide/source/_patterns/01-atoms/11-text/wrapped-link.json
- patternlab/styleguide/source/_patterns/01-atoms/11-text/wrapped-link.md
- patternlab/styleguide/source/_patterns/01-atoms/11-text/wrapped-link.twig
- patternlab/styleguide/source/_patterns/02-molecules/relationship-indicators.json
- patternlab/styleguide/source/_patterns/02-molecules/relationship-indicators.twig
- patternlab/styleguide/source/assets/js/modules/headerTags.js

There was some changes in the base branch, so this branch needed to get updated as well.

Can you take a look again and let me know for any questions.  Sorry, this is a messy one. Thank you.

danstafford: OK this looks to me @ygannett , the only thing I see is that I think `<div class="ma__relationship-indicators--section primary">` still has 15px padding-top. I am not sure if it should?

ygannett: @danstafford Good point!  That's true that screen reader users are not typically navigate a page with tab key. But, I don't want to just assume that's the only way how they navigate a page since we cannot assume all possible preferences or situations they access a page. "Tab key to access next focusable item" is available to JAWS, NVDA and VoiceOver:  https://dequeuniversity.com/screenreaders/survival-guide
So, I added those screen readers' general keystrokes to access next item, which are:
- down arrow key (40)
- ctr + alt + right arrow keys (17 + 18 + 39) (Note:  The command works, but not in Voiceover.  Need more digging.)

Thank you so much for bringing this up.

ygannett: I found that it's been a known issue that JS to manage focus order doesn't work with VoiceOver. Replaced the JS feature with CSS. 

By placing the button as the second item in the list and adding `order: 1;`(default order value is 0) to the `li..ma__relationship-indicators__expand-indicators`, keyboard navigation and screen readers respect the order of the items in DOM, and JS is not necessary.

danstafford:
Great solution to this @ygannett ! I took a look at the changes to the files you mentioned and they look good:
```
02-molecules/relationship-indicators.twig
assets/scss/02-molecules/_relationship-indicators.scss
patternlab/styleguide/source/assets/js/modules/headerTags.js
```
And I even checked this out locally using Voiceover and I able to read the hidden links after opening the toggle. This PR looks good to me.

PR #912, opened 2020-01-28
Automation/auto release conditionally

clairesunstudio:
Do not create a release branch or cut a release tag if there are no changelogs


rbayliss (approved):
Ok.  This should cause the release job to fail.  That seems ok for right now, anyway.

clairesunstudio:
@rbayliss Were you suggesting maybe we should use exit code 0 instead?

PR #932, opened 2020-02-10
React/dp 17436 rerender icon

clairesunstudio:
Icon knobs are not updating the story. 
e.g. 
- Go to https://mayflower-react.digital.mass.gov/?path=/story/brand-icons--icon
- Change icon color in the knobs panel
- Icon story is not updating


clairesunstudio:
So far, moved dynamic imports into componentDidMount to prevent memory leak. Added a button knob to trigger key update to force rerender Icon. Icon do update the first time "update" button is clicked, but then according to the following error the Icon unmounts before the next rerender:
Screen Shot 2020-02-10 at 11 54 41 AM

clairesunstudio:
This will be addressed in storybook upgrade with controls --> https://github.com/massgov/mayflower/pull/1162

PR #939, opened 2020-02-18
All/dp 17528 mayflower theme package

clairesunstudio:
The goal of this PR is to expose mayflower theme as a package.
Approach: prep and release `assets/` as an npm package `mayflower-theme`, expose meaningful SCSS variables and styles, consume `mayflower-theme` in react and patternlab

Prep assets:
- Added style linting to mayflower scss in assets
- Fixed all linting issues


assets/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "keywords": [
+    "theme",
+    "scss"
+  ],
+  "author": "Massachusetts Digital Services (MDS)",
+  "license": "MIT",
+  "dependencies": {
+    "bourbon": "^5.1.0",
+    "bourbon-neat": "^1.7.2",
```
sawyerh: Can you add a bit more context for how/what `bourbon` and `bourbon-neat` are used for? This is a really large PR, making it hard to review and grasp all at once. I'm wondering if there are ways to either replace them (if it's minor) or break them out as separate packages so that this core Sass package doesn't have them as dependencies.

assets/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "normalize-scss": "^7.0.1"
```
sawyerh: Same comment as above. Can you provide context for how/what `normalize-scss` is used for?

sawyerh (commented):
Feel free to ignore this, but I think this PR could benefit from being broken into a few separate PR's which would make it easier to review. It looks like a lot of the diff is formatting-related. Doing a large formatting change in an isolated PR would make this PR much smaller and easier to understand the larger change being made.

assets/package.json
clairesunstudio: agreed, this PR should be broken down into several steps. 
I added style lint to scss (something we have been wanting to do) so that the package won't have malformatted scss, and would enforce a check going forward. which is causing a lot of diffs. I will break that piece out first. 
After chatting with @rbayliss about the issue that I'm having with the paths for external dependencies (bourbon, bourbon-neat), there isn't a straightforward way to resolve it, and the plan is to do an audit and cleanup those dependencies. I will make a separate PR for that.

clairesunstudio: bourbon and bourbon-neat were used in a lot of component scss as mixins and variables... It might take me a while to identify all the cases that we are using those for

assets/package.json
clairesunstudio: I think that's just normalizing stylings for different browsers. We can exclude this from the mayflower theme package

assets/package.json
```diff
@@ -1,8 +1,9 @@
 {
-  "name": "mayflower-theme",
-  "version": "1.0.0",
+  "name": "@massds/mayflower-theme",
```
sawyerh, line 2: Carrying this convo over from #942: 

> Is this a theme or is it more accurate to say this is a package of styles and static assets? It could be confusing to name this `mayflower-theme` if there becomes a `mayflower-uswds-theme` for example. This might not be the right PR for this discussion though, so feel free to ignore this comment.

> which package name makes more sense mayflower-styles, mayflower-assets, mayflower-styles-and-assets?

I think `mayflower-styles` is accurate. Technically the fonts, icons, and imagery is part of the styling, so that doesn't seem too far of a reach.

PR #942, opened 2020-02-20
All/dp 17528 add style lint

clairesunstudio:
Prep assets for exports (This PR is broken out from #939)
- Added style linting to mayflower scss in assets
- Fixed all linting issues


assets/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "mayflower-theme",
+  "version": "1.0.0",
+  "description": "Mayflower theming assets, including color variables, fonts, icons and other imagery, along with componentized scss used to style Mayflower markup",
+  "main": "index.js",
```
sawyerh, line 5: nit: I think we can get rid of `main` if `index.js` doesn't exist in this directory

assets/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "mayflower-theme",
```
sawyerh, line 2: Is this a theme or is it more accurate to say this is a package of styles and static assets? It could be confusing to name this `mayflower-theme` if there becomes a `mayflower-uswds-theme` for example. This might not be the right PR for this discussion though, so feel free to ignore this comment.

sawyerh (approved):

assets/package.json
clairesunstudio: thanks good catch! It's missing some commits here

clairesunstudio: Can I leave these in here. I think the rest of the PR #939 address updates this file and adds the entry points

assets/package.json
clairesunstudio: ok, let's table this for the next PR, and see which package name makes more sense `mayflower-styles`, `mayflower-assets`, `mayflower-styles-and-assets`?

PR #947, opened 2020-02-21
All/dp 17532 publish mayflower tokens

clairesunstudio:
Created a color token variable scss stylesheet and separated it from the Mayflower usage mapping, to include into the Mayflower-tokens package
- Added new color gradient variables into color tokens
- Some refactoring was needed to structure the package

See published package --> [@massds/mayflower-tokens](https://www.npmjs.com/package/@massds/mayflower-tokens)

--- 

### Changed:
  - project: Assets
    component: SCSS
    description: Separated color token variable scss from the color usage mapping. (#947)
    issue: DP-17532
    impact: Minor

### Added:
  - project: Assets
    component: SCSS
    description: Added and refactored mayflower color gradients into color token stylesheet. (#947)
    issue: DP-17532
    impact: Minor
  - project: Assets
    component: SCSS
    description: Registered a new `mayflower-token` package in npm to include color tokens, fonts and images. (#947)
    issue: DP-17532
    impact: Minor

PR #948, opened 2020-02-24
All/dp 17532 publish mayflower tokens followup

clairesunstudio:
### Description
This PR is a follow up for #947. 
It structured the package to look like the tree below, and it added a build step to allow mapping the src to a different structure for distribution `npm run build`
```
├── dist   
    ├── fonts                 Texta web fonts      
    ├── images             
    │   ├── svg-icons         SVG icon files
    │   ├── stateseal         stateseal png
    │   ├── stateseal-color   stateseal png in color grayscale
    ├── scss        
    │   ├── _colors.scss      Color tokens
    │   ├── _fonts.scss       Texta font imports (needs to set $assets-path)
```

This PR also added `lighter` `lightest` `darker` `darkest` variables consistently across all brand colors, adjusted the variable labels in the storybook to reflect consistency.

After label cleanup in storybook:
Screen Shot 2020-02-25 at 11 48 54 AM
Screen Shot 2020-02-25 at 11 49 11 AM

Before label cleanup in storybook:
https://mayflower-react.digital.mass.gov/?path=/story/brand-colors--colors
https://mayflower-react.digital.mass.gov/?path=/story/brand-colors--gradients-light

### Test: 
To test in PFML portal:
https://github.com/EOLWD/pfml/pull/50

### (resolved):
`@use` is not supported in node-sass. We will keep using `@import` for now
>I can't get `@use` to work, and I think that it makes more sense to do that on the consumer side of the package. 
To replicate:
>1. checkout PFML portal codebase and `minghua/integrate-mayflower-tokens` branch
>2. uncomment https://github.com/EOLWD/pfml/pull/50/files#diff-de83abbac905f608bed4081c4d6a599cR12
>3. currently showing errors undefined variables


assets/package.json
```diff
@@ -1,19 +1,16 @@
 {
   "name": "@massds/mayflower-tokens",
-  "version": "1.0.1",
+  "version": "1.3.0",
   "description": "Mayflower design tokens — it includes color token SCSS variables, fonts, icons and other imagery",
-  "style": "colors/_mayflower-color-tokens.scss",
+  "style": "dist/scss/_colors.scss",
```
rbayliss: Does it make sense to have a default export here?  I think if we do, it should be an export of all of the variables, not just colors.  I guess we could always switch that later though, as long as we're only adding variables to the default export, not removing them.

clairesunstudio: Do you mean an index.scss under dist/scss that imports _colors.scss and _fonts.scss?

clairesunstudio: We originally talked about only including colors, fonts, icons into the mayflower-tokens package, are there other things you think make sense to include?

rbayliss:
Looks good to me!

PR #954, opened 2020-02-27
[TEST] DP-17483 dataviz textwrapping container

ygannett:

## Description

Set up a `section.ma__rich-text__container` wrapper for `.ma__rich-text` content.
## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


patternlab/styleguide/source/_patterns/01-atoms/09-media/figure.twig
```diff
@@ -1,16 +1,63 @@
...
-  {% set image = figure.image %}
   {% if figure.title %}
   <span class="ma__figure__title">
     {{ figure.title }}
   </span>
   {% endif %}
-  {% include "@atoms/09-media/image.twig" %}
+
+  {# For image. #}
+  {% if figure.image %}
```
rbayliss: Ooh... This is a lot of conditional logic in what should be a very basic "fundamental" template.  I know we've got this kind of pattern all over Mayflower, but in my opinion, `figure` should not have any opinion about what goes inside of it.  Instead, it should allow you to inject whatever kind of thing you want, and allow the class to be set accordingly.  Eg:
```twig
# figure.twig
{% set alignmentClass = figure.align ? "ma__figure--" ~ figure.align : "" %}

<figure class="ma__figure {{ alignmentClass }}{{ figure.class ? figure.class ~" " : "" }}">
  {% if figure.title %}
  <span class="ma__figure__title">
    {{ figure.title }}
  </span>
  {% endif %}
  {{ figure.content }}
  {% if figure.caption %}
    <figcaption class="ma__figure__caption">
      {{ figure.caption }}
    </figcaption>
  {% endif %}
</figure>
```
Then, we could create wrapper components for the different types of figures:
```twig
{# figure--image.twig #}
{% if figure.image %}
  {# Render figure.content, which will be the main content of the figure #}
  {# Important: we wrap this in an `if` statement to allow consuming systems
     to either use figure.image or figure.content (if they prefer to externally
     render figure.content
   #}
  {% set figure = figure|merge({
    content: include("@atoms/09-media/image.twig", {image: figure.image})
  }) %}
{% endif %}
{% include "figure.twig" with {
  figure: figure|merge({
    class: "ma__iframe__container js-ma-responsive-iframe"
  })
} %}

{# figure--iframe.twig #}
{% if figure.iframe %}
  {# Render figure.content, which will be the main content of the figure #}
  {# Important: we wrap this in an `if` statement to allow consuming systems
     to either use figure.iframe or figure.content (if they prefer to externally
     render figure.content
   #}
  {% set figure = figure|merge({
    content: include("@atoms/09-media/iframe.twig", {iframe: figure.iframe}),
  }) %}
{% endif %}
{% include "figure.twig" %}

{# figure--dataviz.twig #}
{% if figure.dataviz %}
  {# Render figure.content, which will be the main content of the figure #}
  {# Important: we wrap this in an `if` statement to allow consuming systems
     to either use figure.iframe or figure.content (if they prefer to externally
     render figure.content
   #}
  {% set content %}
    <script type="text/javascript">
      window.addEventListener("DOMContentLoaded", function() {
        new tableau.Viz(document.getElementById("{{ figure.dataviz.id }}"), "{{ figure.dataviz.url }}");
      });
    </script>
  {% endset %}
  {% set figure = figure|merge({
    content: content,
  }) %}
{% endif %}
{% include "figure.twig" with {
  figure: figure|merge({
    class: "ma__iframe__container js-ma-responsive-iframe"
  })
} %}
```
I know this is a pretty big deviation from how we've done things in other parts of Mayflower, but this would allow us to reuse a very simple core "figure" template, while still allowing for complex variation through the use of these wrapper templates.  Additionally, it allows us to write the Tableau embed code in Drupal, then inject it into `figure--dataviz.twig` via the `content` variable, which is what we really want to do.  

What do you think?

rbayliss (commented):
Hey Yukiko,
I think the core idea here is great.  I have a suggested direction change below.  Happy to discuss if you want - I think you probably have your finger on the pulse of Mayflower better than I do, but the approach I suggested makes a lot of sense to me as a way of keeping complex logic out of simple templates.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
```diff
@@ -0,0 +1,31 @@
+{% if figure.dataviz %}
+  {# Render figure.content, which will be the main content of the figure #}
+  {# Important: we wrap this in an `if` statement to allow consuming systems
+     to either use figure.image or figure.content (if they prefer to externally
```
rbayliss: This comment should probably say `figure.dataviz` rather than `figure.image`.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
```diff
@@ -0,0 +1,31 @@
...
+     render figure.content
+   #}
+  {% set align = figure.align|default('left') %}
+  {% set id = figure.dataviz.id %}
```
rbayliss: I don't think we need this variable - we're referencing `figure.dataviz.id` below.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
```diff
@@ -0,0 +1,31 @@
...
+    </script>
+  {% endset %}
+  {% set figure = figure|merge({
+    'content': content,
+  }) %}
+{% endif %}
+{% include "@atoms/09-media/figure.twig" with {
+  figure: figure|merge({
+    'align': align,
+    'id': id,
```
rbayliss: `id` doesn't seem to be used in `figure.twig`, so I don't think we want/need to pass it through.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
```diff
@@ -0,0 +1,31 @@
...
+     render figure.content
+   #}
+  {% set align = figure.align|default('left') %}
+  {% set id = figure.dataviz.id %}
+  {# TODO:  SET SIZE VALUES #}
+  {# Set default size to large. #}
+  {% set size = "ma__figure--" ~ figure.dataviz.size|default('large') %}
+  {% set class = size ~ ' ' ~ 'ma__dataviz' %}
+  {% set content %}
+    <div id="{{ figure.dataviz.id }}"></div>
```
rbayliss: Can you see a reason that we would want a static ID here?  We just need it to be able to reference the div a moment later, and it seems like needing to generate a unique ID might be a pain for consuming systems.  I'd probably recommend that we randomly generate one.  Eg: 
```
{% set id = 'dataviz-' ~ random(9999) %}
```

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
```diff
@@ -0,0 +1,31 @@
...
+  {# Render figure.content, which will be the main content of the figure #}
+  {# Important: we wrap this in an `if` statement to allow consuming systems
+     to either use figure.image or figure.content (if they prefer to externally
+     render figure.content
+   #}
+  {% set align = figure.align|default('left') %}
+  {% set id = figure.dataviz.id %}
+  {# TODO:  SET SIZE VALUES #}
+  {# Set default size to large. #}
+  {% set size = "ma__figure--" ~ figure.dataviz.size|default('large') %}
```
rbayliss: I think the `size` should be a property of `figure`, given that we will want these same sizes to apply to images and other things in the future, right?  Also, you're formulating the class inside of the `{% if figure.dataviz %}` block.  Keep in mind that Drupal will not use `figure.dataviz` - it will use `figure.content` instead, so we need to move size and class calculation outside of the `if` statement here.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure.twig
```diff
@@ -1,16 +1,14 @@
-{% set figureClass = figure.align ? "ma__figure--" ~ figure.align : "" %}
-
-<figure class="ma__figure {{ figureClass }}">
-  {% set image = figure.image %}
+{% set alignmentClass = figure.align ? " ma__figure--" ~ figure.align : "" %}
```
rbayliss, line 1: Should we also add a `sizeClass` calculation here?  My understanding is that we want to be able to apply consistent size classes to tableau embeds to start with, but that images and iframes will be coming soon. If that's right, the `figure` template could accept a `size` argument that allows you to set the size:
```twig
{% set sizeClass = figure.size ? " ma__figure--" ~ figure.size : "" %}
```
Then, below, we'd print it:
```twig
<figure class="ma__figure{{ alignmentClass }}{{ sizeClass }}{{ figure.class ? " "~ figure.class : "" }}">
```

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure~dataviz.json
```diff
@@ -0,0 +1,12 @@
+{
```
rbayliss: I think we can remove this file now, right? We just want `figure--dataviz.json` I believe.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure~iframe.json
```diff
@@ -0,0 +1,13 @@
+{
```
rbayliss: Same with this one - seems like it can be removed.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
rbayliss: Implemented in #971.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
rbayliss: Implemented in #971.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
rbayliss: Implemented in #971.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
rbayliss: Implemented in #971.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
rbayliss: Implemented in #971.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure.twig
rbayliss: Implemented in #971.

rbayliss (changes requested):
Hey @ygannett, this is looking really close! Some suggestions below. Because we're running short on time for this, and I think it's pretty critical, I went ahead and submitted an adjustment PR to merge into this branch here: https://github.com/massgov/mayflower/pull/971. If you're on board with those changes (which I've given the reasoning for below), feel free to merge it into this branch.

Other than that, I think we just need to capture some backstop reference images and we're ready for merge.  Let me know if you want to discuss anything!

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--iframe.twig
```diff
@@ -0,0 +1,15 @@
+{% if figure.iframe %}
+  {# Render figure.content, which will be the main content of the figure #}
+  {# Important: we wrap this in an `if` statement to allow consuming systems
+     to either use figure.image or figure.content (if they prefer to externally
+     render figure.content
+   #}
+  {% set figure = figure|merge({
+    'content': include("@atoms/09-media/iframe.twig", {'iframe': figure.iframe})
```
rbayliss: Ah, I just tested this on the Drupal side, and the function form of `include` doesn't work (the markup gets escaped).  In my suggestions PR, I just converted it to the block form for `figure--iframe` and `figure--image`.

rbayliss:
I rolled a Drupal branch to test this work with: `DP-17483-dataviz-wrap`.  Right now, the branch does not deal with sizing or alignment for Tableau embeds - it just fixes any backward compatibility issues with the changes in this PR.

Right now, the Backstop diff I'm getting is really only on the spacing for rich text paragraphs (see screenshot).
Screen Shot 2020-03-09 at 1 12 39 PM
I'm not sure whether this is of concern or not.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--iframe.twig
rbayliss: Implemented in #971.

patternlab/styleguide/source/_patterns/01-atoms/09-media/figure--dataviz.twig
ygannett: I thought the id value passed from Drupal would be unique enough. So, I set it as static.  That's not the case?

PR #957, opened 2020-02-28
All/dp 17662 mayflower reorganization

clairesunstudio:
### Updated S3 bucket paths and urls:
• **Patternlab**: `mayflower.digital.mass.gov` → `mayflower.digital.mass.gov/patternlab`
• **React storybook**: `mayflower-react.digital.mass.gov` → `mayflower.digital.mass.gov/react`

### Remove patternlab npm package:
• Stop publishing `@massds/mayflower`

--- 

## Deployment steps (in order):
1. [ ] Deploy AWS changes
2. [ ] Merge PR into develop
3. [ ] Mayflower release


rbayliss (approved):
This looks pretty good to me.  I guess the one question I have is whether there is anything blocking the cutover to the new URLs right now.  I see the following things that might or might not block merge:

1. I know early on we were recommending that people use this bucket as a CDN for Mayflower.  Do you know if anyone actually ended up doing that?  I guess we could check the request logs for this bucket, but that's a pretty involved process.
2. Gitbook hosting - have we determined how to actually put Gitbook on this domain?  Is it gonna live at `/`, or somewhere else? 
3. Mayflower Patternlab has an issue where some of its JS patterns only work when it's viewed through `/index.html`, not `/b/XXX/index.html`.  I forget what actually breaks, but this will probably also affect us if we move the whole thing a directory deeper.  Maybe @ygannett remembers what it was that broke a week or two ago?  

Things I see that will need followup after this PR lands:
* Cleanup the S3 bucket by removing or moving all of the `/b/*` and `/v/*` directories, as well as all of the files deployed at the root.
* Cleanup old S3 buckets (mayflower-react.digital...)
* Redirect: 
    * `mayflower-react.digital.mass.gov` to `mayflower.digital.mass.gov/react`.
    * Mayflower Gitbook URL -> `mayflower.digital.mass.gov`

clairesunstudio:
> I know early on we were recommending that people use this bucket as a CDN for Mayflower. Do you know if anyone actually ended up doing that? I guess we could check the request logs for this bucket, but that's a pretty involved process.

This is a very good point! A couple of things that I know of are definitely pointing to the CDN, e.g. https://uploads.massgis.digital.mass.gov/signup.html, https://github.com/massgov/mayflower-starters/blob/master/index.html I'm sure there are other things. There are probably other things that are not on my radar. I need to check with a few customers. Maybe we could keep those resources there for now before other things catch up.

> Gitbook hosting - have we determined how to actually put Gitbook on this domain? Is it gonna live at /, or somewhere else?

I was thinking about creating a landing page using Gatsby, and then linked all the other resources off of that landing page, so Gitbook will be /docs. Thoughts? If we are building that page, should the code live in the monorepo? My plan is to slowly move component documentation into storybook, using storybook docs MDX instead of gitbook to couple live components with usage documentation

> Mayflower Patternlab has an issue where some of its JS patterns only work when it's viewed through /index.html, not /b/XXX/index.html. I forget what actually breaks, but this will probably also affect us if we move the whole thing a directory deeper. Maybe @ygannett remembers what it was that broke a week or two ago?

I didn't know about this. Let's loop in @ygannett on this?

PR #958, opened 2020-03-02
[React] upgrade storybook to lastest version 5.3.18

clairesunstudio:
Changed:
  - project: React
    component: Dependency
    description: Upgrade storybook to lastest version 5.3.18. (#958)
    issue: DP-17437
    impact: Patch


react/.storybook/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+    "plugins": ["@babel/proposal-export-default-from"]
```
rbayliss: I think we don't need this file anymore - this was just a step we took in debugging, right?  Babel settings are configured in `main.js`

react/src/README.md
```diff
@@ -0,0 +1,124 @@
+# Mayflower-React
```
rbayliss: Did you mean to copy this file into `src/`?  This was living at the package root, and I can't see a reason to duplicate it unless Storybook requires it for some reason.

rbayliss (commented):
This seems good to me - a couple of questions/cleanup items noted below though:

react/src/README.md
l0c0luke: initially `npm start` was barfing saying that it was referencing a file outside of the `src` directory. That is not the case anymore, so I removed it and am pointing to the root one.

clairesunstudio:
Screen Shot 2020-04-05 at 4 18 18 PM

I'm getting this error for all stories

clairesunstudio:
Error resolved by recloning the repo

react/src/components/base/Typography/Typography.stories.js
```diff
@@ -13,7 +13,7 @@ import {
   styles
 } from './typography.json';
 import typographyOptions from './Typography.knobs.options';
-import TypographyDocs from '../../../../../docs/style/typography.md';
+import TypographyDocs from './typography.md';
```
clairesunstudio: Is there a reason to duplicate this doc from the [gitbook subrepo](/docs)

react/src/components/atoms/media/SiteLogo/SiteLogo.stories.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 import { storiesOf } from '@storybook/react';
 import { withKnobs, text, number } from '@storybook/addon-knobs';
-import logo from 'SharedAssets/images/stateseal.png';
+import logo from '../../../../assets/images/stateseal.png';
```
clairesunstudio: Can we add back the `sharedAssets` alias? We might want to change the path in the future if we can find a better solution than creating symlinked folder from `/assets`

clairesunstudio (commented):
Looks good! Nice job!

react/src/components/base/Typography/Typography.stories.js
clairesunstudio: Currently, the docs sub repo is syncing from Gitbook, which allows non-tech people to use the wysiwyg interface to collaborate on documentation, so will be nice to keep that instead of creating a duplicate

PR #985, opened 2020-03-17
update symbolic links

clairesunstudio:
### Issue:
In [9.42.0](https://github.com/massgov/mayflower/releases/tag/9.42.0), we started publishing `@massds/mayflower-tokens` package from '/assets' and it's causing issues in the`@massds/mayflower-react` package, possibly relating to the package.json in /assets

#####  (9.41.0):
Screen Shot 2020-03-18 at 10 57 39 AM

#####  (9.42.1):

Screen Shot 2020-03-18 at 10 56 55 AM

##### To Replicate:
- [ ] checkout out branch `react/npm-package-fix`
- [ ] `cd site`
- [ ] change mayflower-react version in package.json to before and after 9.42.1
- [ ] `yarn`
- [ ] look in the node_module/@massds/mayflower-react/assets

### Quick Fix:
Update symlink to `/assets` to only include the following folders in `/react/src`
```
assets
|—images
|—fonts
|—scss
```

### Help wanted:
New symlinks to the folders don't seem to work
- [ ] checkout out branch `react/npm-package-fix`
- [ ] cd react
- [ ] `npm run develop`


rbayliss:
@clairesunstudio, this didn't actually work for me. It looks like these symlinks are wired up wrong.  For a quick fix, how about if we just run `rm react/es/assets/package.json` during the build process?  That works for me (I tested using `npm pack`), and we could just throw it in CircleCI's `react_build_storybook` job.

PR #1006, opened 2020-03-31
React/dp 17982 replace rich text component

clairesunstudio:
### Goal
The goal of this PR is to replace the existing RichText component which has a set template with a more flexible solution, that can render from raw HTML. As this is proven needed and more utilitarian for other projects rather than the existing RichText functionality.

The PressTeaser component (deprecated) is using RichText to render its description. Since this component is no longer used in other React projects and can be fully replaced with GenTeaser, we can remove the component from the mayflower-react library

----
### Changelog

Removed:
  - project: React
    component: PressTeaser
    description: Remove PressTeaser molecule from mayflower-react npm package. (#1006)
    issue: DP-17982
    impact: Major

Changed:
  - project: React
    component: RichText
    description: Render welcome page to Mayflower-react. (#1006)
    issue: DP-17982
    impact: Patch


react/src/components/organisms/RichText/RichText.md
```diff
@@ -1,57 +1,5 @@
 ### Description
-This pattern displays the contents of a rich text editor with an optional title and learn more link.
+The RichText component is a wrapper component that renders raw Html as React components, using the [react-html-parser](https://www.npmjs.com/package/react-html-parser) library. It is to avoid the use of dangerouslySetInnerHTML. The RichText component can render raw HTML nodes passed as children, or an HTML string passed via prop `rawHtml`.
```
l0c0luke: `Html` should be `HTML` for consistency with how it is further below.

l0c0luke: use back ticks around `dangerouslySetInnerHTML`

react/src/components/organisms/RichText/RichText.stories.js
```diff
@@ -1,76 +1,39 @@
...
-          <Paragraph {...paraOptionsWithKnobs} />
-          <Paragraph text={text('Paragraph2: text', 'foobarbaz', 'Paragraph')} />
-          <Heading text={text('Heading: text', 'This is a heading', 'Heading')} level={1} />
-          <UnorderedList {...orderedOptionsWithKnobs} />
-          <DecorativeLink {...decoreOptionsWithKnobs} />
-        </RichText>);
+        <RichText {...props}>
+          <h1>This is a heading.</h1>
+          <p>
+          An <b>unordered list</b> is a list in which the sequence of items is not important. Sometimes, an unordered list is a bulleted list. And this is a long list item in an unordered list that can wrap onto a new line.
```
l0c0luke: My IDE is saying this indent is incorrect (should be 2 more), but seems eslint passed anyways?

react/src/components/organisms/RichText/RichText.stories.js
```diff
@@ -1,76 +1,39 @@
...
-                  { text: 'This is a nested list item' },
-                  { text: 'This is another nested list item in an unordered list' }
-                ]
-              },
-              { text: 'This is the last list item' }
-            ]
-          }
-        }]
+      const props = {
+        rawHtml: text('rawHtml', '<h1>This is heading 1</h1>'),
```
l0c0luke: you include this `knob` but it is never used or has no visible change in the preview. Maybe I am misunderstanding it's use?

react/src/components/organisms/RichText/RichText.stories.js
```diff
@@ -1,76 +1,39 @@
...
+        className: text('className', ''),
+        id: text('id', ''),
+        htmlTag: text('htmlTag', 'div'),
```
l0c0luke: when I change this `knob` I create an error seen in the preview, after backspacing the `knob` to an empty string. If I put `span` or `div` back in, the error never goes away.

Screen Shot 2020-04-02 at 12 12 24 PM

l0c0luke (changes requested):

react/src/components/organisms/RichText/RichText.stories.js
clairesunstudio: This is because if children are passed, it will render children nodes rather than the rawHtml string. I added a conditional check there so it's less confusing in storybook

l0c0luke (approved):

PR #1028, opened 2020-04-27
React/dp 18466 refactor label

clairesunstudio:
Added:
  - project: React
    component: Label
    description: Added Label component to render label and legend for input and fieldset elements. (#1028)
    issue: DP-18462
    impact: Patch
    
    
---

JIRA Ticket --> https://jira.mass.gov/browse/DP-18466

This is also an accessibility fix


react/src/components/forms/SelectBox/index.js
```diff
@@ -42,36 +43,42 @@ class SelectBox extends React.Component {
...
     const selectBoxInline = stackLabel ? '' : 'ma__select-box__field--inline';
     const getValueByText = (array = [], text) => {
       const matchedItem = array.find((item) => item.text === text);
       const matchedValue = matchedItem && matchedItem.value;
       return matchedValue;
     };
-    const valueInOptions = getValueByText(this.props.options, selected);
-    const selectedValue = valueInOptions || this.props.options[0].value;
+    const valueInOptions = getValueByText(options, selected);
+    const selectedValue = valueInOptions || options[0].value;
```
smurrayatwork, line 57: What happens when options[0] does not exist? Is options a required prop?

react/src/components/forms/Input/index.js
```diff
@@ -16,18 +17,17 @@ const Input = (props) => {
...
     [classes && classes.length > 0 && classes.join(' ')]: true
   });
-  const inputLabelClasses = classNames({
-    ma__label: true,
-    'ma__label--hidden': (labelText && hiddenLabel),
-    'ma__label--required': (labelText && required),
-    'ma__label--optional': (labelText && !required),
-    'ma__label--disabled': (labelText && disabled)
-  });
+  const conditionText = required ? null : 'optional';
```
smurrayatwork: Technically this should be set to empty string instead of null. The Label component should check to see if the length of condtionText is greater than zero, since the propType is set to be a string and not null.

react/src/components/forms/SelectBox/index.js
clairesunstudio: yes, updated the prototype. I envision a lot of clean up work in form elements coming up. I didn't pay much attention to things other than the label in this pr

PR #1047, opened 2020-05-06
Assets Integration Branch: React Support with Mayflower Assets Package

smurrayatwork:
## Description
This adds support for Mayflower React to use Mayflower Assets.

- The symlink to assets has been removed and replaced by using Mayflower Assets from NPM.
- Support for using Dart Sass instead of node-sass has been implemented.
- The npm scripts `build-css` and `watch-css` have been updated to use Dart Sass' `sass` command instead. The `sass` command supports watching for changes.
- A `.env` file has been added with the sass include paths needed for compilation. The npm package `env-cmd` uses this file before `sass` is ran.
- Global styles for storybook are now included under `.storybook/preview.js` and configured by `.storybook/mayflowerTheme.js`.
- Storybook's configuration has been updated to point to the `src/` directory for stories. In addition, the babel plugin for dynamic imports has been removed, as it was breaking builds for the Icon component. The css rule addition for mini-css-extract-plugin was actually being fired on both development and production modes - this was updated to by dynamic and to only run for production bundles. Finally, SharedAssets was updated to point to the `static/` directory used by Mayflower Assets.
- React component `.scss` files have been updated to replace `@import` statements with `@use` where appropriate.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-18263)


react/.storybook/main.js
```diff
@@ -1,33 +1,38 @@
-const path = require("path");
+const path = require('path');
 module.exports = {
-  stories: ['../**/*.stories.js'],
+  stories: ['../src/**/*.stories.js'],
   addons: [
-    '@storybook/addon-knobs',
+    //'@storybook/preset-create-react-app',
```
smurrayatwork:
```suggestion
```

clairesunstudio (approved):

react/.storybook/preview.js
```diff
@@ -1,6 +1,7 @@
 import { addDecorator, addParameters } from '@storybook/react';
 import { withA11y } from '@storybook/addon-a11y';
 import { withInfo } from '@storybook/addon-info';
+import '../src/index.css';
```
clairesunstudio, line 4:
```suggestion
import '../src/index.css';
```
why is this import here?

smurrayatwork: That pulls in the global styling for storybook for things that should be imported only once. So things like font files, or box sizing, etc.

PR #1052, opened 2020-05-11
Assets Integration:  Patternlab

smurrayatwork:
## Description
This PR adds support for the Mayflower Assets package with Patternlab. Changes:

- node_sass has been replaced with dart sass' JS implementation.
- The `shared` path in gulpfile.js has been updated to point to the mayflower assets package under node_modules.
- The `patterns` gulp task has been changed to not use parallel, as assets referenced by the `patternlab:patterns` task need to be copied over before it can begin.
- `icon.function.php` has been updated to use the public directory set in `gulpfile.js` for the svg-icon pathing.
- `source/_data/data.json` has been updated to use a relative path at the root to find stateseal.png.
- `source/assets/scss/index.scss` and `source/assets/scss/pattern-scaffolding.scss` have been updated to use the `@use` syntax instead of `@import`. In addition, the need for bourbon and bourbon-neat has been removed.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-18263)


patternlab/styleguide/source/_data/data.json
```diff
@@ -5,6 +5,6 @@
   "bodyClass": "",
   "directory": "assets",
   "googleLanguages": "ar,es,fr,pt,ht,it,km,ko,ru,vi,zh-CN",
-  "sealImage": "../../assets/images/stateseal.png",
+  "sealImage": "/assets/images/stateseal.png",
```
clairesunstudio, line 8: is this using the mayflower-assets package?

clairesunstudio: oh it is nvm

clairesunstudio (approved):

patternlab/styleguide/source/_data/data.json
smurrayatwork: Yes

PR #1066, opened 2020-05-21
React/dp 18576 deprecate input text type ahead

clairesunstudio:
- Remove deprecated InputTextTypeAhead and change it to InputTextFuzzy in the following components:
  - Typeaheaddropdown
  - HeaderSearch
  - SearchBanner
- [InputTextTypeAhead] Prevent onSuggestionClick to trigger without a match on "enter", and added onKeyDown callback function
----

-  Deprecate and remove PressFilters

---- 
- Deprecated and removed OrgSelector. 
> OrgSelector should be replaced by combining InputTextFuzzy and ImagePromo directly, the logic is moved into the consumer side using their own state or context.  See search implementation for example --> https://github.com/massgov/massgov-search/blob/develop/src/components/OrgInfo/index.js


react/src/components/forms/InputTextFuzzy/index.js
```diff
@@ -67,19 +67,23 @@ class InputTextFuzzy extends React.Component {
...
+      if (is.fn(this.props.onSuggestionClick) && match) {
         this.props.onSuggestionClick(event, { method: 'enter', suggestion: { item: { text: value } } });
       }
     }
   }
   handleFocus = (event) => {
     if (is.fn(this.props.onFocus)) {
       event.persist();
-      this.props.onFocus({ event, value: this.state.value, suggestions: this.state.suggestions });
+      this.props.onFocus(event, { value: this.state.value, suggestions: this.state.suggestions });
```
smurrayatwork: This is a breaking change to any projects using InputTextFuzzy that relies on event. For now, it might be best to do:
```suggestion
      this.props.onFocus(event, { event, value: this.state.value, suggestions: this.state.suggestions });
```

react/src/components/forms/TypeAheadDropdown/index.js
```diff
@@ -80,17 +80,18 @@ class TypeAheadDropdown extends React.Component {
...
     }
   }
-  handleSelect(event, input) {
+  handleSelect(event, { suggestion }) {
     // Stop the filters form submission if enter is pressed in the selector.
     event.preventDefault();
     // Update this component state and pass the event out to the calling code.
-    if (input.suggestion.text !== '') {
+    const { text } = suggestion.item;
+    if (text !== '') {
```
smurrayatwork:
```suggestion
    if (text.length > 0) {
```

smurrayatwork (approved):

PR #1068, opened 2020-05-21
add fonts for vietnamese

clairesunstudio:
Add fonts (Muli) for Vietnamese support to compliment Texta for its ransom note effect across the accent glyphs. 
> Please note this is a stop-gap before we decide on a final candidate to replace Texta.

---
Before:
Screen Shot 2020-05-21 at 12 56 22 PM

After:
Screen Shot 2020-05-21 at 1 06 26 PM

---
Before:
Screen Shot 2020-05-21 at 1 40 18 PM

After:
Screen Shot 2020-05-21 at 1 46 33 PM

----

robb - Last Call Media  10:28 AM
Hey all, we’re about to launch Vietnamese on unemployment.mass.gov, and we’ve got that classic ransom note font effect going on.  I recall @minghuas saying that she was working on a fix for this at one point.  Was that ever implemented?
unemployment.mass.govunemployment.mass.gov
MA Unemployment Benefits Application
Massachusetts Unemployment Insurance (UI) Online Application for Benefits

minghuas  10:40 AM
not Vietnamese, I added suppport for Khmer.  I can take a look at the vietnamese font options too

minghuas  10:55 AM
ya this is a known issue of Texta --> mayflower/?path=/story/brand-typography--texta-main-font, you do see the ransom note effect across its character set. I'm currently prototyping two Texta replacement candidates, one of which is Noto and it has very comprehensive language support and we can also conditionally load different font face based on unicodes. I can add a temporary patch to the ui app to switch out the fonts for vietnamese for the launch, while we wait till we get a final candidate approved for mayflower, does that work for you @robb - Last Call Media?

PR #1076, opened 2020-05-29
Revert "add fonts for vietnamese"

tom-fleming:
Reverts massgov/mayflower#1068.

It looks like Drupal CSS aggregation is having issues with this font. Rolling back until we can implement a fix.


clairesunstudio:
Per our conversation in slack, let's do this instead --> https://github.com/massgov/mayflower/pull/1077

PR #1077, opened 2020-05-29
[hotfix] remove param from font import statement

clairesunstudio:
>Drupal is having some issues working with the new Vietnamese fonts, preventing the CD environment from loading. 
https://www.drupal.org/project/drupal/issues/2936067
Drupal.org
CSS aggregation fails if an @import contains a semicolon
Problem/Motivation If you have an @import statement in one of your CSS files, \Drupal\Core\Asset\CssCollectionOptimizer::optimize() tries to move it at the beginning of the aggregated document according to the W3C spec.

@import url('https://fonts.googleapis.com/css2?family=Muli:ital,wght@0,300;0,400;0,500;0,600;0,700;0,800;1,300;1,400;1,500;1,600;1,700;1,800&display=swap'); --> @import url('https://fonts.googleapis.com/css2?family=Muli');

Screen Shot 2020-05-29 at 12 35 27 PM

Removing the params from font import statement doesn't seem to change things visually


tom-fleming:
@clairesunstudio The Drupal site loads again locally with this fix in place. I checked various pages and they looked good. The exceptions were the main horizontal navigation and Hide Alerts button. I checked Mayflower locally with both this branch and assets/fonts-add-vi-support and both versions looked the same (see screenshot). Is that something we were looking to address with this change?
Screen Shot 2020-05-29 at 12 56 33 PM

tom-fleming (approved):
@clairesunstudio See my comments in the Conversation tab, but approving since it looks like the behavior between the original PR and hotfix match in Mayflower and Mayflower now loads in Drupal

clairesunstudio:
@tom-fleming, good catch, fixed!

Screen Shot 2020-05-29 at 2 24 41 PM
Screen Shot 2020-05-29 at 1 47 25 PM

clairesunstudio:
hmm it's causing react vrt to fail --> https://27605-137521197-gh.circle-artifacts.com/0/backstop/data/html_report/index.html

I'm going to revert 811eb1a and create a ticket for it to address later

PR #1098, opened 2020-06-23
Assets integration: Fix React ES6 Exports

smurrayatwork:
This updates the ES6 exports to use the full extension in their import/export paths so that importing a component doesn't fall back to the commonjs index.js file. Instead, the ES6 paths should be using the component's index.mjs file.

### DIFF
--- a/react/gulpfile.js
+++ b/react/gulpfile.js
@@ -1,11 +1,13 @@
+import gulp from 'gulp';
+import babel from 'gulp-babel';
+import rename from 'gulp-rename';
+import del from 'del';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
 const {
   src, dest, series, parallel
-} = require('gulp');
-const babel = require('gulp-babel');
-const rename = require('gulp-rename');
-const del = require('del');
-const path = require('path');
-
+} = gulp;
 function clean() {
   return del(['dist']);
 }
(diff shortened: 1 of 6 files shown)


clairesunstudio:
Confirmed full path import is working in import via npm link
`import HeaderSlim from '@massds/mayflower-react/dist/HeaderSlim/index.mjs';`

Can we name es6 index.js and es5 index.cjs, so that import can default to es6 module, e.g. 
`import HeaderSlim from '@massds/mayflower-react/dist/HeaderSlim';`

react/gulpfile.js
```diff
@@ -68,6 +68,31 @@ function resolvePath(sourcePath, currentFile, opts) {
     if (check.test(sourcePath)) {
       const matches = check.exec(sourcePath);
       resolvedPath = `${rootPath}${matches[1]}`;
+      // ES Modules need to list out the extension on the end of the path,
```
clairesunstudio, line 71: Can you explain what this is doing? I looked at index.js in `dist/` before and after and don't see anything different with this change:
https://github.com/massgov/mayflower/pull/1098/files#diff-1ed6f9d93c1ab4ea1481408cc9aab6daR71-R95''

clairesunstudio:
can you also validate that the new changes work in pfml/portal and mayflower/site via npm/yarn link
and feel free to merge

clairesunstudio (approved):

PR #1117, opened 2020-07-10
Removes logo import from SiteLogo component.

smurrayatwork:
Only NarrowTemplate is affected by this change. I added the logo back to NarrowTemplate's story.

### DIFF
--- a/react/src/components/organisms/Header/index.js
+++ b/react/src/components/organisms/Header/index.js
@@ -18,7 +18,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import is from 'is';
 import classNames from 'classnames';
-import logo from '@massds/mayflower-assets/static/images/stateseal.png';
 import UtilityNav from 'MayflowerReactOrganisms/UtilityNav';
 import MainNav from 'MayflowerReactMolecules/MainNav';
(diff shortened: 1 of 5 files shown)


clairesunstudio (commented):
Instead of moving the import into Narrow template, can we just make the image optional?

react/src/components/organisms/Header/index.js
```diff
@@ -295,7 +294,6 @@ Header.defaultProps = {
       domain: 'https://www.mass.gov/'
     },
     image: {
-      src: logo,
       alt: 'Massachusetts state seal',
```
clairesunstudio: can we remove this from default prop? This only makes sense if src is the seal

react/src/components/templates/NarrowTemplate/NarrowTemplate.stories.js
```diff
@@ -13,6 +14,9 @@ storiesOf('templates', module)
       const siteLogoDomainProps = {
         url: {
           domain: text('NarrowTemplate siteLogoDomain: url domain', 'https://www.mass.gov/')
+        },
+        image: {
+          src: logo
```
clairesunstudio:
```suggestion
          src: logo,
          alt: 'Massachusetts state seal'
```

clairesunstudio (approved):

PR #1123, opened 2020-07-17
Mayflower v10: Replace SVG Sprite Loader with SVGR Icons

smurrayatwork:
This PR replaces the `svg-sprite-loader` package with custom generated SVG Icons. Each `.svg` file located under `src/components/base/Icon/assets` will have a React component generated for it and placed inside `dist/Icons`.

Icons may be referenced using either named or default exports:
```
import { Alert, Building } from '@massds/mayflower-react/dist/Icon';
// You can also do this:
import { Alert, Building } from '@massds/mayflower-react';
// And finally default exports:
import Alert from '@massds/mayflower-react/dist/Icon/Alert';
```

Each SVG component supports passing in normal html attributes directly to the `<svg>` element:
```
<Alert width="10" height="10" className="foo" />
```

Please stick to using default exports where possible when updating mayflower-react. Consuming packages may use named or default exports where needed.

In cases where you're unsure which icons your component may need, and you need to load all icons, you can do the following:
```
// Namespaced import. Icon is an array of all the available icons, keyed by the name of the icon file, case sensitive.
import * as Icon from '@massds/mayflower-react/dist/Icon';
```


react/src/components/organisms/TeaserListing/TeaserListing.stories.js
```diff
@@ -85,7 +85,7 @@ storiesOf('organisms/TeaserListing', module)
         usage: 'quaternary-simple',
         capitalized: true,
         expanded: open,
-        icon: <Icon name="chevron" svgWidth={16} svgHeight="16" />
+        icon: <Chevron width={16} height="16" />
```
clairesunstudio:
```suggestion
        icon: <Chevron width={16} height={16} />
```

react/gulpfile.js
```diff
@@ -5,13 +5,13 @@ const babel = require('gulp-babel');
...
 const path = require('path');
-
+const run = require('gulp-run-command').default;
 function clean() {
   return del(['dist']);
 }
 
 function styles() {
-  return src(['./src/components/**/*.scss'])
+  return src(['./src/components/**/*.scss', '!src/components/**/Icon/**'])
```
clairesunstudio, line 14: can you add a comment for this path?

react/src/components/organisms/UtilityNav/index.js
```diff
@@ -118,13 +120,14 @@ const NavItem = (obj) => {
 const NavItemLink = (obj) => {
   const item = obj.data;
   const iconProps = {
-    name: item.icon,
     ariaHidden: true
   };
+  const IconComponent = item.icon === 'building' ? Building : Login;
```
clairesunstudio: not sure why we are only allow these two icon options.. we should make this a rendered prop to pass in whatever Icon, but this is not related and can be in a separate PR.

react/gulpfile.js
```diff
@@ -56,7 +185,8 @@ const sources = [
   '!src/**/*.knobs.options.js',
   '!src/**/*.knob.options.js',
   '!src/**/Colors/**',
-  '!src/**/storyutils.js'
+  '!src/**/storyutils.js',
```
clairesunstudio: I don't see files named storyutils.js

react/icon-template.js
```diff
@@ -0,0 +1,32 @@
+// Generates Icon components using SVGR.
+// https://react-svgr.com/docs/custom-templates/
+function defaultTemplate(
+  { template },
+  opts,
+  { imports, interfaces, componentName, props, jsx, exports },
```
clairesunstudio: Where is `componentName` passed from? 
Can we make the icon component Name more specific to icons, to avoid naming conflicts with other components especially when using named imports.
e.g. `Alert` --> `IconAlert` or `Icon.Alert`

clairesunstudio:
Why is recommended to use default exports for icon imports instead of something like `import { Alert, Building } from '@massds/mayflower-react/dist/Icon';`? 
> Please stick to using default exports where possible when updating mayflower-react. Consuming packages may use named or default exports where needed.

react/icon-template.js
clairesunstudio: instead of 
```
import { Alert } as IconAlert from '@massds/mayflower-react/dist/Icon';
// we can't directly use the Icon Alert if there's another component named Alert
```
If we do `IconAlert`
```
import { IconAlert } from '@massds/mayflower-react/dist/Icon';
<IconAlert />
```

or `Icon.Alert`
```
import Icon from '@massds/mayflower-react/dist/Icon';
<Icon.Alert />
```
not sure if we can do something like this with sub component
```
import {Icon.Alert} from '@massds/mayflower-react/dist/Icon';
<Icon.Alert />
```

clairesunstudio:
A few things to wrap up:
1. lint fixes to pass the ci checks
2. add a changelog following changelogs/template.yml
3. create a PR into https://github.com/EOLWD/pfml/pull/587 with the updated Icon using npm link
4. update Icon imports and usage in mayflower gatsby site

react/icon-template.js
smurrayatwork: The changing of the naming scheme to `Icon.Alert` or `IconAlert` is also not all that unique and is in use by other icon react component packages too. Cases of uniqueness should use something with Mayflower or Mass in the name, or the aliasing as your first code example shows.

clairesunstudio: I think at least we should ensure uniqueness within our library. If they are using both our Icons and icons from another library,  chances are low that they will be using the same icon from multiple different libraries. Right now at least I can see a couple of components within our own library are having naming conflicts

react/.gitignore
```diff
@@ -15,3 +15,9 @@ backstop/data/*report/
 
 .DS_Store
 **/.DS_Store
+src/components/base/Icon/**
+!src/components/base/Icon/assets
+!src/components/base/Icon/Icon.stories.js
+!src/components/base/Icon/IconDisplay.js
+!src/components/base/Icon/Icon.knob.options.js
+!src/components/base/Icon/_icon-display.scss
```
clairesunstudio:
```suggestion
!src/components/base/Icon/_icon-display.scss
!src/components/base/Icon/index.js
```

react/gulpfile.js
smurrayatwork: Good catch. Not sure why I added that...

clairesunstudio:
- [x] change Icon naming to `IconXXXX`
- [ ] add a changelog following changelogs/template.yml
- [x] test build in pfml footer PR
- [x] release alpha.2
- [ ] add Icon development documentation into react/README.md
- [ ] add yalc steps into documentation to replace recommended `npm link` in react/README.md
- [x] lint fixes to pass the ci checks
- [x] update Icon imports and usage in mayflower gatsby site

react/icon-template.js
```diff
@@ -0,0 +1,32 @@
...
+) {
+  const plugins = ['jsx']
+  if (opts.typescript) {
+    plugins.push('typescript')
+  }
+  const typeScriptTpl = template.smart({ plugins })
+  return typeScriptTpl.ast`import React from 'react';
+import PropTypes from 'prop-types';
```
clairesunstudio, line 15: can you format this file with the correct indentations?

clairesunstudio:
Can you fix these visual regressions --> https://29283-137521197-gh.circle-artifacts.com/0/backstop/data/html_report/index.html

clairesunstudio (approved):
Just need to address the VR, otherwise, works great! Feel free to publish alpha.2 and bring that into mayflower sites and pfml, and add changelog and documentation.

changelogs/DP-19414.yml
```diff
@@ -0,0 +1,6 @@
+Changed:
+  - project: React
+    component: Icon
+    description: Icon has been refactored from one component into many icon components. Each .svg icon file is now generated into its own React component at build time with SVGR and SVG Sprite Loader is no longer used.
```
clairesunstudio:
```suggestion
    description: Icon has been refactored from one component into many icon components. Each .svg icon file is now generated into its own React component at build time with SVGR and SVG Sprite Loader is no longer used. (#1123)
```

clairesunstudio (approved):

PR #1129, opened 2020-07-21
use relative links to avoid error withouth sass include path

clairesunstudio:
Currently in mayflower-v10, scss in mayflower-assets package assumes the include paths exist when importing other scss files.
e.g. 
```
@use "00-base/global" as *;
```
This requires the mayflower-assets consuming projects to always have to configure the sass include paths `node_modules/@massds/mayflower-assets/scss`:
e..g 
```
  sassOptions: {
    // It is required to include path @massds/mayflower-assets/scss when referencing a mayflower-assets SCSS file that imports 00-base/global.
    includePaths: [
      path.join(__dirname, "./node_modules/@massds/mayflower-assets/scss"),
    ],
  }
```
See the thread in pfml --> https://github.com/EOLWD/pfml/pull/587#discussion_r455841997

--- 

Requiring include path `node_modules/@massds/mayflower-assets/scss` VS. using the direct relative paths
**Pros:**
- In mayflower-assets package when importing other scss files, don't have to spell out the relative paths based on nested directories
- anything else missing?

**Cons**
- Requires consuming project to configure include path, otherwise will error out

Weighing the current pros and cons, I think it's more important to prioritize consumers and internal development shortcuts. Unless there are other things missing we should include in our considerations.

### DIFF
--- a/assets/scss/01-atoms/_address.scss
+++ b/assets/scss/01-atoms/_address.scss
@@ -1,4 +1,4 @@
-@use "00-base/global" as *;
+@use "../00-base/global" as *;
 
 .ma__address {
 
--- a/assets/scss/01-atoms/_arrow-button.scss
+++ b/assets/scss/01-atoms/_arrow-button.scss
@@ -1,4 +1,4 @@
-@use "00-base/global" as *;
+@use "../00-base/global" as *;
 
 .ma__arrow-button {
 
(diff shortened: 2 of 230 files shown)


clairesunstudio:
https://github.com/EOLWD/pfml/pull/587#discussion_r456655447

clairesunstudio:
See this thread for why we don't want to use relative pathing in scss
https://github.com/EOLWD/pfml/pull/587#discussion_r456655447

PR #1160, opened 2020-08-31
DP-19782 flyout scrolling

ygannett:

## Description

Current issue:  
The behavior is inconsistent when a page has active alerts.  It varies when the alert accordion is open /close and window sizes.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-19782)
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

### DIFF
--- a/patternlab/styleguide/source/assets/js/modules/mainNavHamburger.js
+++ b/patternlab/styleguide/source/assets/js/modules/mainNavHamburger.js
@@ -26,21 +26,23 @@ const utilNavWideCheck = function() {
 /** DP-19336 begin: add padding to hamburger menu to allow scrolling when alerts are loaded */
 const hamburgerMainNav = document.querySelector(".ma__header__hamburger__main-nav");
 let emergencyAlerts = document.querySelector(".ma__emergency-alerts__content");
-
-let alertHeightValue = "";
-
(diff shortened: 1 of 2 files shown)


patternlab/styleguide/source/assets/js/modules/mainNavHamburger.js
```diff
@@ -79,15 +93,14 @@ if (siteAlertWrapper !== null) {
...
     }
+
     hamburgerMenuAlertScrolling();
+
   });
   jsonApiObserver.observe(siteAlertWrapper, {
     childList: true
   });
 }
-else {
```
ygannett, line 88: When no alerts on a page, css takes care of the flyout height.  So, the else part is not necessary.

tom-fleming:
@ygannett Let me know what you think. The primary change I made was to apply the padding to either menu inside the container based on whether the utility menu is visible or not. I don't think we can add padding and set the height to the outer container because of its absolute positioning and other CSS we need to make the menu fly out and stick to the left.

ygannett:
@tom-fleming This works great!  I think this approach is the best to the given situation.

PR #1161, opened 2020-09-01
Allow value 0 to be returned in onBlur callback

clairesunstudio:
Added:
  - project: React
    component: InputCurrency
    description: Fix onBlur bug to allow value 0 to be returned in onBlur callback. (#916)
    impact: Patch

To test:
http://localhost:6006/?path=/story/forms-atoms--inputcurrency
- type in '0' and click outside of the input, onBlur is triggered
Screen Shot 2020-09-01 at 1 31 08 AM
- type in a non-number input and click outside of the input, onBlur is not triggered
Screen Shot 2020-09-01 at 1 31 35 AM

### DIFF
--- a/react/src/components/forms/InputCurrency/index.js
+++ b/react/src/components/forms/InputCurrency/index.js
@@ -140,7 +140,7 @@ const Currency = (props) => {
           const handleBlur = (e) => {
             const { type } = e;
             const inputEl = ref.current;
-            const { value } = inputEl;
+            const value = inputEl && inputEl.value;
             const numberValue = Number(numbro.unformat(value));
             // isNotNumber returns true if value is null, undefined or NaN vs Number.isNaN only checks if value is NaN
             /* eslint-disable-next-line   no-restricted-globals */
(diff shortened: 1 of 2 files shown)


react/src/components/forms/InputCurrency/index.js
```diff
@@ -140,15 +140,15 @@ const Currency = (props) => {
...
             /* eslint-disable-next-line   no-restricted-globals */
-            const isNotNumber = !stringValue || isNaN(Number(numbro.unformat(stringValue)));
+            const isNotNumber = isNaN(value);
             if (isNotNumber) {
               inputEl.setAttribute('placeholder', props.placeholder);
-            }
-            let newValue = isNotNumber ? '' : Number(numbro.unformat(stringValue));
-            if (!is.empty(newValue)) {
+            } else {
+              let newValue = isNotNumber || numberValue;
```
smurrayatwork: In the else, we already know `isNotNumber` is false, so this should be:
```suggestion
              let newValue = numberValue;
```

react/src/components/forms/InputCurrency/index.js
```diff
@@ -140,15 +140,15 @@ const Currency = (props) => {
           const handleBlur = (e) => {
             const { type } = e;
             const inputEl = ref.current;
-            const stringValue = inputEl.value;
-            // isNotNumber returns true if stringValue is null, undefined or 'NaN'
+            const { value } = inputEl;
```
smurrayatwork: If `ref.current` is null, then this line will error out, as you can't destructure null.

smurrayatwork (approved):

PR #1179, opened 2020-09-16
DP-19877: Height of blue global nav bar is not tall enough on small screens

arthurbaghdas:
## Description
Changes global nav blue bar height form 40px into 44px

## Related Issue / Ticket

- [JIRA issue]( https://jira.mass.gov/browse/DP-19877 )


ygannett (approved):
LGTM.

PR #1195, opened 2020-09-30
Mayflower v10  rush monorepo

smurrayatwork:
## Monorepo Package Management
This PR adds support for the creation of a monorepo using [Rush](https://rushjs.io/) and [PNPM](https://pnpm.js.org/). All uses of `npm` or `yarn` have been replaced by `pnpm` which works similarly. For the moment, the following are defined as packages for the mayflower monorepo:

- Mayflower Assets (located under `packages/assets/`)
- Mayflower React (located under `packages/react/`)
- Mayflower Artifacts (located under `packages/patternlab/styleguide/`)
- `packages/site/`

Each of the packages' dependencies from npm are installed using Rush's cli tools. To install packages initially, use the command `rush install` at either the root directory or within any of the package directories. When installing packages using `rush install`, rush will use `pnpm install` within each package's directories. Each dependency is actually installed in a directory (.pnpm) located in your home directory and symlinked to within Rush's internal temporary store at `common/temp/`. Each dependency of a package located under `<package-name>/node_modules/` is actually a symlink to rush's own temporary store at `common/temp/`.

Rush configuration is stored in the root directory as `rush.json`.

Each package also no longer uses package-lock.json or yarn.lock. Instead, a single pnpm-lock.yaml file is made under `common/temp/` which is created/updated when running rush commands.

Rush also either symlinks (by default) or local installs a copy of every package that uses one another from the monorepo. An example of this is Mayflower React. Mayflower React depends on Mayflower Assets, so under Mayflower React's `node_modules/` there is a symlink that points directly to the location of Mayflower Assets on disk instead of the rush temporary store under `common/temp/`.

## Useful Commands

- `rush add` - Use this to add a new dependency to a package.
- `rush install` - Use this to install dependencies for all packages.
- `rush update` - Use this after updating rush/pnpm configurations.
- `rush build` - Use this to make production builds for all packages.

## Other Commands
When needing to run other commands for a package (such as starting up development mode), use `pnpm` instead of `npm` or `yarn` in the package directory. For example, to run development mode for react, instead of using `npm run start`, use `pnpm run start`.


README.md
```diff
@@ -17,6 +17,37 @@ This project is licensed under the is licensed under the GNU General Public Lice
 [react-npm-badge]: https://img.shields.io/npm/v/@massds/mayflower-react.png
 [npm]: https://www.npmjs.com/package/@massds/mayflower-react
 
+## Monorepo Package Management
+This PR adds support for the creation of a monorepo using [Rush](https://rushjs.io/) and [PNPM](https://pnpm.js.org/). To install both globally, run the following command:
```
clairesunstudio:
```suggestion
This monorepo is managed using [Rush](https://rushjs.io/) and [PNPM](https://pnpm.js.org/). To install both globally, run the following command:
```

clairesunstudio:
Works great! I tested the rush commands and package linking in patternlab/styleguide/, react/ and site/ all develop and build properly. 

Change suggestions:

1. I think if we can add some most used scripts at the root (by adding a package.json scripts) so that it's not required to cd into each packages for development. Otherwise, I think we just need to add the steps for quick start local development for each package into README. 
e.g. 

>To start developing mayflower react
>- `rush install`
>- `cd react & pnpm start` (or a pnpm command that can be run at the root)
>
>To start developing mayflower patternlab
>- `rush install`
>- `cd patternlab/styleguide & pnpm run gulp` (or a pnpm command that can be run at the root)
>
>To start developing mayflower site
>- `rush install`
>- `rush build (build local @massds/mayflower-react)`
>- `cd site & pnpm run develop` (or a pnpm command that can be run at the root)

2. Can you also add a high level tree structure for the repo? 

3. Is `pnpm` replacing all `npm` commands, including `npm publish`? or it doesn't make any difference? In circle config, we should remove yarn and npm installs and related commands. 

4. For `rush add`, can you document how to add a dependency scoped to a package? https://rushjs.io/pages/commands/rush_add/ It requires to be in the same directory and use `--package` flag

clairesunstudio:
oh, you also removed the package.json at the root, so the auto release scripts won't work now. We need to add that back and make sure the new structure doesn't disrupt the auto release scripts

clairesunstudio:
`pnpm run start:react` or `cd packages/react & pnpm start` no longer works from the root. can you update README and the package.json at the root to reflect the changes?

clairesunstudio (approved):

PR #1224, opened 2020-10-20
Mayflower v10: display rendered styles

smurrayatwork:
This PR adds the ability to see compiled HTML and CSS for Header, HeaderSlim, Footer, and FooterSlim. Compiled assets are displayed using ReactSyntaxHighlighter, with performance tweaks added using [React Virtualizer](https://github.com/bvaughn/react-virtualized). While the docs pages will load slower in general, without virtualization it is much more of a noticeable delay to the user.

The way this works is by modifying webpack's configurations to only use sass-loader and css-loader on `.scss` files. style-loader has been turned off for all `.scss` files except on storybook's `src/index.scss` file, while MiniCssExtractPlugin is used on production builds. This allows development builds to still include styles as it always has, including using css sourcemaps.

Header, HeaderSlim, Footer, and FooterSlim's stories import in an `_index.scss` file. This returns the file as an object (as processed by both sass-loader and css-loader). This object is then passed to preview.js for use in displaying rendered css styles.

The site logo for the site displayed in both the header and footer is transformed to be a base 64 image, allowing users to copy / paste out the rendered HTML easier. Right now, rendered styles includes fonts, normalize-css, and some other misc. layout files that may or may not be needed for each story.

We still need to come up with a way for the rendered styles to link out to hosted font files. Testing this currently, you can use  the following link (which doesn't use variable fonts): https://fonts.googleapis.com/css2?family=Noto+Sans&amp;display=swap

screencapture-localhost-6006-iframe-html-2020-10-20-16_57_28


packages/react/src/components/organisms/Header/_index.scss
```diff
@@ -0,0 +1,19 @@
+@use "~normalize-scss/sass/normalize";
```
clairesunstudio: Can you separate out the global scss like normalize, elements, layout, fonts into a different scss file? And import it separately in the stories.js?

packages/react/src/components/organisms/Header/_index.scss
```diff
@@ -0,0 +1,19 @@
...
+@use '../../molecules/HeaderSearch';
+@use '../../molecules/MainNav';
+@use "00-base/elements";
+@use "00-base/layout";
+@use "00-base/fonts";
+@use '01-atoms/decorative-link';
+@use '01-atoms/button-with-icon';
+@use '01-atoms/button-search';
+@use '01-atoms/input-typeahead';
+@use '01-atoms/site-logo';
```
clairesunstudio, line 7: Ideally, I think these scss for SiteLogo, HeaderSearch, UtilityNav scss should be imported in the respective index.scss for those specific components and then imported here. So that the lower-level component scss indexes are included in the upper-level component scss indexes

packages/react/src/components/organisms/Header/Header.stories.js
```diff
@@ -6,6 +6,7 @@ import MainNavData from 'MayflowerReactMolecules/MainNav/MainNav.knob.options';
 import UtilityNavData from 'MayflowerReactOrganisms/UtilityNav/UtilityNav.knob.options';
 import Header from './index';
 import HeaderDocs from './Header.md';
+import styles from './_index.scss';
```
clairesunstudio, line 9: Once the global scss is broken out in scss
```
import componentStyles from './_index.scss';
import globalStyles from '../index.scss';
```
only show component css in storybook
```
<StoryPage styles={componentStyles} Description={HeaderDocs} />
```

Do we still need `styles/_index.scss`? If not, can we consolidate `components/index.scss` with `components/styles/index.scss`?

packages/react/.storybook/syntax-highlighter.js
```diff
@@ -0,0 +1,173 @@
+import React from 'react';
+import { AutoSizer, List } from 'react-virtualized';
+import { PrismLight as ReactSyntaxHighlighter } from 'react-syntax-highlighter';
+import { ActionBar, createSyntaxHighlighterElement, EmptyBlock  } from '@storybook/components';
+import { styled, ThemeProvider, convert, themes, Global, getScrollAreaStyles } from '@storybook/theming';
+import createElement from 'react-syntax-highlighter/dist/esm/create-element';
+import { okaidia } from 'react-syntax-highlighter/dist/esm/styles/prism';
```
clairesunstudio: is this being used?

packages/react/src/components/organisms/Header/_index.scss
smurrayatwork: I think that the contents of this _index.scss file should live in assets as 03-organisms/header.scss. Then, this file should only be:
```
@use '03-organisms/header';
```
We should not be combining `_index.scss` files from the react package together to make a story's rendered and displayed css files. Instead, we should be pulling scss files from the assets package into `_index.scss` for each story.

packages/react/.storybook/preview.js
```diff
@@ -33,46 +37,60 @@ const storyKindOrder = [
...
-        htmlWhitespaceSensitivity: 'ignore',
-        endOfLine: 'auto',
-        parser: 'html',
-        plugins: [parserHtml]
-      }
-    );
-  }
+  
+  let html = ReactDOMServer
+  .renderToStaticMarkup(
```
clairesunstudio: should we use `useMemo` for html rendering too?

packages/react/.storybook/syntax-highlighter.js
smurrayatwork: Good catch, no it is not. I'll remove.

packages/react/src/components/organisms/Header/_index.scss
smurrayatwork: This wouldn't be imported in the story itself, it'd be imported in the scss file.

packages/react/.storybook/preview.js
```diff
@@ -15,12 +15,16 @@ import {
...
 import prettier from 'prettier/standalone';
 import parserHtml from 'prettier/parser-html';
-import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
-import { okaidia } from 'react-syntax-highlighter/dist/esm/styles/prism';
+import parserCss from 'prettier/parser-postcss';
+import SyntaxHighlighter, { Renderer, Wrapper } from './syntax-highlighter';
+
 import '../src/index.scss';
 
+import logo from '!url-loader!@massds/mayflower-assets/static/images/stateseal.png';
```
clairesunstudio, line 25: I think we should use the UNPKG CDN for linking the assets (pending on the mayflower-assets release)

packages/react/.storybook/preview.js
```diff
@@ -15,12 +15,16 @@ import {
   Canvas, Story
 } from '@storybook/addon-docs/blocks';
 import { ActionBar, Source } from '@storybook/components';
+
 import prettier from 'prettier/standalone';
 import parserHtml from 'prettier/parser-html';
-import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
-import { okaidia } from 'react-syntax-highlighter/dist/esm/styles/prism';
+import parserCss from 'prettier/parser-postcss';
```
clairesunstudio:
```suggestion
```
this is no longer needed right?

clairesunstudio (approved):

PR #1247, opened 2020-10-28
Core/storybook use env var for CDN path versioning

clairesunstudio:
We should version control core documentation and assets CDN to prevent breakage for the mayflower markup consumers

This PR explores using the environment variable for controlling the version variable in the CDN paths. We should find a way to auto-update this variable during releases.


clairesunstudio:
See this error when using Story in .mdx
Screen Shot 2020-11-03 at 5 14 21 PM

Seems to be related to multiple react versions with pnpm
```
storyStore gets passed in by a react context, and gets consumed by a useContext hook, and I've observed that hooks fail if there are multiple versions of react.

```

https://github.com/storybookjs/storybook/issues/6766#issuecomment-641611970

smurrayatwork (approved):

PR #1254, opened 2020-11-02
change path to be relative to current directory

clairesunstudio:
On v10.1.0 production sites, the production urls for the sitelogo are relative to the root path. Fixing the path in patternlab to allow site logo to render in all places.
Screen Shot 2020-11-02 at 10 40 12 AM
Screen Shot 2020-11-02 at 10 40 39 AM

https://mayflower.digital.mass.gov/patternlab/
https://mayflower.digital.mass.gov/patternlab/v/10.1.0

PR #1255, opened 2020-11-02
Release 10.1.1

clairesunstudio:
### Fixed 
- (Patternlab) [CircleCI] Fix patternlab deploy tag paths. (#1252)
- (Patternlab) [SiteLogo] Set Patternlab SiteLogo global path in patternlab to be relative to the current path in order to fix production urls. (#1254)


.circleci/config.yml
```diff
@@ -24,11 +24,11 @@ patternlab_defaults: &patternlab_defaults
 backstop_defaults: &backstop_defaults
   working_directory: /home/circleci
   docker:
-    - image: circleci/python:2.7.14
+    - image: massgov/drupal-container:1.0.8-ci
```
```suggestion
    - image: cimg/node:12.14
```

.circleci/config.yml
```diff
@@ -165,31 +159,19 @@ jobs:
...
+            aws s3 sync packages/patternlab/styleguide/public "s3://mayflower.digital.mass.gov/patternlab/v/$CIRCLE_TAG"
             if [[ "$CIRCLE_TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
               # Use cp instead of sync for root copy.  sync lists all keys and is too slow.
-              aws s3 cp --recursive patternlab/styleguide/public/ "s3://mayflower.digital.mass.gov/patternlab"
+              aws s3 cp --recursive packages/patternlab/styleguide/public/ "s3://mayflower.digital.mass.gov/patternlab"
             fi
 
   react_build_storybook:
-    <<: *react_defaults
+    <<: *patternlab_defaults
```
clairesunstudio, line 169: why does this one need to use php?

smurrayatwork: I want to keep all the jobs on the same images as possible, due to how circleci cache and workspace mounting works.

When you save cache, it saves to the path of the working_directory. react_defaults and patternalb_defaults have different paths, so you can't restore cache to both using the cache created by the `build` job. Having all jobs that need cache on the same image and working_directory fixes this issue. Since it's easier to fix anything missing from node vs anything missing from php on circleci, I chose the php image. That image has node available anyway.

Additionally since we now use rush, and rush installs all dependencies for all packages, I can cache on build early.

clairesunstudio: got it, thanks! maybe we should rename the images in the future

PR #1260, opened 2020-11-10
Updates Rush, PNPM versions.

smurrayatwork:
This PR just updates Rush and PNPM to their latest stable versions. It also bumps the allowed NodeJS version to be between 12.13 and 14.x.


common/scripts/install-run.js
```diff
@@ -1,7 +1,27 @@
 "use strict";
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See the @microsoft/rush package's LICENSE file for license information.
+var __createBinding = (this && this.__createBinding) || (Object.create ? (function(o, m, k, k2) {
```
clairesunstudio, line 4: Are these new lines auto-generated from the new version?

rush.json
```diff
@@ -121,7 +121,7 @@
    * Specify a SemVer range to ensure developers use a Node.js version that is appropriate
    * for your repo.
    */
-  "nodeSupportedVersionRange": ">=12.13.0 <13.0.0",
+  "nodeSupportedVersionRange": ">=12.13.0 <15.0.0",
```
clairesunstudio, line 124: validated on node v12.18.0
Did you test if v14 worked? my zsh export PATH is having issues locating rush on v14 `command not found: rush`

clairesunstudio (approved):

common/scripts/install-run.js
smurrayatwork: Yes

rush.json
smurrayatwork: Currently we can't, but in the future we can.

PR #1267, opened 2020-11-12
Mayflower Core / Mayflower Assets: Generate Static HTML, JS for Core

smurrayatwork:
This adds the ability to package up static HTML and its corresponding JS for the following components:

- Header
- Header Slim
- Header Hamburger
- Header Mixed
- Footer
- Footer Slim

Using the static assets placed in the Mayflower Assets package, you should be able to copy / paste and have a working header or footer for a website.

The static HTML I placed in Mayflower Assets was generated by Patternlab, but I could also use the same HTML generated by Storybook and get the same results.

The static JS comes transpiling together scripts from patternlab's js assets.

I've provided normal and minimized versions of all JS files.

Lastly, I updated the pathing for build in Mayflower assets. Now there's both `build/scss` and `build/js` directories.


.circleci/config.yml
```diff
@@ -291,6 +291,7 @@ jobs:
       - restore_cache:
           keys:
             - root-npm-v4-{{ checksum "common/config/rush/pnpm-lock.yaml" }}-12
+      - run: { name: 'Build', command: 'cd packages/assets && node ../../../common/scripts/install-run-rushx.js build'}
```
clairesunstudio: https://app.circleci.com/pipelines/github/massgov/mayflower/1557/workflows/888d4a2a-ded0-4757-b69b-ac1b59b555dc/jobs/33453

```suggestion
      - run: { name: 'Build', command: 'cd packages/assets && node ../../common/scripts/install-run-rushx.js build'}
```

packages/assets/build/scss/global.scss
```diff
@@ -4,6 +4,10 @@
 
 @forward "layout";
 
+@forward "paths" with (
+  $assets-path: "https://unpkg.com/@massds/mayflower-assets@10.1.0/static"
```
clairesunstudio: can the version be updated by the latest release version?

clairesunstudio:
For now, can we also compile JS for (header)[https://mayflower.digital.mass.gov/patternlab/?p=organisms-header], since we publish header.css. We can deprecate Header and replace it with HeaderMixed once we have the react version built

clairesunstudio:
Tested compiled JS and works! 

Header Mixed: https://codepen.io/clairesun/project/editor/ZWkrBa#
Header Hamburger: https://codepen.io/clairesun/project/editor/ZWkrBa#

How is the JQuery in header mixed imported?

packages/assets/build/scss/global.scss
clairesunstudio: If we can figure out a way to do that, we can also link to the seal image CDN https://unpkg.com/@massds/mayflower-assets@[version]/static/images/logo/stateseal.png instead of using base64 embed in react headers?

clairesunstudio:
Should we commit the JS files? They are not the source files. I feel like it makes sense to only commit the source files

Also, currently, the HTML files are manually copied over? Are you thinking these files will be published in the assets package temporarily until we have a way to generate in the core storybook? It that's the case, I don't think we should commit those files either

packages/assets/build/scss/global.scss
smurrayatwork:
```suggestion
  $assets-path: "https://unpkg.com/@massds/mayflower-assets/static"
```

smurrayatwork:
@clairesunstudio I didn't commit the js files for the same reasoning that we don't commit the css files: they're generated on build based on the js files from the patternlab package.

As for the HTML files, those were included as the minimum needed HTML for testing. I can remove as that should come from what's in storybook / patternlab.

packages/assets/build/scss/global.scss
smurrayatwork: I've committed a fix to always point the version to the latest published.

clairesunstudio:
> @clairesunstudio I didn't commit the js files for the same reasoning that we don't commit the css files: they're generated on build based on the js files from the patternlab package.

I'm seeing the JS files committed in this PR though

smurrayatwork:
> > @clairesunstudio I didn't commit the js files for the same reasoning that we don't commit the css files: they're generated on build based on the js files from the patternlab package.
> 
> I'm seeing the JS files committed in this PR though

You're right; I'll remove.

packages/assets/gulpfile.js
```diff
@@ -53,10 +64,104 @@ function compileMiniScss() {
...
+    .pipe(dest('./js'));
+}
+
+function compileMainNav() {
+  return src([
+    '../patternlab/styleguide/source/assets/js/modules/mainNav.js'
+  ])
+  .pipe(babel({
+    plugins: [
+      'babel-plugin-rewire-exports',
```
clairesunstudio, line 104: why mainNav.js needs to be transpiled differently?

smurrayatwork: This is because I need to remove the export statement only for this file while preserving the IIFE syntax.

changelogs/core-header-footer.yml
```diff
@@ -0,0 +1,47 @@
...
+    issue:
+    impact: Minor
+  - project: Patternlab
+    component: js/modules/mainNav.js
+    description: Changed main nav js for window resize to use addEventListener instead of undefined addEventHandler. (#1267)
+    issue:
+    impact: Patch
+  - project: Patternlab
+    component: js/modules/mainNavHamburger.js
+    description: Fixed a bug causing non-smooth scrolling in browser when clicking hamburger nav items. (#1267)
```
clairesunstudio: what is this bug? I'm not seeing that in the prod site https://mayflower.digital.mass.gov/patternlab/?p=organisms-header-hamburger

clairesunstudio:
Two last comments:
1. build/js/header-hamburger-vendor.js should be committed 
2. I thought HeaderHamburger doesn't need jquery, is that right?

clairesunstudio (approved):

packages/assets/package.json
```diff
@@ -6,7 +6,9 @@
   "files": [
     "static",
     "scss",
-    "css"
+    "css",
+    "js",
+    "html"
```
clairesunstudio:
```suggestion
```

packages/assets/package.json
```diff
@@ -6,7 +6,9 @@
   "files": [
     "static",
     "scss",
-    "css"
+    "css",
+    "js",
```
clairesunstudio:
```suggestion
    "js"
```

packages/assets/package.json
clairesunstudio: remove html from package

PR #1273, opened 2020-11-19
Assets/add seal variants and svg

clairesunstudio:
Changed:
  - project: Assets
    component: Logos
    description: Optimize and standardize existing state seals PNGs in assets to significantly reduce file sizes for web usage without sacrificing qualities
Added PNG and SVG formats for all official state seal variations. (#1273)
    issue:
    impact: Patch
Added:
  - project: Assets
    component: Logos
    description: Added PNG and SVG formats for all official state seal variations: `stateseal.[png|svg]`, `stateseal-color.[png|svg]`, `stateseal-black.[png|svg]`, `stateseal-white.[png|svg]`. (#1273)
    issue:
    impact: Minor


tkoleary (changes requested):
They look good to me. 
Only one question. Do we need different one-color versions or could we change colors with CSS?
(obviously only applies to the SVGs)

clairesunstudio:
@tkoleary good question! we are currently not using the svg versions in the mayflower components, we are only using the png version, maybe it's an option we should consider adding to the SiteLogo component. 

For the purpose of this PR, I want to make the SVG options available on the mayflower CDN for all usage, not just for the web.  So I think it's helpful to save those files separately, so users can download them directly without needing to change anything.

tkoleary (approved):
Good point.

PR #1274, opened 2020-11-20
Core/auto update version env var

clairesunstudio:
This PR is to add the update core version task into the auto release script, so that the mayflower CDN version in the core documentation is updated during the release process.

To test:
`npm install -g js-yaml@3.13.1 @octokit/rest semver@6.3.0 shelljs@0.8.3 simple-git@1.126.0`
`export NODE_PATH=$(npm root --quiet -g) && node scripts/release-branch.js`

PR #1285, opened 2020-12-01
Core/template docs

clairesunstudio:
Added template documentation (will be replacing the starter templates in https://github.com/massgov/mayflower-starters)

Screen Shot 2020-12-02 at 12 29 14 PM

Code test playground --> https://jsfiddle.net/qfb7z2kt/
> Note that 10.3.0 don't have general.css and layout.css in the CDN, once  #1274 gets in and we do a release the resources urls will get updated to the correct version

### DIFF
--- a/packages/core/stories/styles/index.scss
+++ b/packages/core/stories/styles/index.scss
@@ -39,7 +39,7 @@
 @use '@massds/mayflower-assets/scss/02-molecules/footer-links';
 @use '@massds/mayflower-assets/scss/02-molecules/social-links';
 
-@use '@massds/mayflower-assets/css/layout.css';
+@use '@massds/mayflower-assets/scss/00-base/layout';
 @use '@massds/mayflower-assets/scss/01-atoms/place-holder';
 
 // Global
--- a/packages/core/stories/styles/index.scss
+++ b/packages/core/stories/styles/index.scss
@@ -39,6 +39,8 @@
 @use '@massds/mayflower-assets/scss/02-molecules/footer-links';
 @use '@massds/mayflower-assets/scss/02-molecules/social-links';
 
+@use '@massds/mayflower-assets/css/layout.css';
+
 // Global
 .sp {
   &--top {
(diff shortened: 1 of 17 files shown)


packages/core/stories/styles/index.scss
```diff
@@ -39,6 +39,9 @@
 @use '@massds/mayflower-assets/scss/02-molecules/footer-links';
 @use '@massds/mayflower-assets/scss/02-molecules/social-links';
 
+@use '@massds/mayflower-assets/css/layout.css';
+@use '~@massds/mayflower-assets/scss/01-atoms/place-holder';
```
smurrayatwork: Why does this line need the `~` but the others do not?

clairesunstudio:
```suggestion
@use '@massds/mayflower-assets/scss/01-atoms/place-holder';
```

clairesunstudio: oh it doesn't need `~` because we have included node_modules path already

packages/core/stories/styles/index.scss
```diff
@@ -39,6 +39,9 @@
 @use '@massds/mayflower-assets/scss/02-molecules/footer-links';
 @use '@massds/mayflower-assets/scss/02-molecules/social-links';
 
+@use '@massds/mayflower-assets/css/layout.css';
```
tkoleary: shouldn't this be `'@massds/mayflower-assets/scss/00-base/layout'` ?

tkoleary (changes requested):
See comment on index.scss. That broke my build when I pulled the branch.

smurrayatwork (approved):

clairesunstudio:
@tkoleary I was testing the compiled css, I think should be the same as layout.scss let me update. Did using .css break development for you?

update: oh you would need to build assets `rush build:assets` to build the css first. I will point to .scss instead

packages/core/stories/styles/index.scss
clairesunstudio:
```suggestion
@use '@massds/mayflower-assets/scss/00-base/layout';
```

tkoleary (approved):

PR #1293, opened 2020-12-09
Core/bug fixes

tkoleary:
## Description
Fixed a couple of incorrect icon names that kept popping up as warnings in the console

## Related Issue / Ticket
- [JIRA issue](none)

## Steps to Test
-Run rush start core
-open localhost 6006
-Check console for warnings

### DIFF
--- a/packages/react/src/components/organisms/UtilityNav/UtilityNav.knob.options.js
+++ b/packages/react/src/components/organisms/UtilityNav/UtilityNav.knob.options.js
@@ -2,7 +2,7 @@ export default {
   items: [{
     text: 'All Budgets',
     ariaLabelText: '',
-    icon: 'IconBuilding',
+    icon: 'Building',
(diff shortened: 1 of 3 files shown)


clairesunstudio:
Screen Shot 2020-12-09 at 3 34 23 PM
 I'm still seeing the console errors ^

packages/react/src/components/organisms/UtilityNav/UtilityNav.knob.options.js
```diff
@@ -7,7 +7,7 @@ export default {
   }, {
     text: 'State Organizations',
     ariaLabelText: '',
-    icon: 'building',
+    icon: 'Building',
```
tkoleary, line 10: I just realized this is wrong. It should be lower case.

clairesunstudio:
Some of these errors should be resolved from steven's header/nav refactor work

PR #1295, opened 2020-12-09
Core/storybook auto deploy

clairesunstudio:
Auto deploy to S3 on mayflower releases:

https://mayflower.digital.mass.gov/core/
https://mayflower.digital.mass.gov/core/v/[VERSION]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -419,7 +419,7 @@ workflows:
       - react_build_storybook:
           requires: [build]
           filters:
-            branches: { ignore: /(docs|patternlab\/|site\/|core\/).*/ }
+            branches: { ignore: /(docs|patternlab\/|site\/).*/ }
       - react_test:
           requires: [react_build_storybook]
           filters:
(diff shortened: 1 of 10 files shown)


clairesunstudio:
This setup will not remove old hashed files from the root. To do that, we can set up a lambda function that reroutes the root `/core` url to the latest version url `/core/v/LATEST`, and stop uploading files to the root.

This is still WIP:
- In Lambda, we want to list all the object names from the mayflower.digital.mass.gov bucket under `/core/v` path
- Create a function to get the latest version 
- Redirects `mayflower.digital.mass.gov/core` to `mayflower.digital.mass.gov/core/v/LATEST`

.circleci/config.yml
```diff
@@ -391,17 +425,21 @@ workflows:
           requires: [react_build_storybook]
           filters:
             branches: { ignore: /(docs|patternlab\/|react\/|core\/).*/ }
+     # core
+      - core_build_storybook:
+          requires: [react_build_storybook]
```
smurrayatwork, line 430: I don't think this should depend on react_build_storybook unless it uses something from it?

.circleci/config.yml
```diff
@@ -468,3 +506,12 @@ workflows:
           filters:
             branches: { ignore: /.*/ }
             tags: { only: /.*/ }
+      - core_build_storybook:
+          requires: [react_build_storybook]
```
smurrayatwork, line 510: Same comment as above - this isn't needed here.

PR #1299, opened 2020-12-11
Core/base elements content only

tkoleary:
## Description
Removed added markup and layout changes and the related CSS

## Related Issue / Ticket
- [JIRA issue](DP-20309)

## Steps to Test
1. Switch the branch
2. Run `rush start:core`
3. Go to docs/elements and compare to screenshots

## Screenshots
Screen Shot 2020-12-11 at 10 51 19 AM
Screen Shot 2020-12-11 at 10 51 08 AM
Screen Shot 2020-12-11 at 10 50 57 AM

### DIFF
--- a/packages/core/stories/elements/Button/Button.stories.mdx
+++ b/packages/core/stories/elements/Button/Button.stories.mdx
@@ -13,7 +13,7 @@ import * as stories from './Button.stories.js';
 
 # Buttons
 
-Buttons are used in forms and interactive applications, but also as prominent call-to-action links on website pages.
+Buttons are used in forms and interactive applications to trigger actions.
 
 <Canvas withSource="none">
   <Story
--- a/packages/core/stories/elements/Button/Button.stories.mdx
+++ b/packages/core/stories/elements/Button/Button.stories.mdx
@@ -15,7 +15,7 @@ import * as stories from './Button.stories.js';
 
 Buttons are used in forms and interactive applications, but also as prominent call-to-action links on website pages.
 
-<Canvas>
+<Canvas withSource="none">
   <Story
 		name="button-primary"
     parameters={{ docs: { transformSource: () => renderReactMarkup(<Button usage="" text="Button"/>) } }}>
@@ -23,15 +23,14 @@ Buttons are used in forms and interactive applications, but also as prominent ca
   </Story>
 </Canvas>
 
-Following web best practices buttons have a target area of 45 pixels. To provide more visual affordance that they are clickable—not just text in a box—they have a shadow to appear slightly raised. Small buttons can be used for secondary tasks on forms—for example image upload—and filters on listing pages. Large buttons should be used for form submits or primary calls to action. There should usually be only one  primary button in any UI. Font sizes reduce slightly on mobile.
+Following web best practices buttons have a target area of 45 pixels. To provide more visual affordance that they are clickable—not just text in a box—they have a shadow to appear slightly raised. Small buttons can be used for secondary tasks on forms—for example image upload—and filters on listing pages. Large buttons should be used for form submits or primary calls to action. There should usually be only one primary button in any UI. Font sizes reduce slightly on mobile.
 
 ---
 
 ### Primary Buttons
 
-**Usage:** For the main action in a form, or for a call-to-action link on a page.
-
-**Style:** Default buttons are solid color with contrasting uppercase text. For actions with text longer than two words use the `title-case` class. On hover the Background color becomes lighter.
+- **Usage:** For the main action in a form, or for a call-to-action link on a page.
+- **Style:** Default buttons are solid color with contrasting uppercase text. For actions with text longer than two words use the `title-case` class. On hover the Background color becomes lighter.
 
 <Canvas>
   <Story
@@ -61,9 +60,8 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 
 ### Secondary Buttons
 
-**Usage:** For secondary actions.
-
-**Style:** The outline button has less visual prominence than the primary button so it can be used alongside it as a secondary action, for example "Cancel". The hover state of the secondary button inverts the background and foreground colors.
+- **Usage:** For secondary actions.
+- **Style:** The outline button has less visual prominence than the primary button so it can be used alongside it as a secondary action, for example "Cancel". The hover state of the secondary button inverts the background and foreground colors.
 
 <Canvas>
   <Story
@@ -93,9 +91,8 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 
 ### Tertiary Buttons
 
-**Usage:** For third-level actions on a page.
-
-**Style:** The gray color is even less prominent than outline button to de-emphasize third-level tasks. The hover state inverts the background and text colors.
+- **Usage:** For third-level actions on a page.
+- **Style:** The gray color is even less prominent than outline button to de-emphasize third-level tasks. The hover state inverts the background and text colors.
 
 <Canvas>
   <Story
@@ -110,9 +107,8 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 
 ### Quaternary Buttons
 
-**Usage:** For fourth-level actions or actions.
-
-**Style:** Appears as a link.
+- **Usage:** For fourth-level actions or actions.
+- **Style:** Appears as a link.
 
 <Canvas>
   <Story
@@ -127,9 +123,8 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 
 ### Disabled Buttons
 
-**Usage:** Disabled buttons should not be clickable. The disabled attribute (as opposed to hidden) can be applied for actions where the user should see the action and it's state but may not be able to perform it because they either don't have permission or some other condition in the form has not been met.
-
-**Style:** Disabled buttons have no box shadow (since the box shadow is what suggests clickability). Their opacity is `0.5.`, no hover or focus states are provided and the cursor is set to `not-allowed`.
+- **Usage:** Disabled buttons should not be clickable. The disabled attribute (as opposed to hidden) can be applied for actions where the user should see the action and it's state but may not be able to perform it because they either don't have permission or some other condition in the form has not been met.
+- **Style:** Disabled buttons have no box shadow (since the box shadow is what suggests clickability). Their opacity is `0.5.`, no hover or focus states are provided and the cursor is set to `not-allowed`.
 
 
 <Canvas>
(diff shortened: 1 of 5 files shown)


packages/core/stories/elements/Button/Button.stories.mdx
```diff
@@ -12,160 +12,153 @@ import * as stories from './Button.stories.js';
...
 
-<hr className="sp--top" />
+---
 
-### Tertiary Button
-**Usage:** For the third action on a page.
+### Tertiary Buttons
 
-**Style:** This is a solid gray button with a lighter background color than the foreground text. The muted color makes this button less prominent on a page than the secondary outline button a providing an option for a tertiary action. The hover state of this button will invert background color with white color text.
+- **Usage:** For third-level actions on a page.
```
clairesunstudio:
```suggestion
- **Usage:** For third-level actions.
```

packages/core/stories/elements/Button/Button.stories.mdx
```diff
@@ -12,160 +12,153 @@ import * as stories from './Button.stories.js';
 />
 
 # Buttons
-Buttons express what action will occur when the user clicks or touches it. Buttons are used to initialize an action, either in the background or foreground of an experience.
-<Canvas withSource="none" >
+
+Buttons are used in forms and interactive applications, but also as prominent call-to-action links on website pages.
```
clairesunstudio: So here I'm thinking whether we should make the distinction between buttons and links. In the Link doc, there's a variant called "Link styled as a button", meaning button "styles" are allowed to be used as links (whether or not it's a good practice it's up for debate).  Technically, a button should only be used to trigger actions vs links are used for navigation. 
```suggestion
Buttons are used in forms and interactive applications to trigger actions.

Comment (clairesunstudio) [reply packages/core/stories/elements/Button/Button.stories.mdx]: What I'm trying to say here is whether this doc is about the element button or the styles of buttons — if we are keeping "button" just as the element itself, then I think to distinguish it from links will make more sense.

File: packages/core/stories/elements/Button/Button.stories.mdx [hunk 3]
```diff
@@ -12,160 +12,153 @@ import * as stories from './Button.stories.js';
...
+---
 
-## Variations
-<br />
+### Primary Buttons
 
-### Primary Button
-**Usage:** For the principle call to action on the page.
+- **Usage:** For the main action in a form, or for a call-to-action link on a page.
+- **Style:** Default buttons are solid color with contrasting uppercase text. For actions with text longer than two words use the `title-case` class. On hover the Background color becomes lighter.
```
clairesunstudio: so the "title-case" option is not currently available to Button, it is though in ButtonWithIcon in react (but not in patternlab), which we should make available to all buttons. These are the things I was talking about that we need to follow up on refactoring and consolidating in code. 
```suggestion
- **Style:** Default buttons are solid color with contrasting uppercase text. For actions with text longer than two words use the title case instead of all caps. On hover, the Background color becomes lighter.
```

packages/core/stories/elements/Button/Button.stories.mdx
```diff
@@ -12,160 +12,153 @@ import * as stories from './Button.stories.js';
...
-<hr className="sp--top" />
+---
 
-### Tertiary Button
-**Usage:** For the third action on a page.
+### Tertiary Buttons
 
-**Style:** This is a solid gray button with a lighter background color than the foreground text. The muted color makes this button less prominent on a page than the secondary outline button a providing an option for a tertiary action. The hover state of this button will invert background color with white color text.
+- **Usage:** For third-level actions on a page.
+- **Style:** The gray color is even less prominent than outline button to de-emphasize third-level tasks. The hover state inverts the background and text colors.
```
clairesunstudio:
```suggestion
- **Style:** The gray color is even less prominent than outline button to de-emphasize third-level tasks. Background color of the primary button lightens when being hovered over.
```

Sorry, these are the things I missed in the first draft.

packages/core/stories/elements/Link/Link.stories.mdx
```diff
@@ -14,27 +14,28 @@ import * as stories from './Link.stories.js';
...
-#### Inline Link
-**Usage:** Use when you are linking just a few words of text or when you are doing inline links.
+### Inline Link
 
-**Style:** [To-do: Regular inline links should also have underline decoration.. Look into generating the appropriate CSS]
+- **Usage:** Use when you are linking just a few words of text or when you are doing inline links.
+- **Style:** Inline links take the style of the surrounding text, adding only the blue color and an underline.
 
-<Canvas>
+<Canvas className="borderless">
```
clairesunstudio:
```suggestion
<Canvas>
```

clairesunstudio: not used

packages/core/stories/elements/Link/Link.stories.mdx
```diff
@@ -14,27 +14,28 @@ import * as stories from './Link.stories.js';
...
 
-#### Decorative Link
-**Usage:** Used for standalone links that require more emphasis.
+---
 
-**Style:**
+### Decorative Link
+
+- **Usage:** Used for standalone links that require more emphasis, usually inside `ma__section-links`.
+- **Style:** the decorative link is slightly larger in size and appended with an arrow icon.
```
clairesunstudio, line 38: Can we also add the hover state to "Style" for decorative link

packages/core/stories/elements/Link/Link.stories.mdx
```diff
@@ -61,34 +63,37 @@ Links are used to embed actions or pathways to more information in a sentence or
...
 </Canvas>
 
+---
 
-#### Link styled as a button
-**Usage:** Use when you are linking just a few words of text or when you are doing inline links. In general, we should style buttons as button and style links as links, to provide . [*For material honesty, we shouldn't encourage styling links as buttons, should use callout links instead, unless we can come up with appropriate examples*]
+### Link styled as a button
 
-**Style:**
+- **Usage:** If absolutely necessary this can be used for calls-to-action if there are technical reasons not to use `ma__button`.
```
clairesunstudio: So "Link as a button" is using the `ma__button` className (expand the to show the HTML snippet), the difference here is that it's actually an anchor link `<a>` not a `<button>` (again, technically different elements).  Although there are use cases of in mayflower implementations, my opinion is that we should try to differentiate the styling between buttons and links, as they function differently, and that will help set expectations of the perceived behaviors.

packages/core/stories/styles/index.scss
```diff
@@ -96,6 +97,11 @@ ul.sg-colors {
   }
 }
 
+// Links story
+span.ma__decorative-link  {
+	display: inline !important;
```
clairesunstudio:
```suggestion
	display: inline;
```

hmm interesting, I think when I refactor the stories will resolve this. But for now I tested, the important flag is not necessary.

packages/core/stories/elements/Button/Button.stories.mdx
clairesunstudio: for now, a class is not available, so let's just describe the rule (and add that as a to-do in code)

packages/core/stories/elements/Link/Link.stories.mdx
tkoleary: That makes sense

packages/core/stories/styles/index.scss
tkoleary: Great!

packages/core/stories/elements/Button/Button.stories.mdx
tkoleary: Ah, right. I would maybe word it a bit differently:
"The hover state lightens the primary button background color."

packages/core/stories/elements/Link/Link.stories.mdx
tkoleary: Sure

packages/core/stories/elements/Button/Button.stories.mdx
clairesunstudio: oh sorry, this is not the primary button, but the hover animation is similar to the primary button, so I just copied from the primary. This is actually the tertiary button

packages/core/stories/elements/Button/Button.stories.mdx
```diff
@@ -30,7 +30,7 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 ### Primary Buttons
 
 - **Usage:** For the main action in a form, or for a call-to-action link on a page.
-- **Style:** Default buttons are solid color with contrasting uppercase text. For actions with text longer than two words use the title case instead of all caps. On hover, the Background color becomes lighter.
+- **Style:** Default buttons are solid color with contrasting uppercase text. THe hover state lightens the primary button background color.
```
tkoleary: For consistency with the wording below.

packages/core/stories/elements/Button/Button.stories.mdx
```diff
@@ -92,7 +92,7 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 ### Tertiary Buttons
 
 - **Usage:** For third-level actions.
-- **Style:** The gray color is even less prominent than outline button to de-emphasize third-level tasks. The hover state inverts the background and text colors.
+- **Style:** The gray color is even less prominent than outline button to de-emphasize third-level tasks. The hover state lightens the primary button background color.
```
clairesunstudio: I guess should be tertiary instead

packages/core/stories/elements/Link/Link.stories.mdx
clairesunstudio:
```suggestion
- **Usage:** If absolutely necessary this can be used for calls-to-action if there are good design reasons not to use Callout Link. In general, links should be styled differently than buttons, to help set expectations of their different behaviors.
```

packages/core/stories/elements/Button/Button.stories.mdx
```diff
@@ -108,7 +108,7 @@ Following web best practices buttons have a target area of 45 pixels. To provide
 ### Quaternary Buttons
 
 - **Usage:** For fourth-level actions or actions.
-- **Style:** Appears as a link.
+- **Style:** Appears as a link, with a small amount of letterspacing and underline (Actually a `border-bottom` not a text underline). The hover state darkens the underline and increases it's weight.
```
tkoleary, line 111: We can't say this because they would need to create an override to the button class which has text-transform: uppercase to even do it, plus we'd also want to remove the letterspacing. Can we just add a global helper class to Mayflower scss like:
```
.ma__case-title {
  text-transform: capitalize;
  letter-spacing: 0;
}
```

packages/core/stories/elements/Link/Link.stories.mdx
```diff
@@ -61,34 +63,37 @@ Links are used to embed actions or pathways to more information in a sentence or
   <Story story={stories.linkCalloutCSS} />
 </Canvas>
 
+---
 
-#### Link styled as a button
-**Usage:** Use when you are linking just a few words of text or when you are doing inline links. In general, we should style buttons as button and style links as links, to provide . [*For material honesty, we shouldn't encourage styling links as buttons, should use callout links instead, unless we can come up with appropriate examples*]
+### Link styled as a button
```
clairesunstudio:
```suggestion
### Link Styled as a Button
```

just nitpicking to be consistent as other headings

packages/core/stories/elements/Button/Button.stories.mdx
tkoleary: If we do that ^^ it would probably also make sense to do: 
```
.ma__button--quaternary {
-  text-transform: none;
+  text-transform: capitalize;
}

clairesunstudio: lets do this separately, as this will impact other implementations and create visual regressions for react. we should add a way to config casing consistently across all buttons, currently it's not consistent (This is currently in the to-do list --> https://docs.google.com/document/d/1HEep0C1ySxiNKM3r5oIMHjaAqJN4KXnPlg2NVImOw4g/edit#heading=h.5jchh3ntd4nt)

clairesunstudio (approved):
lgtm!

PR #1301, opened 2020-12-11
Core/storybook linting

clairesunstudio:
Add linting to Core and Core CircleCI build task 

To run lint:
- cd into packages/core
- `rushx lint` or `rushx lint-fix` 

Fixed all lint errors and added missing dependencies for core


clairesunstudio:
@smurrayatwork this is having issues with using the published version of mayflower-assets instead of the local package... 
In the core branch, button-copy.scss exist but not in the published package, how does rush determine which version to use what versions of mayflower-react and mayflower-assets should be specified in package.json?

```
SassError: Can't find stylesheet to import.
   ╷
15 │ @use '@massds/mayflower-assets/scss/01-atoms/button-copy';
```

packages/core/stories/tokens/colors/ColorDisplay.js
```diff
@@ -3,12 +3,19 @@ import PropTypes from 'prop-types';
 import ButtonCopy from '@massds/mayflower-react/dist/ButtonCopy';
 import './_color-display.scss';
```
smurrayatwork, line 4: Why is this import here?

smurrayatwork (approved):

PR #1313, opened 2020-12-17
Core/new header docs

clairesunstudio:
Render new header and header hamburger in core documentation

### DIFF
--- a/packages/core/stories/components/Header/Header.stories.js
+++ b/packages/core/stories/components/Header/Header.stories.js
@@ -1,12 +1,15 @@
 import React from 'react';
 import HeaderSlim from '@massds/mayflower-react/dist/HeaderSlim';
 import Header from '@massds/mayflower-react/dist/Header';
+import HeaderHamburger from '@massds/mayflower-react/dist/HeaderHamburger';
 import SiteLogo from '@massds/mayflower-react/dist/SiteLogo';
 import ButtonWithIcon from '@massds/mayflower-react/dist/ButtonWithIcon';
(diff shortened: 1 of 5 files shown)


packages/core/stories/components/Header/Header.stories.js
```diff
@@ -91,240 +94,16 @@ const headerUtilSearch = (
...
-                type: 'external'
-              }
-            ]
-          },
-          text: 'Log in to...'
-        }
-      ]
-    }}
+    mainItems={mainNavData}
+    utilityItems={[]}
```
smurrayatwork, line 98: Set UtilityNav to null here:
```suggestion
    UtilityNav={null}
```

packages/core/stories/components/Header/Header.stories.js
```diff
@@ -91,240 +94,16 @@ const headerUtilSearch = (
...
+  />
+);
+
+
+const headerHamburger = (
+  <HeaderHamburger
+    mainItems={mainNavData}
+    utilityItems={[]}
```
smurrayatwork, line 105:
```suggestion
    UtilityNav={null}
```

packages/core/stories/components/Header/Header.stories.js
clairesunstudio: won't setting it to null remove the utility nav items? I'd like to show the utility nav example same what's live on mass.gov

smurrayatwork: Oh! No you're right, disregard.

smurrayatwork (approved):

PR #1334, opened 2021-01-05
React/add brand banner (Part 1)

clairesunstudio:
Part 1 of adding a brand banner component:
Build and export a BrandBanner component in mayflower-react

brand-banner


packages/react/src/components/molecules/BrandBanner/index.js
```diff
@@ -0,0 +1,103 @@
...
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import Image from 'MayflowerReactMedia/Image';
+import ButtonWithIcon from 'MayflowerReactButtons/ButtonWithIcon';
+import IconChevron from 'MayflowerReactBase/Icon/IconChevron';
+import IconBuilding from 'MayflowerReactBase/Icon/IconBuilding';
+import IconLock from 'MayflowerReactBase/Icon/IconLock';
+import logo from '@massds/mayflower-assets/static/images/logo/stateseal.png';
```
smurrayatwork: You shouldn't import files here. Instead, require the user to pass in the src of the image they want to use. Importing this way forces the user to need webpack.

packages/react/src/components/molecules/BrandBanner/index.js
```diff
@@ -0,0 +1,103 @@
...
+  text = 'An official website of the Commonwealth of Massachusetts'
+}) => {
+  const lightTheme = bgTheme === 'light';
+  const brandBannerClasses = classNames('ma__brand-banner', {
+    [`ma__brand-banner--${bgColor}-bg-${bgTheme}`]: bgColor && bgTheme
+  });
+  const [expanded, setExpanded] = useState(false);
+
+  const handleOnToggle = () => {
+    setExpanded(!expanded);
```
smurrayatwork:
```suggestion
    setExpanded((prevExpanded) => !prevExpanded);
```

packages/react/src/components/molecules/BrandBanner/index.js
clairesunstudio: ah ok.. so there's no way to enforce passing in the image of the seal then right?

PR #1337, opened 2021-01-07
DP-20768 Get the page content container width for full width elements

ygannett:
…line width in imageFill.js

## Description

Some page headers contain `div.main-content` in the **pre-content** area in addition to the actual page content area.  These two containers have different widths, one in the pre-content is much smaller than actual page content width.  

**imageFill.js** to set an inline width to some elements with x-large option.  Since the script was specified to look for an element with **.main-content**, as it found the first matched element in this specific case, it got the element width and set it as the inline width value, which didn't fill the page content container width as expected.

banner

To solve this situation, in the JS, set the selector to look for `.main-content` as the first child of `#main-content` to make sure it finds one for the page content.  The one in the **pre-content** area is not the direct child of `.main-content`. 

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-20768)
- [Github issue]()

## Steps to Test

1. Build the Drupal site with this branch.
2. Create an info. details page with a page banner, add an x-large iframe in a section.
3. Find the iframe is displayed with the full width of the page content container.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


packages/patternlab/styleguide/source/assets/js/modules/imageFill.js
```diff
@@ -8,7 +8,7 @@ export default (function (window, document, $, undefined) {
 
   function mediaWidth() {
     // Define wrapper width for use.
-    var wrapperWidth = $(".main-content").width();
+    var wrapperWidth = $("#main-content > .main-content").width();
```
clairesunstudio:
```suggestion
    var wrapperWidth = $("#main-content > .main-content--two").width();
```

I feel like it's more accurate to target `.main-content--two`, as this class sets the container width. e.g. `main-content main-content--full` is not the right container width

clairesunstudio: or `$("#main-content > .main-content.main-content--two")`

ygannett:
The left/right padding are removed from `.ma_iframe_container`.

clairesunstudio (approved):

tom-fleming (approved):
LGTM!

PR #1344, opened 2021-01-21
Hotfix/dp 20986 fix circle ci

clairesunstudio:
This PR fixes circleCI failures
- lint fix
- remove failed caspio example in patternlab
- Change Caspio dataId in Patternlab to use the dedicated testing example to avoid breakage


tom-fleming (approved):
LGTM

PR #1345, opened 2021-01-22
Fix filterbox filter alignment and move into assets

clairesunstudio:
Screen Shot 2021-01-22 at 9 46 28 AM

## To Test in React Storybook
http://localhost:6006/?path=/story/organisms-filterbox--filter-box-example

Before:
Screen Shot 2021-01-22 at 9 43 17 AM

After:
Screen Shot 2021-01-22 at 9 46 14 AM


ygannett (commented):
The option component is aligned.  It looks good.  
In Safari and FF, the button component is pushed down to the next row even there is enough room to line up with the option component.  
safari-ff
In Chrome, it lines up.  Just mentioning this since it's noticed.

As long as this ticket is meant to fix the option component, it's ready to merge.

ygannett (approved):
The option component is aligned.  It looks good.  
In Safari and FF, the button component is pushed down to the next row even there is enough room to line up with the option component.  
safari-ff
In Chrome, it lines up.  Just mentioning this since it's noticed.

As long as this ticket is meant to fix the option component, it's ready to merge.

PR #1341, opened 2021-01-11
DP-19859 Relationship indicator IE display issue

ygannett:
Any PRs being created needs a changelog.txt file before being merged into dev. See: [Change Log Instructions](https://github.com/massgov/mayflower/blob/develop/docs/for-developers/changelog-instructions.md)

## Description

Remove the style causing the IE display issue at the show all state from the Relationship Indicator.

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-19859)
- [Github issue]()

## Steps to Test

1. Go to any page with a relationship indicator in IE.
2. Click the "show X more" in the primary. Note:  The issue only occurs with the primary unit.
3. Find the terms component is still lined up with the label.  It doesn't slide below the label.
4. Find the extra space between the label and the term unit is removed.

## Screenshots
Use something like [licecap](http://www.cockos.com/licecap/) to capture gifs to demonstrate behaviors.

## Additional Notes:

Anything else to add?

#### Impacted Areas in Application

*

#### @TODO

*

PR #1353, opened 2021-02-01
Release 11.1.3

clairesunstudio:
### Changed 
- (Patternlab) [Relationship Indicator] DP-19859: Fix the Relationship Indicator display at the show all state in IE. (#1341)

### Fixed 
- (Patternlab) [Figure] DP-21059: Make a skip link target to be displayed only when its associated skip link is clicked. (#1352)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Mayflower Release Notes
 All notable changes to this project will be documented in this file.
 
+## 11.1.3 (2/1/2021)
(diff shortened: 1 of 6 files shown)


clairesunstudio:
I don't see the code change from #1341 in this release

clairesunstudio:
I can see it in the branch but not in the code change in github. I think this is due to the timing of the merges. This could be avoided in the future by "squash and merge" PRs into develop branch.

PR #1382, opened 2021-03-25
[Patternlab] fix toc js on info-details page

clairesunstudio:
This is not really a Mayflower issue, as when you remove the toc object it doesn't render it in the DOM in Patternlab as expected. 

I think in openmass its still rendering the TOC in DOM when the sections are less than 2, we are just hiding it conditionally using CSS. Ideally I think we should not output the TOC data object. 
https://github.com/massgov/openmass/blob/d791d603365c5c073b68a40669f9703a4d5006ab/docroot/themes/custom/mass_theme/templates/content/node--info-details.html.twig#L109-L115
```
  {% include "@organisms/by-template/sticky-toc.twig" with {
    "stickyTOC": {
      "title": "Table of Contents"|t,
      "parent": "#main-content",
      "sections": ".ma__information-details__content h2"
    }
  } %}
```

---

### To test:

confirm that these links work in Drupal 
Screen Shot 2021-03-25 at 2 35 55 PM
without the overriding css added in the richtext overview field:

````
<style type="text/css">.sticky-toc-jump-target {
display: none;
}
</style>
````

due to the `:before` pseudo-class on `.sticky-toc-jump-target` blocking the overview even when there's no TOC rendered

Screen Shot 2021-03-25 at 2 34 22 PM

### CODE REVIEW

File: packages/patternlab/styleguide/source/assets/js/modules/stickyTOC.js [hunk 1]
```diff
@@ -1,6 +1,7 @@
 export default (function (window, document) {
 
   const tocs = Array.from(document.getElementsByClassName("ma__sticky-toc"));
+  console.log(tocs)
```
clairesunstudio:
```suggestion
```

ygannett (approved):

PR #1437, opened 2021-07-06
Revert "DP-22255: Allow authors to choose the type of ordered list style they need for a list"

clairesunstudio:
Reverts massgov/mayflower#1429
The list attribute used in the drupal module is deprecated https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li#attr-type. This change is too greedy and unpredictable. Instead, target the deprecated type attribute to set `list-style-type` property instead (in rich text only for the drupal module)

This will fix the unexpected styles for the list components on mass.gov and in mayflower, e.g.
- https://www.mass.gov/orgs/massachusetts-registry-of-motor-vehicles (org dropdown list)
Screen Shot 2021-07-08 at 1 07 25 PM

- https://www.mass.gov (header mixed menu sub list)
Screen Shot 2021-07-08 at 1 06 44 PM

The remaining issue is that css attribute selectors are case insensitive - this means that
`ol[type="a"] ` and `ol[type="A"]` are equivalent. So the latter selector will override the previous one. 
https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

### 2 options:
1. we do nothing in css, in Drupal, we remove the option to for either lower or upper latin and lower (a or A) or upper roman (i or I) options. 
2. we add a class on the list item when a list type is selected using the ckeditor list style module, if possible. Then we can keep all options and just target the class name for the different. See W3C example https://www.w3schools.com/css/css_list.asp

Added test content in rich text component
https://mayflower.digital.mass.gov/patternlab/b/revert-1429-patternlab/DP-22255_allow_authors_to_choose_list_style_types/?p=organisms-rich-text

Other test URLs:
https://mayflower.digital.mass.gov/patternlab/b/revert-1429-patternlab/DP-22255_allow_authors_to_choose_list_style_types/?p=pages-organization-boards
https://mayflower.digital.mass.gov/patternlab/b/revert-1429-patternlab/DP-22255_allow_authors_to_choose_list_style_types/?p=pages-Homepage


arthurbaghdas (approved):
LGTM

clairesunstudio:
Reverting the full feature for now. This PR is a reference for the re-implementation

PR #1444, opened 2021-07-14
DP-22395

ddiestra:
## Description
New templates for the alerts

## Related Issue / Ticket

- [JIRA issue](https://jira.mass.gov/browse/DP-22395)
- [Github issue](https://github.com/massgov/mayflower/pull/1444)


packages/patternlab/styleguide/source/_patterns/02-molecules/emergency-header.twig
```diff
@@ -1,8 +1,12 @@
 {% set alertPrefix = emergencyHeader.prefix ?: 'Emergency Alerts' %}
-<h2 class="ma__emergency-header">
+<div class="ma__emergency-header ma__emergency-header--pt">
```
clairesunstudio, line 2: what does `pt` stand for? and why do we need to add a class?

packages/patternlab/styleguide/source/assets/scss/index.scss
```diff
@@ -161,6 +161,7 @@
 @use "03-organisms/header-hamburger";
 @use "03-organisms/header-mixed";
 @use "03-organisms/header-alert";
```
clairesunstudio, line 163: Since we are fulling replacing header-alert with the new design and functionality, I think we should just do a full replacement instead of keeping both

clairesunstudio:
@ddiestra, I looked at both backstop reports and approved the react visual regressions. For the Patternlab report: https://39515-137521197-gh.circle-artifacts.com/0/backstop/reports/html/index.html:
The spacing in the page header on org pages and service page should match what's currently in production. 
Also, for the page wide alerts in the page templates, can you move them below the pager header, instead of the above the page header? see design https://www.figma.com/proto/4JpC94wWhQo9siPiucvV92/Mass.Gov?node-id=383%3A12246&scaling=min-zoom

clairesunstudio:
Code-wise, I think the intention is to replace the existing EmergencyHeader and the HeaderAlert with the new design and functionality. I don't think we need to keep the old style in the system. Hence I was asking earlier about styling the `ma__emergency-header` directly rather than adding a new class `ma__emergency-header--pt`, as well as modifying the HeaderAlert component instead of creating a new HeaderAlerts component. 

The changes to EmergencyHeader might impact Mayflower React (is that the reason why you didn't touch the original class?) I think we should try to keep React and Pattenlab in sync, but if you don't want to address this PR, we should create a ticket to address that. As for the HeaderAlert, this component doesn't exist in React, so changes in Patternlab and assets won't impact Mayflower React

Note: I think we could rename the HeaderAlert to something else -- e.g. Alerts or AlertsExpandable, as the name Header Alerts doesn't make too much sense

clairesunstudio:
One more design/frontend detail, can you add back the hover style to the emergency alerts links? The link in the alert used to have a hover style:
after (no hover style):
Screen Shot 2021-07-16 at 4 41 22 PM
before (on hover border ):
Screen Shot 2021-07-16 at 4 41 27 PM

clairesunstudio:
Please see the latest comments on ticket https://jira.mass.gov/browse/DP-22395

One other minor suggestion:
Can you update the mock data for the statewide alert better show the design intention? https://mayflower.digital.mass.gov/patternlab/b/DP-22395/?p=pages-guide
Current:
Hide 3 The State is experiencing severe weather due to the winter storm Paula.
Suggestion:
Hide 2 travel-related updates and 1 update for the state employees

packages/patternlab/styleguide/source/_patterns/02-molecules/emergency-header.twig
ddiestra: PatternLab

packages/assets/scss/02-molecules/_emergency-header.scss
```diff
@@ -68,7 +68,9 @@
   &__toggle {
     cursor: pointer;
     color: $c-black;
-    font-weight: $fonts-light; 
+    font-weight: $fonts-light;
```
clairesunstudio, line 71: fyi, we have a mixin for this
```
@mixin ma-button-reset {
  background-color: transparent;
  border: none;
  padding: 0;
}

```

clairesunstudio (approved):
approved, waiting on backstop tests

clairesunstudio:
ok Approved the visual regressions https://39628-137521197-gh.circle-artifacts.com/0/backstop/reports/html/index.html

PR #1483, opened 2021-08-27
DP-22857 - Fixing Sticky header

ddiestra:
## Description
Change pointer events on sticky header to avoid collide with other content.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-22857)
- [Github issue]()


clairesunstudio (approved):
tested in mass.local. Looks good!

PR #1490, opened 2021-09-02
DP-22210 fix focus issue

ddiestra:

## Description

Fix navigation issues when using keyboard.

## Related Issue / Ticket

- [JIRA issue](DP-22210)
- [Github issue]()


changelogs/DP-22210.yml
```diff
@@ -0,0 +1,21 @@
...
+# - Removed for now removed features.
+# - Fixed for any bug fixes.
+# - Security in case of vulnerabilities.
+# e.g. Added
+
+#___EXAMPLE WITH MULTIPLE ITEMS___
+Changed:
+  - project: Patternlab
+    component: Leaflet
+    description: Fixing focus issues when navigating with keyboard.
```
clairesunstudio:
```suggestion
    description: Fixing focus issues when navigating with keyboard. (#1490)
```
please reference the PR # in the changelog

packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig
```diff
@@ -9,7 +9,9 @@
   {% endif %}
   <div class="ma__location-listing__columns">
     <div class="ma__location-listing__container">
-      <div class="ma__location-listing__map">
+      <div class="ma__location-listing__map" aria-label="{{leafletMap.title}}">
+        <a href="#location-listing-results" aria-describedby="sr-info">skip to the location listing</a>
```
clairesunstudio, line 14:
```suggestion
        <a href="#location-listing-results" aria-describedby="sr-info">skip to the location listing</a>
```
This should be a real id that's passable and the ID should be placed in the html of the listing div.
I can see this link, be nothing is happening when it's clicked or `selected.```

packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig
```diff
@@ -9,7 +9,9 @@
   {% endif %}
   <div class="ma__location-listing__columns">
     <div class="ma__location-listing__container">
-      <div class="ma__location-listing__map">
+      <div class="ma__location-listing__map" aria-label="{{leafletMap.title}}">
```
clairesunstudio, line 13: not seeing this label being rendered https://mayflower.digital.mass.gov/patternlab/b/feature/DP-22210_fix-focus-issue/?p=pages-location-listing

clairesunstudio (changes requested):
> Change navigation keys for markers from arrow keys to TAB key.  TAB for forward.  Shift + TAB for backward.  All focusable elements in the map are navigated with TAB and Shift + TAB.
The markers are not tabbable, only the first one is. 

Also these changes seem to break the non-keyboard interactions.... The map can't be enabled to move anymore, and also clicking on a marker doesn't open the tooltip anymore.

packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js
```diff
@@ -125,7 +129,7 @@ export default (function (window,document, $) {
         }, 200);
       });
 
-      mymarker.on('popupclose', function() {
+      mymarker.on('popupclose', function(e) {
```
clairesunstudio: don't think `e` is used

packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js
```diff
@@ -99,6 +98,7 @@ export default (function (window,document, $) {
     window.leafletMap = mymap;
 
     window.stopFocus = false;
+    window.stopOffFocus = false;
```
clairesunstudio: what's stopFocus and stopOffFocus mean, can you add a comment to explain these custom variables?

clairesunstudio:
Functionality review:
- Confirmed that the non-keyboard interactions are restored (can click into the map to the pane and can click on the markers to view the tooltip)
- The skip link is working - it skips the map and the next tabbable item is the image in the first listing teaser
- The map controls are focusable, but not in the right order (this is fine as we are limited by the leaflet library. I don't think this is a deal breaker)
----

Outstanding issue:
When a marker is in focus, the link in the popup gets auto-focused, is this the expected behavior? This creates a few issues that I think are not intuitive for the users:
1. You can't navigate between markers when the popup is open
2. There's no way to toggle close the marker (since the focus is changed to the popup) until you tab to the close button. 
3. TAB and SHIFT+TAB are not consistent and reciprocal

>I think maybe the marker popup shouldn't auto open and shift the focus. The tabbing order should remain among the markers, once the user press ENTER or SPACE keys the popup opens. otherwise, TAB and SHIFT+TAB navigates between the markers

Here is a flow I propose we can follow (@ygannett please review and suggest otherwise)
- TAB to focus on the next marker
- SHIFT + TAB to focus on the previous marker
- ENTER/SPACE to open a marker popup
- when a marker is focused and open, TAB brings focus into the popup window
- when the user TAB through all the focusable items in the popup window (the last item is the close button) -- this creates a focus order dilemma:
  1. should the focus go to the currently opened marker? (focus trapping)
  2. should the focus go to the next marker (should we allow the focus to go to a different marker when the popup is open?) 

If we allow the focus to go to the next marker, then the previous marker popup should close when:
 option 1. the focus is on another marker
 option 2. another marker popup is open (after click or ENTER/SPACE is pressed)

ygannett:
@ddiestra Thank you so much for the all work you've done.

@clairesunstudio Thank you for the thoughtful review.

### Marker & Balloon interaction

For sighted users, I think they want to see a balloon when a marker gets focus. So, they can confirm what they are looking at or determine that's the information they are looking for or not.  And, this creates a tricky situation to us.

So, how about this?
1. When a marker gets focus, its balloon is displayed, but its content is not focusable.  Probably managing that with `tabindex`. All focusable elements in the balloon has `tabindex="-1"`.
2. When hitting `TAB` while a marker has focus, focus moves to another marker.  
     At the same time;
      - the first marker's balloon goes away,  
      - the second marker gets focus,
      - the second one's balloon shows up, but its content is not accessible.  
    
    This way, users can go through all markers checking their addresses.  This says that only one balloon is visible at a time, one whose marker has focus.
3. When users want to interact with the balloon content, they hit `ENTER`.  
     This triggers to change;
      - move focus from the marker to its balloon(focus outline disappears from the marker),
      - change `tabindex` values in the focusable elements in the balloon to `0`. Users can tab through the content.
4. When the balloon's close button is hit,
      - the balloon disappears,
      - set `tabindex` values to `-1` for all focusable elements in the balloon,
      - set focus on its marker (this is important, so a user knows where (s)he is in the map.) The rule of thumb with popup, accordion and dropdown is to **come back to where you left at** _when they close and lose focus_.  In the main nav, when a sub menu unit closes, focus goes back to its parent nav item.  When the content container of the "log in to" in the utility nav closes, focus goes back to the "log in to" button. This way, users don't get lost.  They are the ones to take action to move to next/previous element.
   
   Alternatively, a balloon could be closed with `ESC` key when a balloon or its content has focus, but I suggest to leave this for enhancements since things to set up are complicated enough already.

With this option, we need to provide how-to info. to keyboard and screen reader users.

### SHIFT + TAB

This might solve the `SHIFT + TAB` issue as well since balloon don't get focus unless a user takes action to do so.  Besides, when hitting `SHIFT + TAB` twice consecutively, Taunton District Court's balloon opens. Something to watch for at the next testing.

### Tab order with the map controls
I agree with this.
> - The map controls are focusable, but not in the right order (this is fine as we are limited by the leaflet library. I don't think this is a deal breaker)

### Cut-off map 

It seems the map gets cut off on the right side when using the zoom out feature in a browser.  This is 75% in Chrome.
cutoff 2

### Overwrap Feedback button

The Feedback button is on the map and it could cover a marker.
feedbackbutton

### Balloon phone + fax numbers 

In the balloon, when phone and fax have their data, sometimes it's not added as text content in `<a>`.  They have `href` values, and appear to be visually empty links.  The link is not visible with the absence of its text content, but still gets focus, but no visible focus indicator because of that.
location-list-map-marker-phone

fax-phone

Note:  This issue might be related to below issue mentioned in the ticket description: 6.c.ii
> Not sure why some of the titles in the popup window are not links, e.g. “Brookline District Court”. 

### Sporadically items in the listing and markers in the map are not synced.

I'm not sure what triggers this issue, but as tabbing through the list, a focused location in the listing doesn't match the focused marker in the map.
 
location-list-map

### ESC to leave the map

This is not doable?
6.d.Allow users to leave the map as they wish.

### Skip link aria-describedby content 

I am so sorry.  I made a mistake.  `div#isr-info` is supposed to have `display: none;`.  With `.ma__visually-hidden`, the element gets announced, but that's not necessary as it's announced at the skip link.  

Please remove `.ma__visually-hidden` and set `display: none;` to `div#isr-info`.

clairesunstudio:
TL;DR:
@ddiestra after touching base with @ygannett again on our suggestions, let me summarize the decisions for you:
1. Please refer to this diagram https://www.figma.com/file/Gq95WKk6R4BB8lIA7WkXoY/Leaflet-map-marker-keyboard-interaction-diagram?node-id=0%3A1 for the keyboard interactions between markers and popups within the map. (As I mentioned in my comment, the keyboard interaction is still problematic on the markers)
2. When there's no popup open in the map, please add ESC keydown to skip the map (same functionality with the skip link onclick) - This provides keyboard users a way to jump out of the map when they are in between markers. 
3. There is some weird behavior in the popup content -- the title link is missing anchors or the phone or tax numbers are not visible but are focusable. Please make sure the title links work properly and if there are no phone or tax numbers, do not render the fields. 
Screen Shot 2021-10-04 at 12 12 47 PM
4. Please refer to yukiko's comment on "Skip link aria-describedby content" above

packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js
```diff
@@ -113,26 +124,54 @@ export default (function (window,document, $) {
...
 
-      $(mymarker._icon).on('click', function(e){
-        e.stopPropagation();
-        e.preventDefault();
-      });
+      $(mymarker._icon).on('keydown', function(e){
 
-      $(mymarker._icon).on('focus', function(e){
-        if (window.stopFocus) {
+        if (e.keyCode == 16 || e.keyCode == 9) {
```
clairesunstudio: Please use .code and only use .keyCode as a fallback instead, as it's deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

clairesunstudio:
External and Internal reivew:
- [ ] For all the keyboard event.keyCode, add event.code and use .keyCode only as a fallback instead, as it's deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode
- [x] SHIFT+TAB is not closing the previous popup. Please make sure this flow is working properly
>Screen Shot 2021-10-06 at 2 09 14 PM
- [x] ESC should close an open popup if there's any. Please make sure this flow is working properly
>Screen Shot 2021-10-06 at 2 13 51 PM
 - [x] For the phone number checks, see mapGoogle.js for the missing format function. When a phone and fax number is passed unformatted, they should be formatted and render in the popup
 ```
  function infoTransform(data) {
    let infoData = {
      phoneFormatted: formatPhone(data.phone),
      faxFormatted: formatPhone(data.fax)
    };
    return Object.assign({},data,infoData);
  }
```

**Comment by clairesunstudio** [comment]
One more thing that we didn't talk about on the call earlier, is it possible to add ESC map to the mapped location component --> https://mayflower.digital.mass.gov/patternlab/b/feature/DP-22210_fix-focus-issue/?p=organisms-mapped-locations?
These maps can have even more markers on them on mass.gov and are not skippable. 

Update: will create a follow up ticket for the mapped location component

File: packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js (hunk 4 above)
Comment (clairesunstudio) [reply packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js]: @ddiestra

File: packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig (hunk 2 above)
Comment (clairesunstudio) [reply packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig:14]: In order for this feature to be used on other maps other than location listing, the ID should be a passable value. This can be in a followup

File: packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig (hunk 3 above)
Comment (clairesunstudio) [reply packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig:13]: @ddiestra  please fix the aria-label

File: packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js (hunk 5 above)
Comment (clairesunstudio) [reply packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js]: @ddiestra

File: packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js [hunk 7]
```diff
@@ -108,15 +122,78 @@ export default (function (window,document) {
...
-        // Fire the map focus callback on popup close
-        mymarker.on('popupclose', function() {
-          mymap.fire('focus');
-        });
-      }
+      .bindPopup(compiledTemplate(infoData));
+
+      $(mymarker._icon).on('keydown', function(e){
+
+        if (e.code == 'Tab') {
```
clairesunstudio:
```suggestion
        if (e.key == 'Tab' ) {

```
IE doesn't support e.code. I think we can use .key
or add a fallback like (e.code == 'Tab' || e.keyCode == 9 )?

packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js
ddiestra: Those are not longer used, both variables have been removed

packages/patternlab/styleguide/source/assets/js/modules/mapLeaflet.js
ddiestra: This is not longer used either

clairesunstudio:
1. Space key wasn't working in Chrome, fixed it.
2. `String.fromCharCode()` you added always returns empty, so I removed it.

clairesunstudio (approved):

PR #1491, opened 2021-09-03
DP-22657 Migrate org page "What would you like to do" to sections

Javi-er:

## Description
It switches the collapsible content icon and adjust some padding.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-22657)

## Steps to Test

1. 

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

### DIFF
--- a/packages/patternlab/styleguide/source/assets/js/modules/accordions.js
+++ b/packages/patternlab/styleguide/source/assets/js/modules/accordions.js
@@ -1,10 +1,28 @@
+// noinspection JSVoidFunctionReturnValueUsed
+
 import checkActive from "../helpers/cssControlCode.js";
+import {forEach} from "../../../../../../core/storybook-static/vendors~main.047e28ac683eee78436e.bundle";
 
 export default (function (window,document,$,undefined) {
+  let $jsAccordion = $('.js-accordion');
 
-  $('.js-accordion').each(function(index){
+  $jsAccordion.each(function(index){
     init.apply(this, [index]);
   });
+
+  // In case that all paragraphs are expanded by default, change the "expand all" link to "collapse all".
+  let allExpanded = true;
+
+  $('.ma__collapsible-content--extended .js-accordion').each(function(index){
+    if(! $(this).hasClass('is-open')) {
+      allExpanded = false;
+    }
+  });
+
+  if(allExpanded) {
+    toggleAll();
+  }
+
 
   $(document).on('ma:AjaxPattern:Render', function(e,data){
     let $context = data.el;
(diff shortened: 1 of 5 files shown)


packages/patternlab/styleguide/source/assets/js/modules/accordions.js
```diff
@@ -1,11 +1,28 @@
...
     init.apply(this, [index]);
   });
 
+  // In case that all paragraphs are expanded by default, change the "expand all" link to "collapse all".
+  if($(".ma__collapsible-content--extended").length) {
+    let allExpanded = true;
+
+    $('.ma__collapsible-content--extended .js-accordion').each(function(index){
+      if(! $(this).hasClass('is-open')) {
+        allExpanded = false;
```
clairesunstudio, line 17: I think you might also want to add this to the individual accordion toggle function. As, if the user manually expand / collapse all the accordions, the state is wrong. e.g.
Screen Shot 2021-09-14 at 12 29 39 PM

clairesunstudio (commented):
please add an aria-controls to the toggle button and an id to the section for accessibility
https://github.com/massgov/mayflower/pull/1491/files#diff-55da25739845941aa85243d1687b7572c18a4b0080047cb8373801a860d3f666R11

clairesunstudio:
Another nitpicking comment:
Can you align the counters to the top? 
before:
Screen Shot 2021-09-14 at 12 56 28 PM
after:
Screen Shot 2021-09-14 at 12 57 11 PM

Javi-er:
Hi @clairesunstudio , I did  a follow up PR with these changes here: https://github.com/massgov/mayflower/pull/1502

I refactored the accordion logic to better handle collapse all / expand all link text and the mobile aligments.

Regarding the aria labels, those are actually there since before I worked on this, take a look to the screenshot below. Let me know if you think that these should be moved to different elements instead. The only element that doesn't have aria labels is the collapse / expand all link, this is because it doesn't control a single element  but all the elements, `aria-controls` should point to a single ID element.

j-r-capture 2021-09-16 a la(s) 20 34 01

clairesunstudio:
@Javi-er, ok thanks I will take a look at the new PR. 
RE: aria-controls, right, I'm asking you to add an id on the whole expand/collapse all section, and add the aria-controls on the expand/collapse all button to indicate what the button is controlling.

Javi-er:
> @Javi-er, ok thanks I will take a look at the new PR. RE: aria-controls, right, I'm asking you to add an id on the whole expand/collapse all section, and add the aria-controls on the expand/collapse all button to indicate what the button is controlling.

We should probably chat a bit about this, there isn't a single section controlled by this button, but instead each individual accordions that already have aria-label, the container is ma_collapsible-content__items but this element isn't modified at all by the button.

PR #1499, opened 2021-09-14
DP-22028 - Leaflet zoom level

ddiestra:
## Description

Simple update to the zoom level so the whole state can be seen on mobile.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-22028)
- [Github issue](https://github.com/massgov/mayflower/pull/1499)


changelogs/DP-22028.yml
```diff
@@ -0,0 +1,6 @@
+Changed:
+  - project: Patternlab
+    component: Leaftlet Map
```
clairesunstudio:
```suggestion
    component: LeaftletMap
```
Please use PascalCase

changelogs/DP-22028.yml
```diff
@@ -0,0 +1,6 @@
...
+    description: Update min zoom so tha map can zoom out on mobile.
```
clairesunstudio:
```suggestion
    description: Update min zoom so the map can zoom out on mobile. (#1499) 
```
Please

clairesunstudio:
huh ok, so the tiles are now working with the zoom level 7 now. The GIS team might have deployed the requested changes. Nice!
The min zoom is tested locally in openmass. Looks good!
Before:
https://www.mass.gov/rmv-services-at-aaa
Screen Shot 2021-09-16 at 5 51 19 PM
After:
Screen Shot 2021-09-16 at 5 51 07 PM

https://www.mass.local/rmv-services-at-aaa

The other part of this ticket is about the max boundary.
>Also, a related issue to tiles is that, even at the existing min zoom level, we can't set a max bounds:
// maxBounds is disabled due to unexpected shifting when popups are hitting the boundaries.
// This can be turned back on when the tiles are large enough so that the min zoom level shows the whole state and we can find bounds at a min zoom level that provide enough tiles padded for popups.

If you start panning the map, the map doesn't stop at the edge of the tiles, this is problematic because the request to the tile images is returning 404s and is causing problems with the moving of the map as they are technically not part of the map. 
Screen Shot 2021-09-16 at 5 57 02 PM

map-bounds

clairesunstudio:
This is the new **TEST** tiles the GIS team has provided for this purpose --> https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/MassGIS_Basemap_MassGovTest/MapServer
it should cover more area than the current one --> 
https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/MassGISBasemap/MapServer

TO DO:

- [x] need to recalculate the corners and the bounds for the new tiles
```
    const corner1 = L.latLng(43.12916191721289, -67.40279674530031); //northEast ==> need to find the new corners
    const corner2 = L.latLng(41.09188542307055, -76.28524303436281); //southWest ==> need to find the new corners
    const maxBounds = L.latLngBounds(corner1, corner2);
```
- [ ] Once the new corners are found and tested with the maxbounds, we will need to contact GIS team to increase the map zoom level to `Max LOD: 19` and release the new tiles

clairesunstudio:
@ddiestra how did you calculate the corners? I think we probably want to cut back the corners of the at the edge of the tiles a little bit, since the tile is not a perfect rectangle. The goal is to prevent the map moving out of bounds, like this:
https://mayflower.digital.mass.gov/patternlab/b/feature/DP-22028_leaflet-zoom-level/?p=organisms-mapped-locations
Screen Shot 2021-09-30 at 10 56 32 AM
Screen Shot 2021-09-30 at 11 08 12 AM

clairesunstudio:
It'd be great in the comment you can add some notes about how those bounds are determined. Thanks!

clairesunstudio:
For this branch, please add and comment out the snippet you used to get the bounds. 
Knowing that this needs some back and forth with the GIS team. Please just document the existing issue and the followup needed in the ticket and pass it back to me. 


---

Need to get in touch with Michael for the test tiles:
New (test):
- https://massgis.maps.arcgis.com/apps/mapviewer/index.html?layers=962ba688429c4564ae7227f9872468a2
- https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/MassGIS_Basemap_MassGovTest/MapServer

Old (prod):
- https://www.arcgis.com/home/webmap/viewer.html?basemapUrl=https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/MassGISBasemap/MapServer?cacheKey=b0a775ba3795b500
- https://tiles.arcgis.com/tiles/hGdibHYSPO59RG1h/arcgis/rest/services/MassGISBasemap/MapServer?f=html&cacheKey=b0a775ba3795b500

PR #1516, opened 2021-10-05
DP-23034 a11y brand banner

ygannett:

## Description

Initially modified errors from the validator in the markup as screen readers announce incorrect item counts keeping the definition list as it’s semantically appropriate for this case. But, matching the layout with valid markup requires more change in css than necessary.  

Since screen readers announce `dt` and `dd` as a list item, not as a definition title and its definition in their intended association, the definition list has been switched to an unordered list with the original styles for the layout.  With the new version, the original `dt` and `dd` are now under the same `li` and their association is clearer.

The react template also has been switched with an unordered list.
  

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-23034)
- [Github issue]()

## Steps to Test

1. Compare the visuals in both patternlab and react with their prod versions for any differences.

    - [Prod react](https://mayflower.digital.mass.gov/react/index.html?path=/story/molecules-brandbanner--brand-banner-example)
    - [Prod patternlab](https://mayflower.digital.mass.gov/patternlab/?p=molecules-brand-banner)

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


packages/assets/scss/02-molecules/_brand-banner.scss
```diff
@@ -100,6 +104,20 @@
...
             }
         }
+
+        .dark {
+            fill: #fff;
+        }
+
+        // Icon color
+        
+        #official > svg {
```
clairesunstudio:
```suggestion
        #official > svg {
```
These overrides are breaking the dark theme

This Branch:
Screen Shot 2021-10-15 at 9 29 30 AM
Screen Shot 2021-10-15 at 9 30 39 AM

Prod:
Screen Shot 2021-10-15 at 9 31 20 AM
Screen Shot 2021-10-15 at 9 32 03 AM


Please note, the compiled CSS is also used as a CDN for the brand banner on mayflower core documentation https://mayflower.digital.mass.gov/core/index.html?path=/docs/components-brandbanner--brand-banner-example. Just want to make sure that the CSS makes sense with the HTML semantic in "show code"

I'm seeing some alignment issues in core:
Screen Shot 2021-10-15 at 9 43 00 AM

clairesunstudio: It'd be great to remove those css overrides with the IDs, and let the mixin handle the styling. What happens when you remove those styles?

clairesunstudio: update: the layout look fine in mayflower react storybook but not mayflower core storybook. Not sure why the discrepancy.

ygannett: I see core has the prod markup, DL, but the react in this PR has UL.  Does the core get pulled from outside this branch?

ygannett: I replaced the IDs with classes to each li to set icon color in the template.
In scss, the colors for the icons are set accordingly with the new classes.  Also, the color codes are replaced.

ygannett: Core looks aligned.

screenshot-localhost_6006-2021 10 15-14_03_41

clairesunstudio (approved):

clairesunstudio:
false positive https://43001-137521197-gh.circle-artifacts.com/0/backstop/reports/html/index.html

PR #1519, opened 2021-10-07
DP-22982 Iframe, Caspio and Tableau components on org page not rendering correctly

Javi-er:

## Description
Final adjustments for fixing Tableau large size and improved image stacking on medium size.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-22982)

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

### DIFF
--- a/packages/assets/scss/01-atoms/_figure.scss
+++ b/packages/assets/scss/01-atoms/_figure.scss
@@ -262,6 +262,7 @@
 
   // The block above makes large images to be 100% width which is the same as extra large.
   // This block applies when .no-sidebar is applied to the parent of the component page-content.
+  // This specific use case is for stacked row section component on organization pages.
   .page-content .no-sidebar &--large {
 
     @media ($bp-large-max) {
--- a/packages/assets/scss/01-atoms/_figure.scss
+++ b/packages/assets/scss/01-atoms/_figure.scss
@@ -72,7 +72,7 @@
 
     // support pre-existing figure elements besides dataviz.
 
-    &--left:not(.ma__dataviz),
+    &--left:not(.ma__dataviz):not(.ma__iframe__container),
     &--align-left,
     &.align-left {
       float: left;
@@ -80,7 +80,7 @@
       width: 50%;
     }
 
-    &--right:not(.ma__dataviz),
+    &--right:not(.ma__dataviz):not(.ma__iframe__container),
     &--align-right,
     &.align-right {
       float: right;
@@ -164,7 +164,10 @@
 
     .page-content:not(.no-sidebar) &--x-small,
     .page-content:not(.no-sidebar) &--small {
-      width: calc(50% - 15px);
+
+      &:not(.ma__dataviz):not(.ma__iframe__container) {
+        width: calc(50% - 15px);
+      }
     }
   }
 
@@ -206,7 +209,7 @@
     }
 
     @media ($bp-small-min) and ($bp-large-max) {
-      width: calc(50% - 15px);
+        width: calc(50% - 15px);
     }
 
     @media ($bp-small-max) {
@@ -218,7 +221,9 @@
 
     @media ($bp-large-extended-min) {
 
-      width: calc(50% - 15px);
+      &:not(.ma__dataviz):not(.ma__iframe__container) {
+        width: calc(50% - 15px);
+      }
     }
   }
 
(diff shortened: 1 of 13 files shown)


packages/assets/scss/01-atoms/_figure.scss
```diff
@@ -77,15 +77,15 @@
     &.align-left {
       float: left;
       margin-left: 0;
-      width: 50%;
+      width: 47%;
```
clairesunstudio: why are you adjusting the width? This is going to create some visual regressions https://42639-137521197-gh.circle-artifacts.com/0/backstop/reports/html/index.html

packages/assets/scss/01-atoms/_figure.scss
```diff
@@ -262,6 +261,21 @@
     float: none;
   }
 
+  .page-content .no-sidebar &--large {
```
clairesunstudio: Where are you adding the `.no-sidebar` class again? the parent class in the selector`.page-content` is the one inside of the stacked row section?

Javi-er: @clairesunstudio it wasn't necessary to move it, it's still being added to .ma__stacked-row__section

packages/assets/scss/01-atoms/_figure.scss
Javi-er: When there are two medium sized images floated together, it doesn't fit in one row since both of them are 50% of width plus some margin. But I see what you mean, I guess I can add this adjustment on OpenMass just for image paragraphs so nothing else is affected. I'll move it there.

Javi-er: I reverted this line and added an override  to OpenMass instead to address specifically this case, see: https://github.com/massgov/openmass/pull/1059

packages/assets/scss/01-atoms/_figure.scss
clairesunstudio: Is it necessary to have `.page-content` in the selector? I'm saying this because  `.page-content` seems to come from the page level template right? The point for componentizing styles is so that the styles are better scoped at the component level, chaining page-level/high level class selectors is anti-pattern and should be avoided. Can the selectors be simplified as `.no-sidebar &--large {}`?

packages/assets/scss/01-atoms/_figure.scss
clairesunstudio: hmm got it, ok

packages/assets/scss/01-atoms/_figure.scss
Javi-er: .no-sidebar &--large {} seems to be maybe too general to me? in any case .page-content is used everywhere, in this case .page-content is inside .ma__stacked-row__section (see stacked-row-section.twig) and many components have it. But I can make that change if you feel that's safe.

clairesunstudio: I thought `.no-sidebar` is applied at the stacked row section wrapper, no? If the `.page-content` class is inside the stacked row section, shouldn't it be ` .no-sidebar .page-content &--large`?  `.page-content` is very ambiguous and used everywhere, if we can avoid it we should

Javi-er: right, in this case .page-content is also a top level container in the page too so it's using that.

Javi-er: `.page-content  &--large` is the selector affecting the image, I just added the `.no-sidebar` in between.

Javi-er: `.page-content &--large` is the existing selector affecting this image, I just added the `.no-sidebar` in between.

packages/assets/scss/01-atoms/_figure.scss
```diff
@@ -262,6 +260,23 @@
     float: none;
   }
 
+  // The block above makes large images to be 100% width which is the same as extra large.
+  // This block applies when .no-sidebar is applied to the parent of the component page-content.
```
clairesunstudio:
```suggestion
  // This block overrides the override above when the figure component is wrapped inside of a stacked row section with a  .no-sidebar class.
```

clairesunstudio (approved):

PR #1571, opened 2021-12-13
DP-23620 halfImage not being set as string, it may be interpreted as …

beto-aveiga:
…true just because it is not empty

## Description
- When a key message is rendered, there is a JS that generates a margin at the bottom, depending on certain logic. There was an IF that was checking for a condition to modify this margin, but the value that received was a string, hence interpreting any string as a non-false value. 
- I fixed the way that value is assigned to avoid these margin issues on key messages.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-23620)

### Before
image

### After
image

## Additional Notes:

Anything else to add?

*

#### @TODO

*

#### Today I learned...
- To double-check Twig assignments and falsy values on conditions 👨🏼‍🏫  
- Twig is not my friend 😛


changelogs/DP-23620.yml
```diff
@@ -0,0 +1,59 @@
...
+#    issue: Add a Jira ticket or issue number, e.g. DP-12345 or 124
+#    impact: Enter a valid impact, e.g. Minor
+
+#___EXAMPLE___
+Fixed:
+  - project: Patternlab
+    component: Key message
+    description: Fix key message margin bottom. (#1571)
+    issue: DP-23620
+    impact: Minor
```
clairesunstudio:
```suggestion
    impact: Patch
```

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/key-message.twig
```diff
@@ -37,7 +37,7 @@
 <script type="text/javascript">
   var id = "{{ id }}";
   var blckid = "blck{{ id }}";
-  var halfImage = "{{ keyMessage.halfImage }}";
+  var halfImage = {{ keyMessage.halfImage ?: "false" }};
```
clairesunstudio: question, why is it necessary to use the ternary here? Shouldn't it just be the value itself?
```suggestion
  var halfImage = {{ keyMessage.halfImage }};
```

beto-aveiga: @clairesunstudio to do not have the necessity of sending a value for halfImage property.
Other templates related to KeyMessage do not send this value, but the variant I did does send that value.
I thought that validating if that property hasn't a value then doing this will avoid templates displaying the wrong way.
However, I can update other templates as well and put there the value only.

clairesunstudio (approved):

PR #1579, opened 2021-12-27
Patternlab/dp 23425 location listing autocomplete

clairesunstudio:
https://massgov.atlassian.net/browse/DP-23425
OpenMass PR https://github.com/massgov/openmass/pull/1277

location-filter-validation

https://github.com/massgov/mayflower/pull/1580 should be merged in before this one. Merge conflicts need to be resolved.


clairesunstudio:
Attempt to fake the click event on the first match
```
        google.maps.event.addDomListener(locationInput, 'keydown', function(e) { 
            if (e.key == 'Enter') {
                //only submits when the autocomplete dropdown is closed
                if ($('.pac-container:visible').length) {
                  e.preventDefault(); 
                  const match = $('.pac-item').first();
                  match.on( "click", function(e) {
                    console.log(e.target)
                  });
                  
                  match.trigger("click");
                  console.log(match)
                }
            }
        }); 

```
or plain JS 

```
                  const match = document.querySelector('.pac-item');
                  match.onclick = (e) => {
                    console.log('clicked on the first match!')
                    console.log(e.target)
                  }
                  match.click();
                  console.log(match)
```

The click is simulated on the first item, but it's not triggering the google API request using the Google places autocomplete widget https://developers.google.com/maps/documentation/javascript/places-autocomplete#add-autocomplete.

clairesunstudio:
Since it’s not possible to implement Part 2 with the autocomplete widget,
> 2. To further improve the user experience, we should add a feature that will allow the only match to be submitted if Google returns one and only one suggestion. This means that when a user types in a place and hits enter or clicks the submit button, we will automatically pass that suggested place in the request, without requiring them to select it.

the next best thing to do is to persist the autocomplete dropdown when user hit “Enter” without selecting a valid address. it’s confusing when the user hits “enter” and see the error message telling them to select an option while the suggestions are not displayed (currently it only shows the list of suggestions when you re-focus on the input). 

This is the best hack I can do to achieve that - pressing enter again will not submit the term and will persist the dropdown on the front end:
location-filter-enter-2

clairesunstudio:
Updated error message
Screen Shot 2022-03-25 at 12 33 02 PM

`Please select an address in the suggesions. Hit ENTER or start typing to show suggesions.`

tom-fleming (approved):
LGTM

tom-fleming:
Discussed the Patternlab differences with Minghua, which seems to be isolated to the search button. It sounds like this issue should be resolved with this PR (https://github.com/massgov/mayflower/pull/1604/files), so will merge this one in.

PR #1590, opened 2022-02-14
Patternlab/dp 23982 breadcrumb accessibility fix

clairesunstudio:
Accessibility fixes to the new breadcrumb component https://massgov.atlassian.net/browse/DP-23982

### DIFF
--- a/packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/breadcrumbs.twig
+++ b/packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/breadcrumbs.twig
@@ -25,7 +25,7 @@
           aria-label="Show all Breadcrumbs"
           aria-describedby="button-description"
         >…</button>
-        <div id="button-description" class="ma__visually-hidden">This page is located more than 3 levels deep within a topic. Some page levels are currently hidden. Use this button to show and access all levels.</div>
+        <div id="button-description" aria-hidden="true" class="ma__visually-hidden">This page is located more than 3 levels deep within a topic. Some page levels are currently hidden. Use this button to show and access all levels.</div>
       </li>
     </ol>
   </nav>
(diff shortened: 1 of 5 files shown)


packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/breadcrumbs.twig
```diff
@@ -15,9 +15,18 @@
...
+         <li class="ma__breadcrumbs__item--current ma__visually-hidden" aria-current="location">{{ breadcrumbs.currentPage }}</li>
+      {% endif %}
+      <li class="ma__breadcrumbs__item ma__breadcrumbs__item__expand-indicators">
+        <button 
+          type="button"
+          class="js-breadcrumbs-button"
+          aria-label="Show all Breadcrumbs"
+          aria-describedby="button-description"
+        >…</button>
+        <div id="button-description" class="ma__visually-hidden">This page is located more than 3 levels deep within a topic. Some page levels are currently hidden. Use this button to show and access all levels.</div>
```
ygannett: Thank you so much for the all changes.

Please add `aria-hidden="true"` to `div#button-description`.  Without it, its content gets announced when screen readers are on the element.  The goal is to get it announced only when the button gets focus to deliver the information to screen reader users.

clairesunstudio: I see, ok I thought `aria-hidden` will take the div out from the accessibility tree completely so that won't be announced by screen readers when the button gets focus. I will put it back

clairesunstudio:
```suggestion
        <div id="button-description" aria-hidden="true" class="ma__visually-hidden">This page is located more than 3 levels deep within a topic. Some page levels are currently hidden. Use this button to show and access all levels.</div>
```

clairesunstudio:
Please ignore the backstop failures, as they are approved on the based branch already.

ygannett (approved):
Thank you for the change.

PR #1602, opened 2022-03-09
[React] Dp 24095 fix search bug google translate

clairesunstudio:
To test:
- run `rush install`
- in packages/react run `yalc publish`
- go to search repo and checkout `DP-24095-fix-search-bug-google-translate`
- run `npm install` and `yalc add @massds/mayflower-react`
- run `npm start`

Search PR: https://github.com/massgov/massgov-search/pull/763 

Notes:
- Installed an new eslint package in react (`rush add -p  @sayari/eslint-plugin --dev`)  to catch the JSX issues 

/*
  https://gist.github.com/azirbel/51518d919de979197a7c5c25c54a56d6
  Handy tools & articles for developing eslint rules:
    https://astexplorer.net/
    https://www.webiny.com/blog/create-custom-eslint-rules-in-2-minutes-e3d41cb6a9a0
    https://blog.maximeheckel.com/posts/how-to-build-first-eslint-rule/
*/


clairesunstudio:
I was able to fix the main bug which breaks the search app each time a query term is searched when google translate is turned on. However, when a filter or sort is applied, the google translation update collide with the react update, resulting in some content not translated on the page. 
Screen Shot 2022-03-31 at 4 09 40 PM

As the search box is the primary search interface, and people generally stay on the all tab without filtering or sorting, we should get this fix released first to unblock any searches with the use of google translate, and try to address the rest of the issues separately

clairesunstudio:
sort-teaser-title-no-translate

packages/react/src/components/organisms/PageHeader/index.js
```diff
@@ -39,7 +39,7 @@ const PageHeader = (pageHeader) => {
           { category && (
             <span className="visually-hidden">
               { category }
```
clairesunstudio:
```suggestion
             <span>{ category }</span>
```

packages/react/README.md
```diff
@@ -89,14 +89,46 @@ $ ``git add package.json package-lock.json``
...
 $ ``git commit -m 'Added purgecss library'``
 
-#### Icon Component Generation
+### Icon Component Generation
 Mayflower React's SVG icon components are generated using [SVGR](https://react-svgr.com/) at build time for both development and production modes. Currently, the `.svg` files under `src/components/base/Icon/assets` are generated into React components. When icon components are generated for production, both ES6 and CommonJS versions of the components are created using babel.
 
 A custom template used by SVGR to create the icon component files can be found within `icon-template.js`. SVGR itself is configured using `.svgrrc.js`.
 
+### Caveats
+Google translate inserts `<font>` tags for translated strings, causing the react virtual DOM to differ from the real DOM. When a JSX string literal is not wrapped in any HTML tags, and it's not the only child of its parent, React script errors when executing `removeChild` or `insertBefore` on conditionally rendered child nodes. In order to avoid Google translate widget crashing the React apps, Mayflower React components should avoid having string literal unwrapped when it has any siblings.
```
tom-fleming:
```suggestion
Google translate inserts `<font>` tags for translated strings, causing the React virtual DOM to differ from the real DOM. When a JSX string literal is not wrapped in any HTML tags, and it's not the only child of its parent, React script errors when executing `removeChild` or `insertBefore` on conditionally rendered child nodes. In order to avoid Google translate widget crashing the React apps, Mayflower React components should avoid having string literal unwrapped when it has any siblings. 
```
```suggestion
Google translate inserts `<font>` tags for translated strings, causing the react virtual DOM to differ from the real DOM. When a JSX string literal is not wrapped in any HTML tags, and it's not the only child of its parent, React script errors when executing `removeChild` or `insertBefore` on conditionally rendered child nodes. In order to avoid Google translate widget crashing the React apps, Mayflower React components should avoid having string literal unwrapped when it has any siblings. 
```

tom-fleming:
```suggestion
Google translate inserts `<font>` tags for translated strings, causing the React virtual DOM to differ from the real DOM. When a JSX string literal is not wrapped in any HTML tags, and it's not the only child of its parent, React script errors when executing `removeChild` or `insertBefore` on conditionally rendered child nodes. In order to avoid Google translate widget crashing the React apps, Mayflower React components should avoid having string literal unwrapped when it has any siblings. 
```
```suggestion
Google translate inserts `<font>` tags for translated strings, causing the react virtual DOM to differ from the real DOM. When a JSX string literal is not wrapped in any HTML tags, and it's not the only child of its parent, React script errors when executing `removeChild` or `insertBefore` on conditionally rendered child nodes. In order to avoid Google translate widget crashing the React apps, Mayflower React components should avoid having string literal unwrapped when it has any siblings. 
```

tom-fleming (approved):
LGTM

PR #1644, opened 2022-06-17
DP-25158 decreasing 1 pixel from "show # more" element border bottom

beto-aveiga:
### Description
Removes an extra pixel when "show N more" element was added

### Explanation
https://massgov.atlassian.net/browse/DP-25158?focusedCommentId=276555

### JIRA
DP-25158


clairesunstudio:
backstop failures don't see related https://output.circle-artifacts.com/output/job/a2e439ea-9a2b-4ab3-8456-01763a544b3d/artifacts/0/backstop/reports/html/index.html

clairesunstudio:
breadcrumb-pixel-1

The page moves on hover over the show more button. To be honest, I don't think this is worth fixing. It only happens when jumping between pages when different offered by numbers, and with the page load, it's hardly noticeable by 1 or 2 pixels. If we are seeing that in backstop because the offer by data changed once, I think we can ignore it, as this shouldn't happen regularly.

clairesunstudio:
let's hold off making further changes until we chat with Joe to understand the issue in the openmass backstop.

beto-aveiga:
This is outdated. In a conversation with @clairesunstudio, she explained that the real issue was not this pixel difference. 
The issue was the way collapsing/extending relationship indicators work in conjunction with the viewport resize.

PR #1653, opened 2022-07-19
Patternlab/dp 25467 fix console bug

clairesunstudio:
Fixed: 
  - project: Patternlab
    component: skipLinkTarget.js
    description: Fix skipLinkTarget.js console error. (#1653)
    issue: DP-25457
    impact: Patch

Screen Shot 2022-07-19 at 10 07 36 AM

To test:
- confirmed that the error above is fixed 
  - before: https://mayflower.digital.mass.gov/patternlab/b/develop/?p=pages-Homepage
  - after:  https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-25467-fix-console-bug/?p=pages-Homepage
- Bring into openmass PR to confirm that the feedback button on homepage no longer show up
  -  Openmass PR https://github.com/massgov/openmass/pull/1595

### DIFF
--- a/packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
+++ b/packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
@@ -6,7 +6,6 @@
 document.querySelectorAll(".ma__figure__skip-link").forEach(link => {
 
   let linkTarget = "";
-  let linkTargetHash = "";
 
   link.addEventListener("click", (e) => {
     // Find the matched linkTarget.
--- a/packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
+++ b/packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
@@ -32,26 +32,30 @@ document.querySelectorAll(".ma__figure__skip-link").forEach(link => {
       document.querySelector(location.hash).style.display = "none";
     }
   });
+});
 
-  // For TOC skip link target
-  const skipLinkTOC = document.querySelector(".ma__jump-links__skip-link");
-  const skipLinkTOCTarget = document.querySelector("#ma-end-sticky-toc");
 
+// For TOC skip link target
+const skipLinkTOC = document.querySelector(".ma__jump-links__skip-link");
+const skipLinkTOCTarget = document.querySelector("#ma-end-sticky-toc");
+
+if(skipLinkTOC && skipLinkTOCTarget) {
   skipLinkTOC.addEventListener("click", (e) => {
     skipLinkTOCTarget.setAttribute("tabindex", "0");
     skipLinkTOCTarget.focus();
   });
-
+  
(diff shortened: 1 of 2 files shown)


clairesunstudio:
patternlab backstop failure is a false positive https://output.circle-artifacts.com/output/job/25d6f846-5bbf-4fc5-a4f2-ed3623165294/artifacts/0/backstop/reports/html/index.html

packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
```diff
@@ -33,25 +32,26 @@ document.querySelectorAll(".ma__figure__skip-link").forEach(link => {
       document.querySelector(location.hash).style.display = "none";
     }
   });
-});
```
ygannett: This `document.querySelectorAll(".ma__figure__skip-link").forEach(link => {` is for skip links for figures (currently, only set up for Tableau data viz. inserts) and unrelated to the code for the TOC skip links shown below .  So, it closed here.

packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
```diff
@@ -33,25 +32,26 @@ document.querySelectorAll(".ma__figure__skip-link").forEach(link => {
...
-
 
-// For TOC skip link target
-const skipLinkTOC = document.querySelector(".ma__jump-links__skip-link");
-const skipLinkTOCTarget = document.querySelector("#ma-end-sticky-toc");
+  // For TOC skip link target
```
ygannett: When the `forEach` closes above, it causes the error you found.  So, here, we can wrap LL. 40-55 with
`if (skipLinkTOC) { }`.  Then, pages don't have TOC skip links, in other words, no TOC, wouldn't get the error.

I kept the code for these two different skip links together here, which is my bad, but thought it would be easier to manage since they have the same functionality with different targets.

ygannett (commented):
Please see my two comments.

packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
clairesunstudio: I see... I didn't realize that. good catch!

packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
clairesunstudio: I added a check for both the TOC and the Target, as I noticed jump links have TOC but not the target, and this won't apply.

ygannett (approved):

packages/patternlab/styleguide/source/assets/js/modules/skipLinkTarget.js
ygannett: Excellent!  Thank you for the change.

PR #1669, opened 2022-09-02
[React] Tabs ssr bug

clairesunstudio:
There's an issue when server-side rendering the Tab component in Mayflower react.

In Gatsby build (SSR) -  
tabs-ssr-bug

When the page is refreshed, the tabs are not showing any content. 

This is because, in `TabBody` `ComponentDidMount`, it's using document.getElementById to get elements that are not yet rendered in DOM. When server-side rendered, the state was never set when the component was initially mounted. Subsequent changes to the DOM makes the tabs content show up, because it was using `MutationObserver` to track the changes and set the state. 

I have set up a testing repo to demonstrate the issue and to enable a better debugging environment linking to local Mayflower for SSR. 

https://github.com/clairesunstudio/mayflower-ssr-tabs

You can see on initial load, the "rendered" log was never triggered:
Screen Shot 2022-09-02 at 2 27 52 PM

But when you alter a component on the page, the Tab component re-renders with the content.
Screen Shot 2022-09-02 at 2 04 40 PM

So the issue lies with the setState in componentDidMount life cycle in SSR

Related MF PR:
https://github.com/EOANF/budget/pull/306/files

### DIFF
--- a/packages/react/src/components/organisms/TabContainer/tab-body.js
+++ b/packages/react/src/components/organisms/TabContainer/tab-body.js
@@ -5,12 +5,16 @@ import { Portal } from 'react-portal/es';
 class TabBody extends React.Component {
   constructor(props) {
     super(props);
+    // console.log(props)
     this.state = {
-      nodeList: []
+      nodeList: [this.props.tabBodyRef.current] || []
     };
+    // console.log('set state: ' + global.MutationObserver)
     if (global.MutationObserver) {
       this.observer = new MutationObserver((mutations) => {
         const tabContainer = document.getElementById(this.props.tabContainerBodyId);
+        console.log('mutated!!!!')
+        console.log(mutations)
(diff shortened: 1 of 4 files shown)


packages/react/src/components/organisms/TabContainer/tab-body.js
```diff
@@ -23,15 +27,19 @@ class TabBody extends React.Component {
         });
       });
     }
+    this.observer.observe(document, { attributes: true, childList: true, subtree: true });
   }
 
   componentDidMount() {
     const nodeList = document.getElementById(this.props.tabContainerBodyId);
     if (nodeList) {
+      console.log('rendered!!!!!')
```
clairesunstudio, line 36: this was never triggered in SSR because nodeList is null on intial mount as the component has not been rendered yet.

PR #1670, opened 2022-09-05
DP-25774: Problem with the server side rendering of the tabs component

DmitriiTartynovLCM:
## Description
React hydration ignores any differences in attributes for performance reasons. So, it's not able to fix an ID mismatch and we see different IDs in the DOM and in the state. As a temporary workaround, we push different IDs to the state once the component completes its rendering. This triggers re-rendering and fixes the mismatch.

## Related Issue / Ticket

- [DP-25774](https://massgov.atlassian.net/browse/DP-25774)
- #1669

#### Today I learned...

There is a `useId()` hook introduced in React v18 that guarantees consistency of a random HTML ID across a server and a client.


clairesunstudio:
Thanks @DmitriiTartynovLCM! This does fix the tab issue in gatsby! Although, it generates a console error... I will merge this in for now, but we should really refactor this component. 
Screen Shot 2022-09-06 at 11 59 11 AM

clairesunstudio (approved):

PR #1683, opened 2022-10-26
DP-26194: Crazy attempt to automatically generate TS type declaration…

clairesunstudio:


clairesunstudio:
replaced by https://github.com/massgov/mayflower/pull/1687

PR #1685, opened 2022-10-27
React/dp 26194 fix search banner styles

clairesunstudio:
Screen Shot 2022-10-31 at 3 24 11 PM

Search banner with no filter or tabs still render an empty div resulting in excess spacing. This PR is to include the container div into conditional. Once this issue is fixed in the library, we can remove the temporary override in data hub https://github.com/massgov/data-hub-react/pull/6


clairesunstudio:
https://github.com/massgov/mayflower/pull/1729

PR #1699, opened 2022-11-15
DP-26366 Create typescript declarations for Mayflower React components

DmitriiTartynovLCM:

## Description
* Add auto-generated TypeScript declarations based on the components' proptypes.
* Enable support for React 17 and 18. Some dependencies may still be incompatible with those, but at least some components are usable.
* Replace `react-html-parser` with `html-react-parser` (for both `react` and `site` packages).
* Remove `airbnb-prop-types` from dependencies and rewrite prop types without them.
* Add some missing proptypes to the components.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-26366)

## Steps to Test

* Use `yalc` to connect it to a TS project.
* Test the type checks & prop hints from an IDE.

#### Today I learned...
JSCodeShift could be a powerful tool for the code auto-generation.


packages/react/scripts/transform.ts
```diff
@@ -0,0 +1,471 @@
+// Taken from https://github.com/mskelton/ratchet/blob/main/transform.ts
```
clairesunstudio, line 1: I thought the jscodeshift library is handling the transformation from proptypes to types. Is there a reason we need to keep our copy of this file?

packages/react/.svgrrc-ts.js
```diff
@@ -0,0 +1,8 @@
+const baseConfig = require('./.svgrrc.js')
+
+module.exports = {
+  ...baseConfig,
+  "ext": "tsx",
+  "typescript": true,
+  "template": require('./icon-template-ts'),
```
clairesunstudio, line 7: seems like icons are working too? Does the transformation have 100% coverage?

packages/react/gulpfile.js
```diff
@@ -385,8 +389,102 @@ function cleanIconDir() {
...
+    .pipe(tsProject())
+    .dts
+    .pipe(
+      // @todo Is there a better way to do it? Some TS compiler options maybe?
+      replace(/(\s+from\s+['"])(Mayflower[a-z]+)\//ig, '$1../')
+    )
+    .pipe(dest('dist'))
+}
+
+const generateTsDeclarations = series(
```
clairesunstudio, line 459: what components are excluded as of this PR? 
Also, are defaultProps handled by the transformation?

packages/react/scripts/jsx-to-tsx
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+const { existsSync, statSync, copyFileSync } = require('fs')
+const { exec } = require("child_process");
+
+const args = process.argv.slice(2)
+const sourceFilepath = args[0];
+if (!sourceFilepath) {
+  console.error('Please pass the source file path.');
+  process.exit(1);
```
clairesunstudio, line 17: can we let the default convert all components instead of exiting?

packages/react/gulpfile.js
clairesunstudio: e.g. I noticed that CompoundSlider has defaultProps that doesn't seem to reflect in types

clairesunstudio:
Tests failures are false positives - they are not related to this PR.

packages/react/.svgrrc-ts.js
clairesunstudio: confirmed true

packages/react/gulpfile.js
clairesunstudio: defaultProps are not handled, but all components have types declarations

packages/react/scripts/transform.ts
clairesunstudio: custom code to handle additional stuff, e.g. proptype `shape` https://github.com/massgov/mayflower/blame/767d4097e95ac9aad8fe60196e600501c66fddf5/packages/react/scripts/transform.ts

packages/react/scripts/jsx-to-tsx
DmitriiTartynovLCM: Just to keep in text what we discussed already. This script was developed for the manual conversion. Right now it could be used for debugging the TS conversion of a single component.

PR #1700, opened 2022-11-15
Patternlab/dp 26365 new feedback form

clairesunstudio:
Changed:
  - project: Patternlab
    component: MassFeedback, MassFeedbackSuccess
    description: Implement new design and validation logic, and clean up unused components and variants. (#1700)
    issue: DP-26365
    impact: Minor
Added:
  - project: Patternlab
    component: HelpTip
    description: Add HelpTip organism. (#1700)
    issue: DP-26365
    impact: Minor
  - project: Patternlab
    component: IconClose
    description: Add "close" icon. (#1700)
    issue: DP-26365
    impact: Minor
Fixed:
  - project: Patternlab
    component: InputRadio
    description: Fix input radio styles. (#1700)
    issue: DP-26365
    impact: Minor

-----

## Before:
https://mayflower.digital.mass.gov/patternlab/?p=viewall-organisms-feedback
Screen Shot 2022-11-22 at 11 56 51 AM

## After:
https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-26365-new-feedback-form/?p=viewall-organisms-feedback
Screen Shot 2022-11-22 at 11 57 57 AM

---

## Review Notes
1. Default form: https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-26365-new-feedback-form/?p=organisms-mass-feedback
> This screen shows the feedback form by default on every page. 
> Enter and clear inputs to check its validation logic, see [specs in figma] (https://www.figma.com/file/ehFzAlJHDkq2ToJPLsjocn/Feedback-Redesign?node-id=109%3A561&t=JmkmB0scc971W06S-0)
3. Form expanded with the urgent request alert: https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-26365-new-feedback-form/?p=organisms-mass-feedback-alert.
> This screen shows the form and help tip expanded, with the urgent request alert checked off. 
4. Thank you screen: https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-26365-new-feedback-form/?p=organisms-mass-feedback-success
> Note that the content in the thank you screen are fully configurable rich text. The contact link condition will be added as data in openmass.


clairesunstudio:
Not sure why keep seeing the those patternlab backstop failures https://app.circleci.com/pipelines/github/massgov/mayflower/4475/workflows/c72cd22a-e81a-4a1d-ab9b-b694d343f8b7/jobs/50907
These seem like false positives, as I'm seeing the image on the page.
https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-26365-new-feedback-form/?p=pages-contact-row-examples

clairesunstudio:
Changes in this PR impact feedback form on mass.gov https://pr1800-5jp8yiaecdsmotpslwrjkdsdqtgznvmt.tugboatqa.com/topics/higher-education 

Please merge this at the same time as the openness changes are ready.

clairesunstudio:
@ygannett can you do an accessibility review on the new feedback form in Mayflower?

ygannett:
@clairesunstudio Please see my notes on https://massgov.atlassian.net/browse/DP-26365 for the a11y review.

clairesunstudio:
@arthurbaghdas, per our last discussion:

1. Check with Joe on the org contact info flag logic - checking the box will display the contact message by default on both the form and the success screen. I added a `showWarnMsg` boolean for this purpose.
2. I combined the form and the success templates into one. Added a `submitted` boolean, this will control which screen to show. 
3. Submission logic: I didn't implement this in mayflower, I think this should be in openmass, as the success page should be rendered post-submission not before. Since there will still need to be JS in openmass to handle the form submission, I think that's the natural place to add that logic. 


I'm still working through some accessiblity fixes, but hopefully this unblocked you for implementing the form logic in openmass. Let me know in the morning if you want to discuss more. thanks!

arthurbaghdas:
Hi @clairesunstudio , another thing that I've found is that the negative text area always have a required attribute set, which means whenever I try to submit positive feedback the form is not submitting.
I think we should set the required attribute based on the user selection and this should be done in the scope of Mayflower front-end Implementation

clairesunstudio:
Interesting, it was not working like this before. Let me look into this.

clairesunstudio:
@arthurbaghdas, what you mentioned above should be fix now. Can you please check again?

clairesunstudio:
Addressed all accessibility comments

clairesunstudio:
Test in openmass PR https://github.com/massgov/openmass/pull/1785

packages/patternlab/styleguide/source/_patterns/03-organisms/feedback/mass-feedback.twig
```diff
@@ -1,9 +1,10 @@
-<div data-nosnippet="true" class="ma__mass-feedback-form">
-  <h2 class="visually-hidden">Feedback</h2>
-  <form class="ma__mass-feedback-form__form" method="post" novalidate action="{{ formAction }}" id="{{ formId }}">
-    <fieldset class="ma_feedback-fieldset feedback-load" role="radiogroup"{% if title.required %} aria-required="true"{% endif %}>
+<div data-nosnippet="true" class="ma__mass-feedback-form" id="feedback">
+  <h2 class="ma__mass-feedback-form__heading">{{ heading }}</h2><span class="ma__visually-hidden"> with your feedback</span>
```
ygannett: Can you put the `span` in the `h2`?  Then, the content of the span is announced as a part of the heading.  Currently, users need to go to the `span` to hear the content.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/help-tip.twig
```diff
@@ -0,0 +1,47 @@
...
+      <span>{{ helpTip.textBefore }}</span>
+    {% endif %}
+    <button
+      class="ma__help-tip__trigger {{ helpTip.expanded ? 'ma__help-tip__trigger--active' : '' }}"
+      id="trigger-{{ helpTip.id }}"
+      aria-expanded="{{ helpTip.expanded ? 'true' : 'false' }}"
+      aria-controls="help-tip-content-{{ helpTip.id }}"
+    >
+      <span>{{ helpTip.textTrigger }}</span>
+      <span class="ma_help-tip__label-a11y ma__visually-hidden">Open tooltip to learn more.</span>
```
ygannett: When the help tip is set up for `aria-describedby`, the button for the tooltip in the container is not interactive with screen readers although the tooltip container content is announced followed by the button label. Screen reader users don't miss the information even the tooltip is collapsed.  But, the visually hidden text becomes not necessary as they don't need to open the tooltip anymore.  Let's remove L.16.  I'm sorry that I asked you to make this change though. 

I believe that keeping the help tip as `aria-describedby` help ensure SR users receive the information with the proper content flow while keyboard and mouse users can click the button to display the tooltip content.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/help-tip.twig
```diff
@@ -0,0 +1,47 @@
...
+    {% endif %}
+  </span>
+  <div
+    class="ma__help-tip__container ma__help-tip__container--c-primary ma__help-tip__content {{ helpTip.expanded ? '' : 'collapsed' }}"
+    id="help-tip-content-{{ helpTip.id }}"
+    style="max-height:none"
+  >
+    <button
+      class="ma__help-tip__close-mobile"
+      aria-label="close tooltip"
```
ygannett: Based on above comment, this button doesn't make sense to screen reader users.  Let's remove L.30 aria-label.  It can be commented out for the future when interaction within aria-describedby content becomes available.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/help-tip.twig
```diff
@@ -0,0 +1,47 @@
...
+    >
+      {{ icon('close') }}
+    </button>
+    <button
+      class="ma__help-tip__close-desktop"
+      aria-label="close tooltip"
```
ygannett: L.36.  Same as above comment.

ygannett (commented):
Thank you for all the changes.  We have a couple more things to work on. 

When the radio button is clicked, that triggers to display the text box for its associated option.  This is my guess, but it seems the display of the box is considered as a change to trigger the alert and resulting screen readers announce the character count while focus is still on the radio button, which doesn't make sense to SR users.
This is not happening with the current prod. feedback module.  Both markups are set up in the same way.  Not quite sure what's causing this issue.
Array

One way to resolve this could be adding `aria-hidden="true"` to `span.remainingChar` when appending it in /assets/js/modules/formInputs.js just to ensure the span is not visible to screen readers, then as the text area gets focus remove the `aria-hidden="true"`.  Any other solutions you think would be more efficient is welcome!

packages/patternlab/styleguide/source/_patterns/03-organisms/feedback/mass-feedback.twig
```diff
@@ -13,78 +14,75 @@
 
     <div class="ma__mass-feedback-form__fields">
       <fieldset class="ma_feedback-fieldset">
```
ygannett: Let's remove the `fieldset` since it's not necessary in this case as we've replaced the `legend` with `label`.

ygannett:
This error message is displayed by hitting the submit button. 
Array
I'm afraid that I missed this one at the last review.  Screen reader doesn't recognize it, just because it's displayed above the button after hitting it.  

To make the error message reactive like the other error message, it needs `role="region"`, `aria-live="polite"`, and its ID, also added to the `textarea`'s `aria-controls`.

I think we should think about how to handle this user error:
1. Should we wait to display the message until users hit the button?
2. It's a required field. Is it better to flag it when a user leaves the field empty?(= when the field loses focus after it has focus.)

As the error message is announced, I think it's helpful to set focus on the field, so a user can correct their error on the spot.  To screen reader users, it's not easy to find/locate a field with error to make a correction.

The field is announced as "required", then a user still leaves as blank.  It might be better to remind them it's a required field in the error message.

I'm open for discussion for better solution.  This one is a bit tricky and I see people would think different ways.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/help-tip.twig
clairesunstudio: Are you suggesting adding `aria-describedby` instead? So it will always announce the helptext using screen readers?

clairesunstudio:
re: character remaining announcement, I'm getting the opposite experience using VO - the problem you described is in prod mass.gov and not with the new changes. 

re: error message, I think we should keep this experience consistent - it announces when the error shows up, which is when the user hits the submit button. That's usually when required fields are validated, onblur can be too much as they might not be ready to submit yet and it's ok to leave it empty if they don't intend to submit

clairesunstudio:
To handle the error messages appropriately:
>1. To make the error message reactive like the other error message, it needs role="region", aria-live="polite", and its ID, also added to the textarea's aria-controls.

I found that the appropriate role might be "alert", and it's one of the live regions so we don't need to add it explicitly. 

> 2. Add `aria-controls`.
> 3. find the first missed required field and set focus.

ygannett:
Happened to noticed this duplicate component:
Array
It seems that a character count snippet is added whenever the text box gets focus.

ygannett:
I'm afraid that the contact info. error message's aria-live is cancelled as it positions after the character count message:  https://gaurav5430.medium.com/quick-accessibility-wins-multiple-aria-live-on-single-action-caveat-b79a6f41e7cc
So, to solve this, we could group the char. count message and the error message together in one alert message container.  The container has `aria-live="polite"`, `aria-hidden="true`.  Then, by adding `aria-atomic="true"`, we can ensure all messages are announced.

I tried to avoid this approach by playing around with aria attributes and roles, but nothing works.

Also, we can remove `aria-controls` from `textarea` as the current approach doesn't work.

clairesunstudio:
@ygannett is going to experiment with the fixes in a follow-up PR.

clairesunstudio:
Screen Shot 2023-01-10 at 5 35 53 PM

Based on the decision in standup today - I disabled SR announcing the character count live, in favor of the error messages. Making the character count a SR notation as part of the description (help text) in data - this will need to be added in the openmass PR. 

```
    "helpText": "The feedback will only be used for improving the website. If you need assistance, please <a href='mass.gov'>contact the Department of xxxxx</a>. <span class='ma__visually-hidden'>Please limit your input to 500 characters. </span>",
```

tom-fleming (approved):

PR #1704, opened 2022-11-29
DP-26366: Make the title customizable on the `ResultsHeading` component.

DmitriiTartynovLCM:

## Description
The title could be more flexible. The default one doesn't work well for our case, e.g. it adds the ` for:` suffix even without any tags.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-26366)

### DIFF
--- a/packages/react/src/components/molecules/ResultsHeading/index.js
+++ b/packages/react/src/components/molecules/ResultsHeading/index.js
@@ -19,11 +19,11 @@ import Tags from 'MayflowerReactMolecules/Tags';
 
 function getTitle(resultsHeading) {
   if (resultsHeading.title !== undefined) {
-    return resultsHeading.title
+    return resultsHeading.title;
   }
 
   const resultsHeadingTotal = resultsHeading.totalResults ? ` of ${resultsHeading.totalResults} for: ` : '';
-  return `Showing results ${resultsHeading.numResults}${resultsHeadingTotal}`;
+  return`Showing results ${resultsHeading.numResults}${resultsHeadingTotal}`;
 }
 
 const ResultsHeading = (resultsHeading) => {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/molecules/ResultsHeading/index.js
```diff
@@ -17,21 +17,29 @@ import ButtonToggle from 'MayflowerReactButtons/ButtonToggle';
 import SelectBox from 'MayflowerReactForms/SelectBox';
 import Tags from 'MayflowerReactMolecules/Tags';
 
+function getTitle(resultsHeading) {
+  if (resultsHeading.title !== undefined) {
```
clairesunstudio, line 22:
```suggestion
  if (resultsHeading.title) {
```
I think it should check if it has a truthy value here

clairesunstudio (commented):
just one comment, otherwise looks good!

packages/react/src/components/molecules/ResultsHeading/index.js
```diff
@@ -17,22 +17,33 @@ import ButtonToggle from 'MayflowerReactButtons/ButtonToggle';
...
   const { tags } = resultsHeading;
   const selectBoxProps = resultsHeading.selectBox;
   const buttonToggleProps = resultsHeading.buttonToggle;
   return(
     <div className={classes}>
       <div className="ma__results-heading__container">
-        <div className="ma__results-heading__title">
-          {resultsHeadingTitle}
-        </div>
+        { title && (
```
clairesunstudio, line 42: I don't think this check is necessary, as there's always a fallback it seems.

clairesunstudio: nvm I see, you want the empty string to hide the title element completely

clairesunstudio (approved):

PR #1716, opened 2022-12-23
DP-26745: A11y improvements to `Tags` and `ButtonTag`

DmitriiTartynovLCM:

## Description

* Let the tag buttons render child nodes. This could be necessary to add visually hidden content.
* Render the icon in HTML rather than CSS. This way it could be hidden from screen readers.
* Add a visually hidden description prop to the button tag component.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-26745)


packages/react/src/components/atoms/buttons/ButtonTag/index.js
```diff
@@ -25,8 +33,10 @@ ButtonTag.propTypes = {
   type: PropTypes.string.isRequired,
   /** The label text of the sort button */
   value: PropTypes.string.isRequired,
-  /** The label text of the sort button */
-  text: PropTypes.string.isRequired,
+  /** The content of the button */
+  text: PropTypes.node.isRequired,
+  /** The visually hidden description of the button */
+  description: PropTypes.node,
```
clairesunstudio, line 39: why is the description a node? Can't it just be a string, since it's for screen reader only?

packages/react/src/components/molecules/Tags/index.js
```diff
@@ -64,11 +65,13 @@ class Tags extends React.Component {
...
         type: The type of tag <br />
-        text: The text displayed by the tag (required) <br />
+        text: The content displayed by the tag (required) <br />
+        description: The visually hidden description of the tag <br />
         value: The value of the tag  */
   tags: PropTypes.arrayOf(PropTypes.shape({
     type: PropTypes.string,
-    text: PropTypes.string.isRequired,
+    text: PropTypes.node.isRequired,
+    description: PropTypes.node,
```
clairesunstudio, line 74: same comment here. Can we reference the ButtonTag.propTypes here instead? 
```
PropTypes.arrayOf(PropTypes.shape(ButtonWithIcon.propTypes))
```

packages/react/src/components/atoms/buttons/ButtonTag/index.js
DmitriiTartynovLCM: Value of this prop is rendered into the content. It could be anything that is a valid React node. It doesn't look like we should additionally restrict this prop. What if developers want to add additional markup for screen readers, e.g. a few HTML tags with `aria-*` attributes?

packages/react/src/components/molecules/Tags/index.js
DmitriiTartynovLCM: Unfortunately, we can't just reference `ButtonTag.propTypes` since not all of these props are allowed here. We control the `handleClick` prop in this component, so it can't be passed. Once Mayflower switches to TS, it will be possible to only take a a subset of the props. But, our JSCodeShift converter isn't smart enough to handle it at the moment, as I see.

clairesunstudio: I see...ok

clairesunstudio (approved):

PR #1717, opened 2022-12-27
DP-26714: Refactor GenTeaser into multiple separate components

DmitriiTartynovLCM:
## Description

The purpose of the change is to minimize the bundle size. Just one sub-component needs the `moment` library. It doesn't matter if we use this sub-component or not, the `moment` library was in the client-side bundle anyway. With the suggested changes, we saved ~22Kb, which is ~12% of the overall page bundle size in our case.

There are no changes in the sub-components themselves.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-26714)

## Steps to Test

Check the component in the React Storybook.

* `GenTeaser`


clairesunstudio (commented):
Upon review and some research, I'm fine with your approach to convert the subcomponents into separate components to optimize the bundle size - the other alternatives are to remove the moment library (seems like an overkill for the our purpose) or to add treeshaking - which in comparison will require higher effort at the moment. 

This does efficiently eliminate the momentJS library when unused, which is of a substantial size (snapshots taken in datahub). 
Before:
Screen Shot 2022-12-28 at 4 13 29 PM
After: 
Screen Shot 2022-12-29 at 9 44 41 AM



Based on my findings, this issue with subcomponent is usually handled by treeshaking by the bundler. I'm curious - aside from this specific issue - that if there's a way to turn on treeshaking and turn off side effects  in nextJS, which might clean up more unused resources. Thoughts?

packages/react/src/components/organisms/GenTeaser/TeaserOrgs/index.js
```diff
@@ -1,81 +1,3 @@
...
-      </span>
-    );
-  }
-}
-
-TeaserOrgs.propTypes = {
-  /** A comma seperate list of organizations. */
-  orgs: PropTypes.string.isRequired
-};
+import TeaserOrgs from 'MayflowerReactGenTeaser/GenTeaserOrgs';
```
clairesunstudio: Can we remove this file? Now that all sub components are separate, I think we should use the same naming structure for the orgs also

packages/react/src/components/organisms/GenTeaser/TeaserSearch/index.js
```diff
@@ -1,73 +1,3 @@
...
-  /** The target url of the search bar */
-  target: PropTypes.string.isRequired,
-  /** The id of the search bar */
-  id: PropTypes.string.isRequired,
-  /** The query input variable to replace in the target url with the user entered term */
-  queryInput: PropTypes.string.isRequired,
-  /** Placeholder text of the search bar. */
-  placeholder: PropTypes.string.isRequired
-};
+import TeaserSearch from 'MayflowerReactGenTeaser/GenTeaserSearch';
```
clairesunstudio: Can we remove this file? Now that all sub components are separate, I think we should use the same naming structure for the search also

packages/react/src/components/organisms/GenTeaser/GenTeaser.md
```diff
@@ -4,36 +4,36 @@ The GenTeaser component provide a flexible and extensible content container with
...
-    <GenTeaser.Stat>{dataCatalog.stat}</GenTeaser.Stat>
-    <GenTeaser.Title title={dataset.title} />
-    <GenTeaser.Emphasis>
-      <GenTeaser.Date date={dataset.date} />
-      <GenTeaser.Orgs orgs={dataset.orgs} />
-    </GenTeaser.Emphasis>
-    <GenTeaser.Description description={dataset.description} />
-    <GenTeaser.SearchBar search={dataCatalog.search} />
-    <GenTeaser.SubLinks>
+<GenTeaserContainer onClick={(e) => action(e)} onKeyDown={(e) => action(e)}>
```
clairesunstudio, line 7: Can we still call the wrapper `GenTeaser`? This is to follow the naming convention. For example, see MUI https://mui.com/material-ui/api/card/ and ReachUI https://reach.tech/combobox

packages/react/src/components/organisms/GenTeaser/index.js
```diff
@@ -15,708 +15,57 @@
...
+import InfoDetails from 'MayflowerReactGenTeaser/GenTeaserInfoDetails';
+import Tags from 'MayflowerReactGenTeaser/GenTeaserTags';
+
+const GenTeaser = (props) => (
+  <Container {...props} />
+);
+
+GenTeaser.propTypes = Container.propTypes;
+
+GenTeaser.Details = Details;
```
clairesunstudio, line 48: Does this file mean that someone can still implement GenTeaser with subcomponents if they opt-in to import from the index?

clairesunstudio: If so, can you please add to the documentation outlining the two options? And maybe add some explanation of why we did what we did here and that this is the recommended import option because it excludes unused subcomponents.

clairesunstudio (commented):
Just some small naming change and docs comments

packages/react/src/components/organisms/GenTeaser/index.js
DmitriiTartynovLCM: Just added the documentation about the previous approach.

packages/react/src/components/organisms/GenTeaser/TeaserSearch/index.js
DmitriiTartynovLCM: Removed it.

packages/react/src/components/organisms/GenTeaser/TeaserOrgs/index.js
DmitriiTartynovLCM: Removed it too.

DmitriiTartynovLCM:
@clairesunstudio [Tree shaking](https://webpack.js.org/guides/tree-shaking/) must be enabled by default in Webpack 5. This means NextJS also has it enabled unless they  And, Mayflower React package has `sideEffects` specified in the `package.json` file. So, tree shaking should work for us. But, it only handles different imports from a module. As I understand, it can't perform tree shaking with properties of a single object exported from a module.

Please see massgov/data-hub-react#152 for a tree shaking test.

clairesunstudio (approved):

PR #1721, opened 2023-01-09
DP26647 add height to org nav mobile

clairesunstudio:
https://github.com/massgov/openmass/pull/1830


clairesunstudio:
replaced by #1723

PR #1732, opened 2023-01-17
DP-25865: Switch CircleCI to use OIDC for AWS

chrismshelton:

## Description
This PR switches from using AWS keys for deployment to OIDC. The AWS roles used in the deployment have also been restricted in terms of the CloudFront distributions they can access.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-25865)

## Steps to Test

1. Deploy to AWS and ensure the deployment works correctly.

#### @TODO

* Once this is deployed, we can delete the AWS keys from CircleCI. We can also remove the `--profile default` from all of the AWS commands in the build, as it is only needed to override the use of the environment variables.

#### Today I learned...
- https://circleci.com/docs/openid-connect-tokens/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,7 +228,7 @@ jobs:
       - when:
           condition:
             matches:
-              pattern: "^[0-9]+\.[0-9]+\.[0-9]+$"
+              pattern: '^[0-9]+\.[0-9]+\.[0-9]+$'
               value: << pipeline.git.tag >>
           steps:
             - aws-cli/setup:
@@ -271,7 +271,7 @@ jobs:
       - when:
           condition:
             matches:
-              pattern: "^[0-9]+\.[0-9]+\.[0-9]+$"
+              pattern: '^[0-9]+\.[0-9]+\.[0-9]+$'
               value: << pipeline.git.tag >>
           steps:
             - aws-cli/setup:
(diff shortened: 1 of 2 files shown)


tom-fleming:
The changes look good to me. I don't think the assets_build job error is related, but it might be worth rerunning the job to test. Additionally, we might want to test the different testing environments these commands build once merged @chrismshelton @clairesunstudio

tom-fleming (approved):

.circleci/config.yml
```diff
@@ -272,27 +268,33 @@ jobs:
...
+              pattern: '^[0-9]+\.[0-9]+\.[0-9]+$'
+              value: << pipeline.git.tag >>
+          steps:
+            - aws-cli/setup:
+                role-arn: $AWS_DEPLOYMENT_ROLE_ARN
+            - run: |
+                cd packages/core
+                aws s3 sync storybook-static "s3://mayflower.digital.mass.gov/core/v/$CIRCLE_TAG" --delete --profile default
+                # Use cp instead of sync for root copy.  sync lists all keys and is too slow.
+                aws s3 cp --recursive storybook-static "s3://mayflower.digital.mass.gov/core"
```
clairesunstudio: question, does the `--profile default` flag only needs to be set once for each job?  I'm seeing sometimes it's repeated in every step, for example in `react_build_storybook`. Just want to make sure that it works as expected without the flag in subsequent steps

chrismshelton: Good catch! It should always be set on every aws command. It should be good now :+1:

clairesunstudio (approved):

PR #1740, opened 2023-02-06
all/DP-27164-fix-google-translate-styling-caused-by-google

clairesunstudio:
Fixed:
  - project: Assets
    component: GoogleTranslate, UtilityNav
    description: Fix Google Translate styles caused by a classname change by Google. (#1740)
    issue: DP-27164
    impact: Patch

To Test:
- please confirm that both implementations are fixed - including react and patternlab
- test in openmass https://github.com/massgov/openmass/pull/1871


tom-fleming (approved):
LGTM!

PR #1743, opened 2023-02-14
DP-24290 Navigation labelling for screen readers

ygannett:

## Description

- Remove `aria-labelledby` on the `<nav>`s (blue bar and horizontal nav) and its associated content.
- Set the label for the `<nav>` as "main navigation" for the flyout version (non home pages)
- Set up to switch the `<nav>` on the blue bar on home page:
        - page width > 840 (utility nav only):  "Language options and quick access links"
        - page width < 840 (topic links + utility nav):  "main navigation"
        - Keep “Main navigation” for the second set.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-24290)
- [Github issue]()

### Home page

1. In inspection, locate `nav.ma__header__hamburger__nav`.
2. In desktop view (screen width > 840), find the `aria-label` value as "Language options and quick access links" and the blue bar only shows the utility nav.
```
<nav class="ma__header__hamburger__nav" 
         aria-label="Language options and quick access links" 
         id="main-navigation-mobile">
```
3. Locate `nav.ma__header__nav`, find its `aria-label` is "main navigation".
```
<nav class="ma__header__nav" 
         id="main-navigation" 
         aria-label="Main Navigation">
```
4. Set the screen to the mobile view (screen width < 840), find the `aria-label` as "main navigation" and the utility nav is not showing anymore.

 ### Non-Home page
1. In inspection, locate `nav.ma__header__hamburger__nav`.
2. In desktop view (screen width > 840), find the `aria-label` value as "main navigation".
```
<nav class="ma__header__hamburger__nav" 
         aria-label="main navigation" 
         id="main-navigation-mobile">
```
3. In mobile view (screen width < 840), find the `aria-label` value is still "main navigation".

## Additional Notes:

Anything else to add?

*

#### @TODO

*


packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-hamburger.twig
```diff
@@ -10,13 +10,14 @@
 
   <a class="ma__header__hamburger__skip-nav" href="#main-content">skip to main content</a>
 
-  <nav class="ma__header__hamburger__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  <nav class="ma__header__hamburger__nav" id="main-navigation" role="navigation">
+    <span id="header-nav-label" class="ma__visually-hidden">Main navigation</span>
```
clairesunstudio: I think we should remove this line
```suggestion
```
Screen Shot 2023-02-28 at 3 43 02 PM

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-mixed.twig
```diff
@@ -10,13 +10,14 @@
 
   <a class="ma__header__hamburger__skip-nav" href="#main-content">skip to main content</a>
 
-  <nav class="ma__header__hamburger__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  <nav class="ma__header__hamburger__nav" id="main-navigation" role="navigation">
+    <span id="header-nav-label" class="ma__visually-hidden">Main navigation</span>
```
clairesunstudio:
```suggestion
```
same here. The role is sufficient.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
```diff
@@ -1,4 +1,6 @@
 <div class="ma__utility-nav js-util-nav">
+  {# Unable to use aria for labeling since inconsistent support for the tags by screen readers as of Feb. 22, 2023. #}
+  <span id="utility-nav-label" class="ma__visually-hidden">Language options and quick access links</span>
```
clairesunstudio: I like this name for the "utility nav". However, I'd like to understand a little more about the reason why `aria-label` would not work here. Not even `aria-labelledby`? 
Because currently, this is announced as a separate text, and I think should be the label for the list. 
Screen Shot 2023-02-28 at 3 52 09 PM

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-mixed.twig
```diff
@@ -78,7 +79,9 @@
     {% endif %}
   </div>
 
-  <nav class="ma__header__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  {# Unable to use aria for labeling since inconsistent support for the tags by screen readers as of Feb. 22, 2023. #}
+  <nav class="ma__header__nav" id="main-navigation" role="navigation">
+    <span id="horizontal-nav-label" class="ma__visually-hidden" aria-hidden="true">Navigation by topics</span>
```
clairesunstudio: I think ideally this label should be announced together with the role "navigation". Currently it's announced separately on a text element within the navigation element

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-hamburger.twig
ygannett: You don't think it needs a label to group all navigation units as a whole?  I was questioning the necessity of it, but didn't come to a conclusion.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
ygannett: During the testing the changes, I found out that JAWS doesn't support `aria-label` for static elements.  So, tried `aria-labelledby`, then NVDA doesn't support it on static elements.  They don't announce those attributes on static elements.  Both cannot be used because VoiceOver announces them.  

Using headings would be flagged for wrong heading levels since they would appear prior to the page title of `h1` as `h2` or `h3`.  

I hate to leave out the semantic association between the elements and their labels, but in this case, letting the reading order handle their association seem to be a reasonable approach.  This is a compromise to ensure all screen readers announce the labels.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-hamburger.twig
```diff
@@ -10,13 +10,14 @@
 
   <a class="ma__header__hamburger__skip-nav" href="#main-content">skip to main content</a>
 
-  <nav class="ma__header__hamburger__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  <nav class="ma__header__hamburger__nav" id="main-navigation" role="navigation">
```
clairesunstudio:
```suggestion
  <nav class="ma__header__hamburger__nav" id="main-navigation" aria-label="main navigation">
```
since it's already a nav element, role is not necessary

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-hamburger.twig
clairesunstudio: I think this should be an aria-label instead of a piece of standalone text. We can put this as an aria-label in the nav tag, see the APG example https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-mixed.twig
```diff
@@ -10,13 +10,14 @@
 
   <a class="ma__header__hamburger__skip-nav" href="#main-content">skip to main content</a>
 
-  <nav class="ma__header__hamburger__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  <nav class="ma__header__hamburger__nav" id="main-navigation" role="navigation">
```
clairesunstudio:
```suggestion
  <nav class="ma__header__hamburger__nav" id="main-navigation" aria-label="main navigation">

```

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-mixed.twig
```diff
@@ -78,7 +79,9 @@
     {% endif %}
   </div>
 
-  <nav class="ma__header__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  {# Unable to use aria for labeling since inconsistent support for the tags by screen readers as of Feb. 22, 2023. #}
+  <nav class="ma__header__nav" id="main-navigation" role="navigation">
```
clairesunstudio:
```suggestion
  <nav class="ma__header__nav" id="main-navigation" aria-label="main navigation">

```

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-mixed.twig
```diff
@@ -78,7 +79,9 @@
     {% endif %}
   </div>
 
-  <nav class="ma__header__nav" aria-label="main navigation" id="main-navigation" role="navigation">
+  {# Unable to use aria for labeling since inconsistent support for the tags by screen readers as of Feb. 22, 2023. #}
```
clairesunstudio: I think aria-label is fine to use on navigation, see https://www.davidmacd.com/blog/does-aria-label-override-static-text.html

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/header-mixed.twig
clairesunstudio: Also, after inspecting the whole file, I noticed that this nav is at the same level as the `.ma__header__hamburger__nav` and they both have the same id. I think they should have the same label as they are just different views of the same thing. I would suspect there is a duplicate id issue.

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
clairesunstudio: It's good to know that neither `aria-label` nor `aria-labelledby` works consistently with lists. I was aware of that before. Iagree in this case there's no better option if we want to label the list then.

packages/assets/scss/03-organisms/_header-mixed.scss
```diff
@@ -4,6 +4,10 @@
 
   .ma__header__mixed {
 
+    #header-nav-label {
```
clairesunstudio: this ID doesn't seem to exist, does this need to be cleand up?

packages/assets/scss/03-organisms/_utility-nav.scss
```diff
@@ -202,14 +202,18 @@ $utility-nav-height: 43px;
...
       @media ($bp-header-toggle-max) {
 
-        @include ma-accordion-icon;
-        display: block;
-        width: 100%;
+        & span {
 
-        &:after {
+          @include ma-accordion-icon;
+          // display: block;
```
clairesunstudio: are these style changes intentional? What are they for?

packages/patternlab/styleguide/source/assets/js/modules/mainNavHamburger.js
```diff
@@ -969,6 +967,12 @@ if (hamburgerMenuContainer) {
       "iframe.goog-te-menu-frame.skiptranslate"
     );
 
+    // Hide "+" symbol from screen readers.
+    let hamburgerGTSymbol = utilNarrowNav.querySelector(".ma__utility-nav__translate #google_translate_element a span");
+    if (hamburgerGTSymbol) {
+      hamburgerGTSymbol.setAttribute("aria-hidden", "true");
```
clairesunstudio, line 982:
```suggestion
      hamburgerGTSymbol.setAttribute("aria-hidden", "true");
```

Is this for mobile? On mobile, I'm still hearing the "plus" being announced next to "select language". I think it's because the + sign is added in css content. You can try this in css https://www.w3.org/TR/css-content-3/#accessibility to set the alt text to empty

packages/assets/scss/03-organisms/_utility-nav.scss
ygannett: Yes, matching the structure of the other item that has "+", but screen readers don't announce the symbol although the Google Translate has `<a>` instead of `<button>`.  So, `aria-hidden="true"` can be added to the span.
```
<li class="ma__utility-nav__item">
  <button class="ma__utility-nav__link js-util-nav-toggle" aria-haspopup="true" aria-label="Log in links for this page" aria-expanded="false">
    <svg aria-hidden="true" focusable="false">...</svg>
    <span>Log in to...</span>

    <!-- This .toggle-indicator has ::after that contains "+" -->
    <span class="toggle-indicator" aria-hidden="true"></span>
  </button>
  ...
</li>
```

packages/assets/scss/03-organisms/_header-mixed.scss
ygannett: Done.

packages/patternlab/styleguide/source/assets/js/modules/mainNavHamburger.js
ygannett: Yes, it's for mobile.
First, I thought it was because of pseudo content, but all the top menus of the navigation and the log in to button have the "+" as pseudo content of `button > span.toggle-indicator`.  Their symbols don't get announced.

In the current code, the the symbol is added to `#google_translate_element a`.  To add `aria-hidden="true"`, I added `span` to the `a` as its child element.  This new structure matches to the top menus and the log in to button in the flyout that has the "+" symbol.  Only difference is they are `button`s as GT is `a`.
Mass_Gov_and__header-mixed_scss_—_mayflower

With the testing on Nov. 22, It's confirmed that JAWS, NVDA and VO don't announce the symbol.

clairesunstudio: Screen Shot 2023-11-22 at 12 40 13 PM

Weird, I'm still hearing the plus...

clairesunstudio: ok so I gather that the UAT link is not always reliable https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-24290_navigation-labels/?p=pages-Homepage

I pulled this down locally and tested it again, now the phrase select language is not being announced, because it's wrapped inside of the aria-hidden span
Screen Shot 2023-11-22 at 12 48 15 PM

clairesunstudio:
The only other thing I notice is that in hamburger menu, the grouped "utility items" list doesn't have a label. I wonder if we want to add that new utility group label to the `<ul>` element, so that it's more consistent across pages? Do you think that will be an easy change? If that's a straightforward change, maybe we can include that in this PR,  if not, we can address that separately maybe?

Screen Shot 2023-11-22 at 12 53 11 PM

packages/patternlab/styleguide/source/assets/js/modules/mainNavHamburger.js
ygannett: There are `span`s under `a` and all of them got `aria-hidden="true"`.  This was causing the issue although strangely all screen readers announced "select language" as I tested.

To resolve this, targeted the last `span` under `a` to have `aria-hidden` and the pseudo-class to add the symbol.

ygannett: `aria-label` is added to the UL. It's confirmed that JAWS and NVDA announced it.

ygannett:
`aria-label` is added to the UL. It's confirmed that JAWS and NVDA announced it.

ygannett:
Backstop failures are map positioning, not related to this PR's changes:  https://output.circle-artifacts.com/output/job/69146413-3c60-4b56-aac6-32a0b9fa3ab0/artifacts/0/backstop/reports/html/index.html

clairesunstudio (approved):

PR #1752, opened 2023-03-22
DP-26304 [SI-a11y] Empty aira-label with decorative element

ygannett:
## Description

For the press teaser, currently its image link, which is a linked `span` with a background image, doesn't have `aria-label` (= empty) since its `a` has title for the link label.  The `span` is also overridden with `role="img"`, which is not necessary in this case.

- Set up a condition to add `aria-label` and `role="img"` only when `image.alt` value is available.
- Set up a condition to add `title` for the link label on `a` only `pressTeaser.title.text` is available.

The org page press teasers don't have `alt` value, so the `span` stays as a generic container with its background image and no `aria-label`.  No override to be an image element.

In case the component is used differently and the background image has information to deliver,  the `span` is presented as an image element with its `aria-label` value.  

3 patterns for the link label can be set up:
- title value on `a`
- aria-label on `span role="img"`
- title value on `a` + aria-label on `span role="img"`

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-26304)
- [Github issue]()

## Steps to Test

1. Go to the Recent news & announcements section on /patterns/05-pages-organization/05-pages-organization.html
2. Inspect the teaser component for:
    - [Sample 1] no alt value:  no `aria-label` and `role="img"` on `span`
    - [Sample 2] no title value:  no `title` on `a`, `aria-label` and `role="img"` on `span`
    - [Sample 3] title + alt:  `title` on `a`, `aria-label` and `role="img"` on `span` (screen readers announce both)

3. Find no visual changes.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

### DIFF
--- a/packages/patternlab/styleguide/source/_patterns/02-molecules/press-teaser.twig
+++ b/packages/patternlab/styleguide/source/_patterns/02-molecules/press-teaser.twig
@@ -5,9 +5,7 @@
     <a
       class="ma__press-teaser__image"
       href="{{ pressTeaser.title.href }}"
-    {% if pressTeaser.title.text %}
       title="{{ pressTeaser.title.text }}">
-    {% endif %}
       {% set image = pressTeaser.image %}
       <span
       {% if image.alt %}
--- a/packages/patternlab/styleguide/source/_patterns/02-molecules/press-teaser.twig
+++ b/packages/patternlab/styleguide/source/_patterns/02-molecules/press-teaser.twig
@@ -2,14 +2,16 @@
 
 <section class="ma__press-teaser {{ imageClass }}">
   {% if pressTeaser.image %}
-    <a 
+    <a
       class="ma__press-teaser__image"
-      href="{{ pressTeaser.title.href }}" 
+      href="{{ pressTeaser.title.href }}"
       title="{{ pressTeaser.title.text }}">
       {% set image = pressTeaser.image %}
-      <span 
-        aria-label="{{ image.alt }}" 
+      <span
+      {% if image.alt %}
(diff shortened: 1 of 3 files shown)


packages/patternlab/styleguide/source/_patterns/02-molecules/press-teaser.twig
```diff
@@ -2,14 +2,18 @@
 
 <section class="ma__press-teaser {{ imageClass }}">
   {% if pressTeaser.image %}
-    <a 
+    <a
       class="ma__press-teaser__image"
-      href="{{ pressTeaser.title.href }}" 
+      href="{{ pressTeaser.title.href }}"
+    {% if pressTeaser.title.text %}
```
clairesunstudio: I think we don't need the check here because the title link text should be a required prop, instead of optional. Without the link text, this component is not useful.

packages/patternlab/styleguide/source/_patterns/05-pages/organization.json
```diff
@@ -1020,7 +1020,7 @@
                   "eyebrow": "News Article",
                   "title": {
                     "href": "#",
-                    "text": "Tree House Brweing Co. Expands Its Horizons",
+                    "text": "",
```
clairesunstudio:
```suggestion
                    "text": "",
```

Currently backstop is failing. I would recommend adding this text back and remove the condition so match the previous screenshots.

clairesunstudio (commented):
Otherwise this looks good!

clairesunstudio (approved):

PR #1792, opened 2023-05-30
react/DP-27961-fix-main-nav-mobile-keyboard-nav

clairesunstudio:
## Descriptions
This PR fixed the following 3 types of accessibility issues in mainnav:

1. All the markup errors detected by axeDev
2. First item in subnav was skipped on desktop using arrow keys
3. Arrow keys are not working at all in the mainnav mobile tray. 
    1. Left/right arrow keys do not work within the main nav - should navigate between the top level items
    2. Up/down arrow key do not work within the main nav - should open the subnav and move focus to the last/first subnav item
    3. If use Enter/space to open the subnav, using the down arrow key gets stuck on the first item. 

issue in budget repo: https://github.com/EOANF/budget/issues/558

## To test:
1. Pull down this branch and serve locally
5. Compare the before (prod) vs the after (local) for the [Header](https://mayflower.digital.mass.gov/react/iframe.html?id=organisms-header--header-example&viewMode=story) and [HeaderHamburger](https://mayflower.digital.mass.gov/react/iframe.html?id=organisms-headerhamburger--header-hamburger-example&viewMode=story) components

### 1. All the markup errors detected by axeDev
Before:
Screen Shot 2023-05-30 at 4 00 06 PM

After:
Screen Shot 2023-05-30 at 4 08 13 PM

### 2. First item in subnav was skipped on desktop using arrow keys
Before:
main-nav-keyboard-nav-before

After:
main-nav-keyboard-nav


ygannett (commented):
### Keyboard navigation:
| keyboard navigation | Horizontal | Hamburger |
|  :----   |  :----:   | :----: |
| right/left arrow keys to navigate top level items | ✅ | ✅ |
| up arrow key over the top level item to set focus on the last item in its submenu | ✅ | ✅ |
| ENTER/SPACE key to open its submenu on a top level item | ✅ | ✅ |

Hamburger menu only:
When a submenu closes with ESC, focus is set on `body.sb-show-main.show-menu`.  The closed submenu's parent menu item doesn't get focus.

### Markup:
`li.ma__main-nav__item` and `li.ma__main__hamburger-nav__item` have 2 types:
- role="menuitem" with `button` for its child element
  `<li role="menuitem"><button /></li>`
- role="none" with `a` for its child element, which has role="menuitem"
  `<li role="none"><a role="menuitem">MENU LABEL</a></li>`


Basically a `li` itself woulnd't be a menu item because:
- its functional role is:
  1. to be a container for a menu item to present the structure of the menu and 
  2. to define where the contained menu item positions itself in the whole menu structure.
  3. to indicate how many siblings items exist and the order of the item in the siblings, but not as a list item role (this sounds a bit odd, but even a list loses its original role as a list, screen readers announce how many items in it.  Here, we take advantage of this feature for screenreader users' convenience.)
- it's not an interactive element to takes you to somwehre or show submenu (not an actual menu item)

Once `ul` gets *menubar* or *menu* role, the set of `ul` is not a list anymore although it retains its structural presentation as a list.  Because of its new role, it seems natural to give its child `li` `role="menuitem"`, but this is not semantically correct because of the reasons described above.  That's why `li` in a list with `role="menubar"` or `role="menu"` gets `role="none"` to override its original role for list item.

There is, however, an exception.  When `li` has a button for its child element, the button needs to retain its button role for assistive technology, so its users would know it is a button.  In this case, their parent `li`s get `role="menuitem"` for the buttons.  
Note:  Screenreader support for multiple roles on one element is buggy.

So, the COVID 19 item in the top level menu is supposed to be:
Horizontal menu: 
```
<li role="none" class="ma__main-nav__item js-main-nav-toggle" tabindex="-1">
  <a role="menuitem" href="https://www.mass.gov/info-details/covid-19-updates-and-information" class="ma__main-nav__top-link cv-alternate-style" tabindex="0">COVID-19</a>
</li>
```
Hamburger menu:
```
<li role="none" class="ma__main__hamburger-nav__item js-main-nav-hamburger-toggle" tabindex="-1">
  <a role="menuitem" href="https://www.mass.gov/info-details/covid-19-updates-and-information" class="ma__main__hamburger-nav__top-link cv-alternate-style" tabindex="0">COVID-19</a>
</li>
```
and submenu items are supposed to be:
Horizontal menu:
```
<li role="none" class="ma__main-nav__subitem">
  <a role="menuitem" href="https://mass.gov" class="ma__main-nav__link">Health &amp; Social Services</a>
</li>
```
Also, `aria-expanded="false"` is not necessary with `a.ma__main-nav__link`

Hamburger menu:
```
<li role="none" class="ma__main__hamburger-nav__subitem js-main-nav-hamburger__subitem">
  <a role="menuitem" href="https://mass.gov" class="ma__main__hamburger-nav__link js-main-nav-hamburger__link">Health &amp; Social Services</a>
</li>
```


To match Patternlab, you might remove the last link item in each dropdown.  If you see any benefits with the item for react users, you might keep it.   
Horizontal menu:
```
<li role="menuitem" class="ma__main-nav__subitem--main">
  <a aria-expanded="false" href="#" class="ma__main-nav__link">
    <svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 20 21"><path d="M16.721 9.277L19.423 12l-2.702 2.724-5.403 5.445-2.701-2.722 3.498-3.528H2.443v-3.851h9.66L8.616 6.553l2.702-2.723zM3.882.322v13.672H.105V.334z"></path></svg>
    <span><span class="visually-hidden">See all topics under </span>Living</span>
  </a>
</li>
```
Hamburger menu:
```
<li role="menuitem" class="ma__main__hamburger-nav__subitem--main js-main-nav-hamburger__subitem">
  <a href="#" class="ma__main__hamburger-nav__link js-main-nav-hamburger__link"><svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 20 21"><path d="M16.721 9.277L19.423 12l-2.702 2.724-5.403 5.445-2.701-2.722 3.498-3.528H2.443v-3.851h9.66L8.616 6.553l2.702-2.723zM3.882.322v13.672H.105V.334z"></path></svg>
    <span><span class="visually-hidden">See all topics under </span>Living</span>
  </a>
</li>
```

clairesunstudio:
Thanks @ygannett for the thorough review!  
- I will look into the escape key focus issue. 
- Good catch on the aria-expanded on `<a>`s, I will remove those. 
- I will remove the last item too if it doesn't impact any React implementations

Re: roles in markup:
The reason why I added the `menuitem` role to li is that it's failing the axeDev accessibility checks. 
In the accessibility review, Rob pointed out that they are using axeDev to check for issues. These are flagged as issues in the budget site: https://github.com/EOANF/budget/issues/558#issuecomment-1569036766

Screen Shot 2023-05-30 at 4 00 06 PM


These are also flagged as issues on mass.gov 
Screen Shot 2023-06-06 at 2 41 34 PM

clairesunstudio:
I'm also linking to this comment from Chris, as this is related to the above comments https://github.com/massgov/mayflower/pull/1795#discussion_r1220153693

clairesunstudio:
@ygannett mentioned that our current roles in markup follows the W3C examples 
https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/
https://codepen.io/pen, where `<li>` has the role of none, and anchor link has the role of menuitem

packages/react/src/components/molecules/HeaderNav/main-nav.js
```diff
@@ -223,18 +223,17 @@ export const HeaderNavItem = React.memo(({
...
-          role="menuitem"
           href={href}
           className="ma__main-nav__top-link cv-alternate-style"
           tabIndex="0"
         >
           {text}
         </a>
       ) : (
-        <button ref={buttonRef} type="button" role="menuitem" id={`button${index}`} className="ma__main-nav__top-link" aria-haspopup="true" tabIndex="0" aria-expanded={buttonExpanded}>
+        <button ref={buttonRef} type="button" id={`button${index}`} className="ma__main-nav__top-link" aria-haspopup="true" tabIndex="0" aria-expanded={buttonExpanded}>
```
clairesunstudio: @ygannett but if we keep the menuitem role inside of `<li>` but not overriding the role of button, then `<button>` at the same level as `<a>` will not receive a menuitem role. I feel like that's not a consistent way of applying the roles. To that point, I think it makes sense to apply the menuitem role at the `<li>` level. Thoughts?

clairesunstudio:
Discussed with @ygannett, the final decision is to move the `menuitem` role to `<li>` to make it consistent for buttons and anchor links. We will confirm in screenreader that the impact is acceptable

ygannett:
Thank you for your work responding to my comment.  Let's finalize the role matter with the feedback you get from today's meeting.

clairesunstudio:
Rob didn't point out anything regarding the behavior of the submenu lists in JAWS. But I did see in action that in JAWS, moving the role `menuitem` onto `<li>` prevents the number being announced. I adopted a hybrid solution - keep the top level nav li as `menuitem`, regardless of whether its child is a button or a link, this way no need to override the children roles; and in the submenu, add the `menuitem` role to the link instead of the li, to persist the `<li>` functionality.

clairesunstudio:
The react package build is breaking SSR.. 
Screen Shot 2023-06-08 at 11 40 35 AM

ygannett (approved):
It looks good.  The navigation is functioning as we discussed.  Thank you for all your work.

bronzehedwick (approved):

PR #1802, opened 2023-06-09
adding useDocument hook to check for document mounting

clairesunstudio:


clairesunstudio:
The build package doesn't trigger an SSR error, but it's breaking the header interactions, since the initial state of the doc is "null", it requires a deep check. I'm still not sure why previously the document references work without checks...
@bronzehedwick

clairesunstudio:
final fixes is manually applied

PR #1834, opened 2023-08-30
DP-29416 Change twigger .svg to X and add Threads .svg

weitzman:
## Related Issue / Ticket

https://massgov.atlassian.net/browse/DP-29416


weitzman:
See #1835

PR #1857, opened 2023-12-12
DP-28505 Replace the custom responsive table scrollbar with native one

ygannett:

### Table wrapper
- Replace the existing scroll bar and button control with the native scroll bar `overflow-x: auto;`
- Set up the responsive table wrapper focusable for keyboard navigation with arrow keys
     - The wrapper is focusable only when it has overflow content to enable scroll control by keyboard
- Add a label to the responsive table container syncing with the table `caption` via `aria-labelledby`
    - Add a separate instruction for screen reader users `#sr-instructions` via `aria-labelledby`

### Table caption
- Add `caption` to the template to indicate the scroll bar to users
- Style `caption` as a table title
- Hide the scroll bar indication message when the table wrapper has no scroll bar.
- When no table title with no overflow content, `caption` is hidden, not an empty element.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-28505)
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


ygannett:
Backstop failure is not related to the changes in this PR:  https://output.circle-artifacts.com/output/job/d7a0c3b6-2431-4488-8431-e0ef11e85e94/artifacts/0/backstop/reports/html/index.html

packages/assets/scss/01-atoms/_table-responsive.scss
```diff
@@ -3,24 +3,36 @@
...
-    overflow: -moz-scrollbars-none;  // Firefox
 
-    &::-webkit-scrollbar {
-      display: none;  // Safari and Chrome
-    }
+  overflow-x: auto;
+
+  @media ($bp-medium-max) {
+
+    overflow-y: clip;
```
clairesunstudio:
```suggestion
    overflow-y: hidden;
```

No difference than hidden if no [overflow-clip-margin](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin) is set. Just want to make sure that this is the desired behavior

clairesunstudio:
@ygannett , I'm still seeing the legacy responsive table in Mayflower Patternlab, can you please update them to use the up-to-date responsive table styles? Please let me know if I'm missing something. 

The patterns I checked:
https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-28505_scrollbar-aria/?p=viewall-atoms-table
https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-28505_scrollbar-aria/?p=pages-information-details-responsive-table
https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-28505_scrollbar-aria/?p=pages-information-details-toc-precontent-responsive-tables

ygannett:
@clairesunstudio I wonder that's cache.  I followed your links and saw the new scroll bar.  Here is one sample.
Screenshot 2024-01-24 at 9 31 03 AM

clairesunstudio:
@ygannett, I see the scrollbar but on mobile the table still inverts:
Screen Shot 2024-01-24 at 11 40 16 AM

For the table patterns https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-28505_scrollbar-aria/?p=viewall-atoms-table, I still see the old versions. can you update those? I think we can clean up all the old JS and assets if they are not being used on mass.gov

packages/assets/scss/01-atoms/_table.scss
```diff
@@ -5,6 +5,26 @@
...
+    &.hide {
+      display: none;
+    }
+
+    &__content {
+      display: inline-block;
+      width: 100%;
+    }
+
+    // TODO:  ADD STYLE FOR ONE COL LAYOUT
```
clairesunstudio: what do you mean by one col layout?

packages/assets/scss/01-atoms/_table.scss
```diff
@@ -5,6 +5,26 @@
...
 
+  .ma__table__caption,
+  & td > table caption {
+    color: $c-font-heading;
+    font-weight: $fonts-bold;
+    line-height: 1.3;
+    text-align: left;
+    margin-bottom: 0.8em;
+
+    &.hide {
```
clairesunstudio, line 16: This seems to be reverse of what's being set for the scroll info above, see .ma__table__caption__scroll-info
Just want to make sure this is needed

packages/patternlab/styleguide/source/assets/js/modules/responsiveTable.js
```diff
@@ -213,141 +207,32 @@ export default (function (window, document, $) {
...
   function handleTableScroll(e) {
     if (skip === 0) {
       const t = e.target;
       // No matter the element, the percentage as a decimal should be the amount of pixels scrolled
       // divided by the max amount that can be scrolled which is the scroll width minus the width.
       const scrollPercent = t.scrollLeft / (t.scrollWidth - t.offsetWidth);
-      ["ma__table--responsive__wrapper", "ma__scroll-indicator", "sticky-thead-wrapper"].map(scrollable => {
+      [
+        "ma__table--responsive__wrapper",
+        "ma__scroll-indicator",
```
clairesunstudio:
```suggestion
        "ma__scroll-indicator",
```

Do we still use this class?

packages/patternlab/styleguide/source/assets/js/modules/responsiveTable.js
```diff
@@ -213,141 +207,32 @@ export default (function (window, document, $) {
...
+        "sticky-thead-wrapper",
+      ].map((scrollable) => {
         if (t.className !== scrollable) {
           const elements = this.getElementsByClassName(scrollable);
           if (elements.length > 0) {
             let el = elements[0];
             // If we are scrolling the scrollbar or the target is the scrollbar, inverse the scroll.
-            if (scrollable === "ma__scroll-indicator" || t.className === "ma__scroll-indicator") {
+            if (
+              scrollable === "ma__scroll-indicator" ||
```
clairesunstudio: are these still relevant?

packages/patternlab/styleguide/source/_patterns/01-atoms/10-table/table-responsive.twig
```diff
@@ -6,21 +10,12 @@
...
-    <button class="ma__table__horizontal-nav__right" aria-controls="{{tableID}}">
-      <span class="visually-hidden">Scroll right</span>
-    </button>
-  </nav>
-  <div class="ma__table--responsive__wrapper {{ mergedCells ? 'ma__table--merged-cells' : ''}}" id="{{tableID}}" tabindex="0" role="region" aria-orientation="horizontal" aria-label="{{table.description}}">
+  <div class="ma__table--responsive__wrapper {{ mergedCells ? 'ma__table--merged-cells' : ''}}" id="{{tableID}}" role="group" tabindex="-1">
     <table class="ma__table {{ tableWideClass }}">
+      <caption id="tbl-{{tableID}}" class="ma__table__caption">
+        {% if table.description %}<span class="ma__table__caption__content">{{ table.description }}</span>{% endif %}
+        <span class="ma__table__caption__scroll-info"> (Table in a horizontal scrolling container)</span>
```
clairesunstudio:
```suggestion
        <span class="ma__table__caption__scroll-info"> Note: Table has hidden columns, scroll horizontally to see more.</span>

```

I think we should use more user-friendly language. Maybe something like:
"Table has hidden columns, scroll horizontally to see more."

clairesunstudio:
Ideally, the description and scrolling instruction should not scroll with the table. But since they are in the caption, they are currently inside of the table. And whether we determine to scroll or not it's based on the table width and the table wrapper width.

I think we should probably check the tbody width and/or the thead width instead of the whole table. So that: 
1. It will prevent the description and instruction from scrolling 
2. In case the description is long, and the table itself actually don't have overflow, I think currently it will trigger the overflow instruction

clairesunstudio:
Please resolve the conflicts on this PR and the openmass PR. Thank you!

packages/assets/scss/01-atoms/_table.scss
ygannett: This is what I forgot to remove. I initially misunderstood the task as implementing the Mayflower in openmass.  It was a note for myself.  It's removed.

packages/assets/scss/01-atoms/_table.scss
ygannett: The class is used by JS to hide `caption` when the table has no overflow content and no author input `caption`.
L. 133 responsiveTable.js

packages/patternlab/styleguide/source/assets/js/modules/responsiveTable.js
ygannett: Thank you for mentioning that.  This is not in use and the class and related lines are cleaned up.

packages/patternlab/styleguide/source/assets/js/modules/responsiveTable.js
ygannett: Not anymore.  The section has been cleaned up.

ygannett:
I see your point.  The reason I put the scrollbar info in `caption` is that it'd be the most reasonable place as the content flow for the info for screen reader users since they know they are in the table and haven't reached the table content although that is not the best presentation for sighted users.  Another thing to consider is which is better in the limited space, especially in mobile: 1. see the table title and the scroll info. all the time by cutting off the space to show more table rows at once, 2. show the table content as much as possible by hiding the table title and the scroll info.  It depends what we put more priority than the other, I guess.  

I played around with `caption`, `table`, and `tbody`.  `caption` always stays within the `table`/`tbody` or `thead` width.  The third screenshot shows that the longer caption than the table content doesn't go over the table content (`tbody` and `thead`) width.  Instead, it gets wrapped within the table content width.  The current implementation wraps the longer caption content than the table width to fit the table width.

Table width: 100%
Caption width: 30%
Screenshot 2024-03-12 at 2 46 46 PM

Table width: 100%
Caption width: 100%
Screenshot 2024-03-12 at 2 47 19 PM

Table width: 80%
Caption width: --
Screenshot 2024-03-12 at 2 48 53 PM

ygannett:
The backstop failures are not related to this PR's change.

packages/assets/scss/01-atoms/_table.scss
clairesunstudio: ah I see, so the scroll info is controlled by the .show class and the title and scroll info are both controlled by .hide

Thanks for clarifying

clairesunstudio:
Thanks for the clear demo! So the caption length doesn't determine the table width, only the table data does. 

To your point here: 
> Another thing to consider is which is better in the limited space, especially in mobile: 1. see the table title and the scroll info. all the time by cutting off the space to show more table rows at once, 2. show the table content as much as possible by hiding the table title and the scroll info. It depends what we put more priority than the other, I guess.

I just want to clarify that I meant horizontal scrolling not vertical scrolling, just so we are on the same page. I was thinking we could treat the title and scroll info in a normal content flow outside the table so that they don't scroll horizontally with the table.  Just wanted to point out that I have not seen table title and description purposely layout like that. But if you think it's the right place for that info and the most accessible content flow,  I'm ok with that.

clairesunstudio (approved):

arthurbaghdas (approved):

PR #1858, opened 2023-12-14
DP-28856 - Update React footer

tom-fleming:
Changed:
  - project: React
    component: Footer
    description: Updated Footer to match PatternLab version. (#1858)
    issue: DP-28856
    impact: Major

Added new Footer component (Mass.gov production) to Mayflower React, and renamed the old Footer to FooterWithColumn.


tom-fleming:
@clairesunstudio Mind taking a look at what I have so far? I can easily move this to just being an update of the current Footer component if that's desired. But, did this as a new component to match Patternlab.

Just a note too that this implementation does not include showNavHeading, socialLinks, or backToTopButton since these are not present in the Patternlab component. Do showNavHeading and backToTopButton need to be present though?

clairesunstudio:
@tom-fleming, as long as it matches the patternlab version - I don't think it has `showNavHeading` and `backToTopButton`, but it does include the feedback button. 

Can you please update the component name to match patternlab? I believe the new footer is called Footer and the old one is called FooterWithColumns. Otherwise looks good! I will wait for this PR to do a release. Thanks!

tom-fleming:
@clairesunstudio Ready for review again. I renamed the old footer component to FooterWithColumns and added the feedback button. Just a note that unlike PatternLab, the markup for the feedback button lives inside `<footer>` rather than outside it. If we need to determine a way to include it outside of `<footer>`, I think we'll need to give it another wrapper tag so that React doesn't error on it.

clairesunstudio:
Can you please fix the circle failures? I'm not able to build the storybook locally.

clairesunstudio (approved):

clairesunstudio:
### Breaking Change
We updated Footer in Mayflower React to match what's currently in Patternlab Mass.gov production. If you are using the previous version of Footer, please import component `FooterWithColumns` instead. We also updated the css file names to match the footer component naming, please refer to the mayflower.digital.mass.gov/core documentation for more information.

PR #1892, opened 2024-03-10
DP-32068: Log in button style improvement

arthurbaghdas:

## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


packages/assets/scss/03-organisms/_footer-with-columns.scss
```diff
@@ -10,7 +10,7 @@
 
   &__back2top {
 
-    @include ma-button-size("small");
+    @include ma-button-size("xsmall");
     font-size: $fonts-xsmall;
     letter-spacing: $letter-spacing-large;
```
clairesunstudio, line 15:
```suggestion
    letter-spacing: $letter-spacing-small;

```

packages/assets/scss/03-organisms/_page-header.scss
```diff
@@ -214,6 +214,10 @@ $page-header-widget-width: 350px;
     text-shadow: none;
     margin-top: 0.5rem;
     margin-right: 0.5rem;
+
+    a.ma__button {
```
clairesunstudio: Instead of overriding it here, can we add this as an option to button as a component? For all buttons, if the label text is longer than 2 words, we should not use ALL CAPS. 

This is what I suggest we do:
We move the "text-transform: uppercase;" on ma__button into a modifier class ".ma__button--uppercase". Add a casing prop to buttons, and by default set uppercase to true. Add here we set it to false. 

Also, we should set "$letter-spacing-large;" with the uppercase option, and for regular casing, we should set letter spacing to "$letter-spacing-small;"

Screen Shot 2024-03-11 at 3 32 25 PM

arthurbaghdas: Changed the implementation as suggested above.

changelogs/DP-32068.yml
```diff
@@ -0,0 +1,6 @@
+Changed:
+  - project: Patternlab
```
clairesunstudio, line 2: Can you please update changelog to describe the changes to the button component?

packages/assets/scss/00-base/mixins/_ma-button.scss
```diff
@@ -108,9 +108,9 @@
       height: $fonts-small;
     }
   }
-  // small
+  // xsmall
 
-  @if ($size == "small") {
+  @if ($size == "xsmall") {
```
clairesunstudio: I think we need to add a ma__button--xsmall class in _button.scss
```
  &--xsmall {

    @include ma-button-size("xsmall");
  }
```

Like the feedback button needs to use the xsmall class to match production
http://localhost:3000/?p=organisms-mass-feedback
https://mayflower.digital.mass.gov/patternlab/b/develop/?p=organisms-mass-feedback

packages/assets/scss/02-molecules/_event-filters.scss
```diff
@@ -33,7 +33,7 @@
 
   &__submit {
 
-    @include ma-button-size('small');
+    @include ma-button-size('xsmall');
```
clairesunstudio: I don't see this style being applied http://localhost:3000/?p=molecules-event-filters

packages/assets/scss/03-organisms/_footer-with-columns.scss
```diff
@@ -10,7 +10,7 @@
 
   &__back2top {
 
-    @include ma-button-size("small");
+    @include ma-button-size("xsmall");
     font-size: $fonts-xsmall;
```
clairesunstudio: this is repetitive

packages/patternlab/styleguide/source/_patterns/03-organisms/feedback/mass-feedback.twig
```diff
@@ -66,7 +66,7 @@
     </div>
 
     <div class="ma_feedback-fieldset submitButtonWrapper">
-      <input class="submitButton ma__button ma__button--small" type="submit" value="Send Feedback" />
+      <input class="submitButton ma__button ma__button--uppercase ma__button--small" type="submit" value="Send Feedback" />
```
clairesunstudio, line 69:
```suggestion
      <input class="submitButton ma__button ma__button--uppercase ma__button--xsmall" type="submit" value="Send Feedback" />
```

To match what's in production, we will need to add a classname (see comment earlier)

clairesunstudio:
Changes look good! Please see my comments for some minor suggestions. Also, I think these changes will make the Mayflower React backstop fail. Can you run `rush backstop:react` and fix those regressions?

After that's done, please update the openmass PR and run the backstop tests in openmass. Thank you!

changelogs/DP-32068.yml
arthurbaghdas: Updated

packages/assets/scss/02-molecules/_event-filters.scss
arthurbaghdas: Because the button was missing the ma__events-filters--submit class

clairesunstudio (approved):

PR #1912, opened 2024-05-16
Scrollbar design improvements

clairesunstudio:
## Problem Overview

This needs to go back to design for re-evaluation for a more complete solution. 

In general, I agree with replacing the custom responsive table scrollbars with native browser scrollbars as the solution - it will take a lot effort to make the keyboard navigation work like native for keyboard and all assistive technology. It adds complexity in code and will be difficult to maintain accessibility for all assistive technology. The existing JS approach is only working for tables, which also creates inconsistencies in user experience with other possible overflow content - such as iframes.

### 1. There's a way to style the default scrollbar for most devices + browsers using the `::-webkit-scrollbar` selectors ([this PR shows how it's done](http://localhost:3000/?p=pages-information-details-responsive-table)).
This customization forces browsers to render the scrollbar (for all overflow content) before there's any interactions. This works for most devices and browsers except for  IOS Safari and Firefox.
Screen Shot 2024-05-17 at 1 56 36 PM

### 2. Instead of displaying a text note above the table, which can be easily overlooked by visual users, I suggest we add a gradient overlay and an scroll icon on the right side of the table to show there's more content. 
Figma link: https://www.figma.com/design/ustqFiDpEvSOfK0dz2smLd/Overflow-Cont
This visual cue should display by default, but should go away as soon as there's scrolling happening within the table. 
All other browsers (2)

Providing this additional visual cue by default is necessary, because the CSS scrollbar customization (in step1) doesn't work with Safari iOS > 13 and Firefox. So this will be the fallback visual cue for those users. 
> Apple made the conscious decision to hide scrollbars in iOS at the system level until there's an interaction. Mac OS (including Safari) and Android quickly followed that decision. 
https://caniuse.com/?search=%3A%3A-webkit-scrollbar-track
Screen Shot 2024-05-16 at 2 05 31 PM (2)
Screen Shot 2024-05-16 at 2 35 37 PM
Currently there's no way to customize the scrollbar for IOS 13 and above in the source code. Only users can set the scrollbar behavior in system settings. This was verified in Browserstack.
https://forums.developer.apple.com/forums/thread/670065

### 1.Sticky table header becomes sticky too quickly 
I think this is because the caption and note (this will be replaced in the newly recommended design) throws of the position top measured for the sticky position.
Screen Shot 2024-05-16 at 3 37 51 PM
Prod:
https://www.mass.gov/guides/absence-and-leave-policies
After:
https://pr2322-kn9cwaccdicqqjfgqlxsckcg0vlipz0i.tugboatqa.com/guides/absence-and-leave-policies
https://pr2322-kn9cwaccdicqqjfgqlxsckcg0vlipz0i.tugboatqa.com/info-details/qag-info-details-table-samples

### 2.Sticky elements should have visible boundaries. (Need a separate ticket)
In this case, two separate sticky elements are blended in together http://localhost:3000/patterns/05-pages-information-details-toc-precontent-responsive-tables/05-pages-information-details-toc-precontent-responsive-tables.html

Screen Shot 2024-05-16 at 3 07 59 PM
Screen Shot 2024-05-16 at 3 11 20 PM


packages/patternlab/styleguide/source/_patterns/01-atoms/10-table/table-responsive.twig
```diff
@@ -14,9 +14,10 @@
     <table class="ma__table {{ tableWideClass }}">
       <caption id="tbl-{{tableID}}" class="ma__table__caption">
         {% if table.description %}<span class="ma__table__caption__content">{{ table.description }}</span>{% endif %}
-        <span class="ma__table__caption__scroll-info"> Note: Table has hidden columns, scroll horizontally to see more.</span>
-
       </caption>
+      <div class="ma__table__caption__scroll-info">
```
clairesunstudio:
```suggestion
      <div class="ma__table__caption__scroll-info" aria-hidden="true">

```

arthurbaghdas (approved):

PR #1928, opened 2024-08-19
DP-34425: allow raw HTML in alertMsg for mass-feedback.twig template.

dstorozhuk:

## Related Issue / Ticket

- [DP-34425](https://massgov.atlassian.net/browse/DP-34425)
- https://github.com/massgov/openmass/pull/2637

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

#### Today I learned...

[DP-34425]: https://massgov.atlassian.net/browse/DP-34425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/patternlab/styleguide/source/_patterns/03-organisms/feedback/mass-feedback.twig
+++ b/packages/patternlab/styleguide/source/_patterns/03-organisms/feedback/mass-feedback.twig
@@ -46,7 +46,7 @@
           {% set textarea = affirmativeTextarea %}
           {% include "@atoms/03-forms/textarea.twig" %}
           {% if alertMsg %}
-            <div class="ma__alert-msg" role="region" aria-live="polite" id="{{ affirmativeTextarea.errorIds }}">{{ alertMsg }}</div>
+            <div class="ma__alert-msg" role="region" aria-live="polite" id="{{ affirmativeTextarea.errorIds }}">{{ alertMsg|raw }}</div>
           {% endif %}
           {% if warnMsg %}
             <div class="ma__warn-msg{{showWarnMsg ? ' has-error-default' : ''}}">{{ warnMsg|raw }}</div>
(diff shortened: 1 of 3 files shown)


packages/patternlab/styleguide/source/_patterns/03-organisms/feedback/mass-feedback.twig
```diff
@@ -46,7 +46,7 @@
           {% set textarea = affirmativeTextarea %}
           {% include "@atoms/03-forms/textarea.twig" %}
           {% if alertMsg %}
-            <div class="ma__alert-msg" role="region" aria-live="polite" id="{{ affirmativeTextarea.errorIds }}">{{ alertMsg }}</div>
+            <div class="ma__alert-msg" role="region" aria-live="polite" id="{{ affirmativeTextarea.errorIds }}">{{ alertMsg|raw }}</div>
```
arthurbaghdas, line 49: Why do we need raw filter here? As i know it is a bad practice to use this filter, because it may expose XSS security issues. We need an alternative solution for this.

dstorozhuk: hi @arthurbaghdas . Thanks for the review.
The entire template variables are outputted with `|raw` filter. 
feedba… 2024-08-20 02-22-23

We are not using this template (or at least `alertMsg`) with data that comes from the database or user input, the data for `alertMsg` hardcoded in another Drupal template.
I think it safe to use it in that way. 

If you think it not safe:
1. We need to refactor the entire template, and remove the `|raw` filter from all variables, not just that 2 places that I added.
2. Please suggest the "right" way how you think we can pass the value for `alertMsg` variable from `node--org-page--feedback.html.twig` in openmass repo.
The content of the variable I need to send to mayflower component is next:
`"Please remove any contact information or personal data from your feedback. <strong>You will NOT get a response<strong>."`

arthurbaghdas: We are sending it from Drupal, where we are sure we can control it. The Mayflower can be used by other applications and I don't think we should have raw filter in Mayflower.

dstorozhuk: @arthurbaghdas  what do you propose to do in this particular case and what do propose to do with entire temple since it has additional 5 raw filters which was there before my change? 

Thanks for your suggestions in advance.

arthurbaghdas: You are right, I'll create a ticket and we can refactor later.

arthurbaghdas (approved):

PR #1930, opened 2024-08-23
DP-34424: Add floats clearing to headings in rich-text styles

dstorozhuk:
## Description
This commit adds clear both to all heading elements in the rich-text SCSS file to prevent floating issues with left or right floating figures. This ensures that headings are not misaligned due to preceding content.

## Related Issue / Ticket

- https://massgov.atlassian.net/browse/DP-34424
- [[Github issue]()](https://github.com/massgov/openmass/pull/2641)


packages/assets/scss/03-organisms/_rich-text.scss
```diff
@@ -73,6 +73,12 @@ $heading-indent: $gutter;
     }
   }
 
+  // clear float for all headings after left of right floating figures
+
+  h1, h2, h3, h4, h5, h6 {
+    clear: both;
```
clairesunstudio:
```suggestion
    clear: left;
```

considering the visual regression, I think we should just clear the left float and keep the right as is

packages/assets/scss/03-organisms/_rich-text.scss
```diff
@@ -73,6 +73,12 @@ $heading-indent: $gutter;
     }
   }
 
+  // clear float for all headings after left of right floating figures
```
clairesunstudio:
```suggestion
  // clear float for all headings after left floating elements
```

arthurbaghdas (approved):

PR #1935, opened 2024-09-24
DP-35080: Usability improvements to rich text tables related to horiz…

arthurbaghdas:
…ontal scrolling

## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


packages/patternlab/styleguide/source/assets/js/modules/responsiveTable.js
```diff
@@ -109,32 +107,15 @@ export default (function (window, document, $) {
     if (rt.horizontalScrollable) {
       $tableWrapper.attr("tabindex", "0");
       $tableWrapper.attr("aria-labelledby", "sr-instructions");
-      $scrollInfo.addClass("show");
+      $stickyHeader.removeAttr("aria-labelledby");
```
clairesunstudio:
```suggestion
      $stickyHeader.removeAttr("aria-labelledby");
```

These should apply everywhere not just when the table has horizontal scrolling - please see my latest comment in the ticket. We no longer need `sr-instructions` and this block of conditional code  and these new attributes should be added directly in the twig template

packages/assets/scss/01-atoms/_table-responsive.scss
```diff
@@ -77,6 +35,25 @@
 
 .ma__table--responsive__wrapper {
```
clairesunstudio:
```suggestion
// Pure CSS scrolling shadows with background-attachment: local. Code taken from https://codepen.io/aardrian/pen/MWeRJWd
```

clairesunstudio: add a reference to the source

packages/patternlab/styleguide/source/assets/js/modules/responsiveTable.js
clairesunstudio: @arthurbaghdas I think these should not be wrapped in conditionals, instead they should be directly set in the html `"<div class='sticky-thead'><div class='sticky-thead-wrapper'><table></table></div></div>"` on line 49. 
For example, the second table on this [page](https://pr2694-w1avdsfe4brh4hyl692txjip3yojfiza.tugboatqa.com//info-details/overview-of-the-pioneer-valley-transit-authority) (with sticky header but without horizontal scrolling) is not getting the aria-hidden and tabindex=-1. These should apply to all sticky table headers as they are initiated

packages/assets/scss/01-atoms/_table-responsive.scss
```diff
@@ -77,6 +35,25 @@
 
 .ma__table--responsive__wrapper {
 
+  background: linear-gradient(to right, #fff 30%, rgba(255, 255, 255, 0)),
+  linear-gradient(to right, rgba(255, 255, 255, 0), #fff 70%) 0 100%,
+  radial-gradient(
```
clairesunstudio: I noticed an issue with the tapered gradients - https://pr2694-w1avdsfe4brh4hyl692txjip3yojfiza.tugboatqa.com//info-details/overview-of-the-pioneer-valley-transit-authority This is a super long scrollable table - the gradient is not in view immediately and requires several scrolls to be visible. We should remove the tapers from the background settings

clairesunstudio: Screenshot 2024-10-07 at 11 56 11 AM

Gradient is not visible due to the tapering on both top and bottom

clairesunstudio (approved):

clairesunstudio:
Please update the patternlab test - looks good!

PR #1938, opened 2024-11-20
DP-35365: Fix image styles in rich text with text wrapping

arthurbaghdas:

## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*


packages/assets/scss/01-atoms/_figure.scss
```diff
@@ -24,6 +24,12 @@ $figures-gutter: 20px;
     margin: 10px $figures-gutter 10px $figures-gutter;
   }
 
+  @media ($bp-x-small-max) {
```
clairesunstudio: Instead of adding a new media query, I wonder if there's any reasons of needing the previous one `@media ($bp-x-small-min)` to add the margins - the only thing I can think of is when the figure expands the full width on mobile and we want to remove the left-right margins.

clairesunstudio:
It looks like the images on mobile got smaller in the [backstop report](https://output.circle-artifacts.com/output/job/2254b154-7897-4570-b3fd-678cfb221e62/artifacts/0/backstop/reports/html/index.html), due to the unwanted margins. It could be related to the previous comment I had https://github.com/massgov/mayflower/pull/1938#discussion_r1876236735

clairesunstudio (approved):

PR #1958, opened 2025-03-14
DP-38011: Create Vertical Cards Paragraph Type

hotsauce:
## Description
This PR adds two new Paragraph types - `Vertical Card Group` and `Vertical Card` to the site, for use on the `Services` Content type. These are based on designs within the [Figma](https://www.figma.com/design/OfnYu3nsvlDPEjdKe5LuEc/EOHLC%3A-Finalized-Designs?node-id=4301-4707&m=dev), which serves as a Variant on the `Card` style [in Mayflower](https://mayflower.digital.mass.gov/patternlab/?p=molecules-card)

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/issues/DP-38011)
- [Github issue](https://github.com/massgov/openmass/pull/2877)

## Steps to Test

1. Visit the deployed site and login as an admin
2. Create a new `Services` piece of content
3. Add a new `Service Section` and within it, add a `Vertical Card Group`
4. Within the `Vertical Card Group`, add at least 3 `Vertical Card`s, filling in various fields other than what is required.
5. Modify the style of the `Service Section` to verify that the `Vertical Card` appears as expected.

## Screenshots

Screenshot on 2025-03-12 at 08-42-16
Screenshot on 2025-03-12 at 08-42-31

## Additional Notes:

Anything else to add?

* Card

#### @TODO

* Determine name of "Vertical Cards"


packages/patternlab/styleguide/source/_patterns/02-molecules/vertical-card.twig
```diff
@@ -0,0 +1,39 @@
+{% set thumbnail = card.thumbnail %}
```
clairesunstudio, line 1: Can you please make the vertical-card a variant of the card component instead of as a new component? I compared the changes and it seems like the main change is that the anchor tag goes one level outside of the wrapper div. Is there a specific reason for that? 

The markups seem to be mostly the same between card and vertical-card, instead of making a copy, the "vertical" or "stacked" version will just be a boolean property of the card component. If there's a reason for the anchor link to be the outmost wrapper, then I feel like we should keep that consistent with the existing card variant. 

To add an example in patternlab, you would just add a `card~stacked.json` with the `stacked:true` in the data object.

Screenshot 2025-03-14 at 4 47 16 PM

hotsauce: @clairesunstudio 
Thank you, I will do that. The biggest changes were the full link (which I can adjust) and adding the eyebrow field, but that's a minor thing to modify.

The naming was still TBD when I was working on this so I figured there would be some adjustments and I can see how the variant naming works better. I'll probably keep this PR open since most of the work (on the Mayflower side) is done, and just rename it to match the new DP-38060.

clairesunstudio: So just to clarify - we don't want to add another twig file, it should be sharing the same `card.twig` with the existing variant - so the full surface link hover state will be beneficial to existing card variant and eyebrow should be conditionally rendered when the data is passed, in the main card.twig. Let me know if you want to discuss on Monday

hotsauce: @clairesunstudio 
Yes, that's the approach I was going to take (looking at the "Card as Secondary" as a semi-reference) so I think we are good. I will ping you if I have any questions and thank you!

PR #1960, opened 2025-03-19
DP-38060: Add image option to card group

hotsauce:

## Description

This adds a variant to the `Card` component to allow for the cards to be displayed vertically. It also adds an optional eyebrow component, to display above the Card title. The entire card is clickable and the hover state for Cards has been updated to underline the title on hover/focus.

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/jira/software/c/projects/DP/issues/DP-38060)
- [Github issue](https://github.com/massgov/openmass/pull/2884)

## Steps to Test

1. Visit the deployed site and login as an admin
2. Create a new `Services` piece of content
3. Add a new `Service Section` and within it, add a `Card Group`
4. Within the `Card Group`, check the `Use image cards with single call to action` box. An image field, eyebrow field and long plain text box should appear and the WYSIWYG content box should disappear. 
5. Add at least 3 `Card`s, filling in various fields other than what is required.
6. Verify that if `Use image cards with single call to action` is checked, the image and link field become required before submitting/saving.
7. Modify the style of the `Service Section` to verify that the vertical version of `Card` appears as expected.

## Screenshots

Screenshot on 2025-03-12 at 08-42-16
Screenshot on 2025-03-12 at 08-42-31

## Additional Notes:

Anything else to add?

* Card


packages/assets/scss/03-organisms/_sections-three-up.scss
```diff
@@ -107,11 +107,16 @@ $bp-section-3up-large-max: "max-width:950px";
     }
   }
 
-  .ma__vertical-card-group--wrapper {
-
-    .ma__card.ma__card--vertical-card {
-      background-color: $c-white;
+  .ma__card__wrapper {
+    &:hover {
+      border-block-end: 2px solid $c-overcast-gray;
```
clairesunstudio: Please use the semantic colors in colors.scss instead of the primitive color tokens when possible, in this case, `$c-bd` instead of `$c-overcast-gray`

packages/patternlab/styleguide/source/_patterns/02-molecules/card.twig
```diff
@@ -5,6 +5,7 @@
 {% set imgClass = card.usage ? card.usage : card.id %}
 {% set detailsClass = card.usage ? " ma__card__details--" ~ card.usage : "" %}
 
+<a class="ma__card__wrapper" href="{{ card.link }}">
```
clairesunstudio: I think moving the wrapper out broke the primary card, can you please update the SCSS to fix this issue? The image should be side-by-side with the next not stacked https://mayflower.digital.mass.gov/patternlab/b/patternlab/DP-38060_add_image_option_card_group/?p=organisms-campaign-feature

packages/assets/scss/03-organisms/_sections-three-up.scss
```diff
@@ -97,4 +107,16 @@ $bp-section-3up-large-max: "max-width:950px";
     }
   }
 
+  .ma__card__wrapper {
```
clairesunstudio: We are trying to componentize the styles as much as we can, so this is considered antipattern and making it hard to debug in the future. Instead, could you add a card property/variant `card.theme = "c-white"` to the card scss and twig file that encapsulate the white card them. And pass that to the org section enhanced in openmass? technically, we'd like to avoid referencing openmass classes in mayflower if possible. Thanks!

clairesunstudio:
I pushed up a few changes:
1. Updated classname to follow BEM naming (see suggestions in openmass PR for changes)
2. Added eyebrow background color in the white variant
3. Added cards to sections-three-up and added cards to a page for visual regression testing
4. Approved backstop screenshots

changelogs/DP-38060.yml
```diff
@@ -0,0 +1,6 @@
+Changed:
+  - project: Patternlab
+    component: Card
+    description: Added new variant of "Card" to make the content vertical, with an included image. An optional eyebrow is also added.
```
clairesunstudio:
```suggestion
    description: Added new variant of "Card" to make the content vertical, with an included image. An optional eyebrow is also added. (#1960)
```

packages/assets/scss/02-molecules/_card.scss
```diff
@@ -14,10 +15,36 @@
...
+    &:focus,
+    &:focus-visible,
+    &:focus-within {
+
+      .ma__card__title span {
+        border-block-end-color: $c-link;
+      }
+    }
+
+    &:has(.ma__card--vertical) {
```
clairesunstudio: Do we need this? What does it do? I don't see any differences - if it's not used, can we remove this block of style?

I noticed the link is not highlighted on focus, only on hover. Can you please fix that? I think that's the only thing left, otherwise the Mayflower PR is good to go.

hotsauce: @clairesunstudio 
Thank you for catching that. It was used previously before I moved the `<a>` tag to be outside of the wrapper, so the styling actually applies further down where I am using `&--vertical`. I've moved it accordingly  :)

I also adjusted the focus state for the cards as well (again because of the move of the `<a>` tag) so now the title gets the underline on hover and the focus states as well.

hotsauce:
@clairesunstudio 
I've made the requested changes, please give it one more look over and approve it if all good :)

packages/assets/scss/02-molecules/_card.scss
```diff
@@ -198,6 +181,20 @@
     display: block;
     max-width: 600px;
     block-size: 100%;
+    border-block-end: 0;
+
+      &:hover,
```
clairesunstudio: Now, I see the link highlighted when my cursor is outside of the card surface.

I still don't get why this block is needed, can you please explain what does it do in addition to hover/focus styles added to the main class?

packages/assets/scss/02-molecules/_card.scss
```diff
@@ -198,6 +181,20 @@
     display: block;
     max-width: 600px;
     block-size: 100%;
```
clairesunstudio, line 183: Also not sure what block-size is doing here, maybe this is why the hover surface takes up the entire viewport width

packages/assets/scss/02-molecules/_card.scss
clairesunstudio: card-hover

hotsauce: @clairesunstudio 
Thank you for noticing that, I hadn't seen that on the openmass side because they are all inline but I see it now. The issue was that I should have kept the `.ma__card__wrapper` class on the actual link itself. That solved a couple of issues, and I addressed this overflow by setting it to `inline-size: fit-content` for the larger sizes since the overflow isn't a problem on tablet/desktop. Tweaking this also allowed me to get rid of/move the extra hover state classes as well.

Fix is pushing now.

packages/assets/scss/02-molecules/_card.scss
hotsauce: @clairesunstudio 
The block size is to ensure that the whole block gets the same height/background with flex-box. As I mentioned in the other comment, swapping in the right wrapper class helped address it, though the `block-size` is still needed here.

hotsauce:
@clairesunstudio 
The changes you requested have been pushed up, please re-review when you can

packages/patternlab/styleguide/source/_patterns/02-molecules/card.twig
```diff
@@ -1,35 +1,39 @@
...
-    <div class="ma__card__content{{ contentClass }}">
-      {% set headingLevel = card.level ? card.level : 3 %}
-      <h{{ headingLevel }} class="ma__card__title"><span>{{card.title}}&nbsp;{{ icon('arrow') }}</span></h{{ headingLevel }}>
-      <div class="ma__card__details{{ detailsClass }}">
-        {% if card.usage %}
-        <div class="ma__card__thumbnail">
-          <img src="{{card.thumbnail}}" alt="">
-        </div>
+<a class="ma__card__wrapper" href="{{ card.link }}" aria-labelledby="{{ card.id }}-title">
+  <div {% if not card.usage %}id="{{ card.id }}" {% endif %}class="ma__card{{ cardClass }} {{ cardThemeClass }} ma__card__wrapper">
```
clairesunstudio, line 10:
```suggestion
  <div {% if not card.usage %}id="{{ card.id }}" {% endif %}class="ma__card {{ cardClass }} {{ cardThemeClass }} ma__card__wrapper">
```

clairesunstudio: hmm can we instead of repeating the wrapper class, apply the necessary styles to .ma__card directly?

clairesunstudio:
Screenshot 2025-03-21 at 1 21 49 PM


I haven't done x-browser testing in feature3 - but it seems like on mobile, the top of the card needs some spacing.

packages/patternlab/styleguide/source/_patterns/02-molecules/card.twig
hotsauce: @clairesunstudio 
We need the repeated wrapper class there to help preserve the flexing with the added link above. I experimented with this lots of ways to get this to work but this is the most fool proof. I am happy to revisit it as a follow up ticket after the MVP goes in though

clairesunstudio: I don't think having nested elements with the same classname is a good practice - it's not clear which styles are needed - for example, the 20px margin bottom is being repeated. You are actually adding 40px but it's not clear in the scss file

clairesunstudio: I suggest repeating the necessary styles for `.ma__card` to fix the link issue - but most styles are probably not needed

clairesunstudio: Screenshot 2025-03-21 at 1 55 48 PM

hotsauce:
Follow up ticket to address some of the CSS cleanup - https://massgov.atlassian.net/browse/DP-38226

clairesunstudio (approved):

PR #1962, opened 2025-03-24
dp 36941/move right margin feedback button to middle of page on desktop/mobile

jarielrrz:
moves right margin feedback button to vertical middle of page on desktop/mobile


packages/assets/scss/01-atoms/_fixed-feedback-button.scss
```diff
@@ -11,7 +11,7 @@
   margin-top: 0 !important;
   position: fixed;
     right: 32px; // the height of the button before rotation
-    bottom: 18rem; // tested with the longest translation option
+    top: 50%; // fixes the button in the verticle middle of the screen
   transform-origin: top right;
   transform: rotate(-90deg);
```
clairesunstudio:
```suggestion
  transform: rotate(-90deg)  translateX(100%);
```
To position it at absolute center to ensure the button width is accounted for

changelogs/DP-36941.yml
```diff
@@ -0,0 +1,6 @@
+Fixed:
+  - project: Assets
+    component: FixedFeedbackButton
+    description: moves feedback button to the vertical middle of the screen.
```
clairesunstudio:
```suggestion
    description: moves feedback button to the vertical middle of the screen. (#1962)
```

clairesunstudio (approved):

clairesunstudio:
Mobile browsers change the height of the viewport. Tried a solution but didn't work in this case. This is the case currently on production. I think this fix is better than before https://stackoverflow.com/questions/44679794/position-fixed-on-chrome-mobile-causing-element-to-move-on-scroll-up-down/44680066#44680066

arthurbaghdas (approved):

PR #1967, opened 2025-03-31
DP-38339-Add-unique-ID-to-login-dropdown-component

jarielrrz:
This PR adds a conditional id attribute in button-dropdown.twig, so we can set gtm-loginto on the "Log In To" button for tracking.


packages/patternlab/styleguide/source/_patterns/02-molecules/button-dropdown.twig
```diff
@@ -5,7 +5,8 @@
     button: (button is iterable ? button : {})|merge({
       'class': buttonClass|default(''),
       'aria-haspopup': 'true',
-      'aria-controls': menuId|default('')
+      'aria-controls': menuId|default(''),
+      'id': button.id|default(null)
```
dstorozhuk: @jarielrrz  if button.id is empty - does id attribute exist on the element?

arthurbaghdas (approved):

PR #1973, opened 2025-04-08
DP-38621 Update header-search to allow more flexible scoping

andy-blum:
## Description

In order to scope search results to a given microsite, we’ll need to pass along some additional URL query parameters as hidden inputs. This conflicts with the existing way search works - defaulting to no scoping with a hard-coded dropdown option.

```html
<input type="hidden" name="microsite" value="microsite-id" />
```

## Existing Behavior
Screenshot 2025-04-16 at 10 46 35 AM

* If a user enters a search query in the search input and submits, the result is unscoped
* If a user clicks the first dropdown option, the result is unscoped. This option has visibly-hidden text saying "search <term>  across the entire site
* If a user clicks a different option below, a new hidden input is added via JavaScript with a name of `org` and a kebab-cased version of the organization's name. These options are added in by a data structure like:
```json
"headerSearch": {
  "suggestedScopes": [
    "Organization One",
    "Organization Two"
  ]
}
```

## New Behavior

The expected data structure has been expanded, to allow for the following:
* human-readable labels that may or may not cleanly convert to kebab-case through the existing JS functionality
* types other than `org`
* the option to hide the label on any dropdown menu item

```json
"headerSearch": {
  "suggestedScopes": [
    {
      "value": "executive-office-of-energy-and-environmental-affairs",
      "type": "org",
      "label": "in Executive Office of Energy and Environmental Affairs",
      "hideLabel": false
    },
    {
      "value": "department-of-conservation-recreation",
      "type": "org",
      "label": "in Department of Conservation & Recreation",
      "hideLabel": false
    },
  ]
}
```

Additionally, the component accepts a new `headerSearch.defaultScope` functionality that, if present, will add a new hidden input on page load. This default scope is removed if one of the suggestedScopes is chosen to avoid filter parameters compounding. The default scope's data structure matches that of the suggested scopes above.

```json
"headerSearch": {
  "defaultScope":  {
      "value": "microsite-title",
      "type": "microsite",
      "label": "in this microsite",
      "hideLabel": true
  },
}
```

## Related Issue / Ticket
- [JIRA issue](https://massgov.atlassian.net/browse/DP-38621)


packages/patternlab/styleguide/source/_data/headerSearch.json
```diff
@@ -8,6 +8,17 @@
       "text": "Search",
       "usage": "secondary"
     },
-    "suggestedScopes": ["Department of Conservation & Recreation", "Executive Office of Energy and Environmental Affairs"]
+    "suggestedScopes": [
+      {
+        "value": "Department of Conservation & Recreation",
```
clairesunstudio:
```suggestion
        "value": "department-of-conservation--recreation",
```

packages/patternlab/styleguide/source/assets/js/modules/headerSearch.js
```diff
@@ -61,37 +61,32 @@ export default (function (window, document,$) {
...
-            if (key(e).arrowUp) { 
+            if (key(e).arrowUp) {
                 focusIndex = [optionsCount -1];
                 setFocus(focusIndex);
             }
         });
-              
 
-        // transform org name to value
-        const slugify = (string = '') => (
```
clairesunstudio, line 92: The org values needs to be slugified - this is based on how the Google search parameter values are tokenized. 

**Before:** 
https://search.mass.gov/?q=tax&org=department-of-conservation--recreation

Screenshot 2025-04-28 at 12 21 23 AM


**After:** 
https://search.mass.gov/?q=tax&org=Department+of+Conservation+%26+Recreation (Google returns no results and search filter tag is empty)

Screenshot 2025-04-28 at 12 21 55 AM

Based on the new data structure, this function probably should be moved into the openmass codebase, the value should be passed as slugified (see code example in the data object)

packages/patternlab/styleguide/source/_data/headerSearch.json
clairesunstudio: Please update header-search.md to document the new data structure. All values should be "slugified" see the slugify function  https://github.com/massgov/mayflower/pull/1973/files#r2063473220

clairesunstudio:
Backstop test failures are false positives and irrelevant

packages/patternlab/styleguide/source/assets/js/modules/headerSearch.js
andy-blum: Correct, we're slugifying that value server side in this paired PR

https://github.com/massgov/openmass/pull/2944/files#diff-7ab1b4b71e92e70b91efbdb1fbe44676f793a4eab00ac5d1269180e1f0bb3e14R1313

clairesunstudio (approved):

PR #1976, opened 2025-04-18
DP-37992:  [Mayflower] Revisit Popover(Tooltip) component for Glossary Terms rendering

andy-blum:
## Description
Addresses accessibility feedback provided in [DP-37992 (comment)](https://massgov.atlassian.net/browse/DP-37992?focusedCommentId=441828)

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-38621)

## Steps to Test

1. Test all elements on the page using the tab key on your keyboard, make sure each action element receives focus and shows a visible focus outline that passes 3:1 contrast requirements.
2. To open a dialog use either the Spacebar or Enter key
3. Focus on open should go to the Close button inside the dialog
4. The escape key should close the dialog
5. While the dialog is open focus will not go behind or underneath the dialog if the user activates the tab key. Focus will remain on action elements inside the dialog until the user closes it using either the close button or the escape key.

## Additional Notes:
Added documentation to .twig file for markup choices

#### Today I learned...
paragraph tags only allow [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#phrasing_content). If other content types are added into a paragraph, the browser will sever the containing paragraph in two.


clairesunstudio (approved):
works great!

clairesunstudio:
The only thing I noticed in screen readers (tested in VoiceOver), the aria label "definition" is never announced when the focus is on the trigger button 
Screenshot 2025-04-23 at 4 00 13 PM

This was not pointed out in the ticket, so I'm fine with getting this in first and follow up with Alexa

PR #1980, opened 2025-04-29
DP-39184: Fix glossary term matching

andy-blum:
## Description
Refactors the poppover component's markup to eliminate whitespace

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-39184)

#### Today I learned...
`{% spaceless %}` has been removed from twig and the `|spaceless` filter is also deprecated.

### DIFF
--- a/packages/patternlab/styleguide/source/_patterns/02-molecules/popover.twig
+++ b/packages/patternlab/styleguide/source/_patterns/02-molecules/popover.twig
@@ -9,16 +9,18 @@
  #
  # Additionally, we add tabindex="-1" to the dialog element to catch focus if
  # users interact with the popover content, preventing it from closing on 'focusout'
+ #
+ # To prevent extra spaces from appearing around the popover, also need to collapse all whitespace, hence the set-blocks.
  #}
-<span class="popover">
-  <button type="button" class="popover__trigger js-popover-trigger">{{tooltip.openText}}</button>
-  <span role="dialog" id="popover-{{ tooltip.controlId }}" aria-labelledby="popover-{{ tooltip.controlId }}-label" aria-modal="true" class="popover__dialog js-popover-dialog" tabindex="-1">
-    <button type="button" class="popover__close js-popover-close" aria-label="{{ tooltip.closeText }}">{{icon('close')}}</button>
-    {% if tooltip.title %}
-      {% set headingLevel = tooltip.level ? : 2 %}
-      <span role="heading" aria-level="{{ headingLevel }}" class="popover__title" id="popover-{{ tooltip.controlId }}-label">{{ tooltip.title }}</span>
-    {% endif %}
-    <span class="popover__body">{{ tooltip.message }}</span>
-  </span>
-  <span class="popover__caret"></span>
-</span>
+
+{% set trigger %}<button type="button" class="popover__trigger js-popover-trigger">{{tooltip.openText}}</button>{% endset %}
+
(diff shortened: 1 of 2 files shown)


packages/patternlab/styleguide/source/_patterns/02-molecules/popover.twig
```diff
@@ -9,16 +9,18 @@
...
-      <span role="heading" aria-level="{{ headingLevel }}" class="popover__title" id="popover-{{ tooltip.controlId }}-label">{{ tooltip.title }}</span>
-    {% endif %}
-    <span class="popover__body">{{ tooltip.message }}</span>
-  </span>
-  <span class="popover__caret"></span>
-</span>
+
+{% set trigger %}<button type="button" class="popover__trigger js-popover-trigger">{{tooltip.openText}}</button>{% endset %}
+
+{% set dialog_opentag %}<span role="dialog" id="popover-{{ tooltip.controlId }}" aria-labelledby="popover-{{ tooltip.controlId }}-label" aria-modal="true" class="popover__dialog js-popover-dialog" tabindex="-1">{% endset %}
```
dstorozhuk, line 18: @andy-blum  is that possible to use `speceless` filter from Twig? 
I see many examples of using it in core, for example docroot/core/modules/system/templates/select.html.twig. `apply spaceless`

So we can preserve the formating on source files, which improve the readability and maintain.

dstorozhuk (changes requested):

packages/patternlab/styleguide/source/_patterns/02-molecules/popover.twig
andy-blum: This is not recommended. Drupal moved instances of `{% spaceless %}` to `{% apply spaceless %}` [when updating to Twig 3](https://www.drupal.org/project/drupal/issues/3094850), but [the  spaceless filter is deprecated as of 3.12](https://twig.symfony.com/doc/3.x/filters/spaceless.html). Core will need to update their usages to something else in the future.

dstorozhuk: Also here https://twig.symfony.com/doc/3.x/templates.html#templates-whitespace-control

andy-blum: Whitespace control only works to trim the start/end of a tag.

> The modifiers can be used on either side of the tags like in {%- or -%} and they consume all whitespace for that side of the tag. It is possible to use the modifiers on one side of a tag or on both sides:
```twig
{% set value = 'no spaces' %}
{#- No leading/trailing whitespace -#}
{%- if true -%}
    {{- value -}}
{%- endif -%}
{# output 'no spaces' #}

<li>
    {{ value }}    </li>
{# outputs '<li>\n    no spaces    </li>' #}

<li>
    {{- value }}    </li>
{# outputs '<li>no spaces    </li>' #}

<li>
    {{~ value }}    </li>
{# outputs '<li>\nno spaces    </li>' #}
```

Since our markup spans several lines, we'll have `\n` and `\s` characters _within_ our tag, which won't get stripped out.

dstorozhuk: Ok, thanks.

dstorozhuk (approved):
Looks good to me. 
Two failed will be fixed separetly as per @clairesunstudio  https://teams.microsoft.com/l/message/19:4318cf9372fa48a1a56333cc6777db8d@thread.v2/1746037802711?context=%7B%22contextType%22%3A%22chat%22%7D

PR #1981, opened 2025-04-30
DP-38077: Switch Mayflower to CSS variables for dark mode

notraiyan:
Added:
  - project: Assets
    component: _color-tokens, _colors
    description: Added css variables for colors (#1981)
    issue: DP-38077
    impact: Patch

Changed:
  - project: Assets, Core, Patternlab, React, Site
    component: scss files
    description: Changed colors to use the new css variables rather than scss variables (#1981)
    issue: DP-38077
    impact: Major

## Description

- SCSS variables are currently used. When files are compiled in css they become hardcoded values and you cannot change it based on user’s preference or system’s settings.
- Changed them to css variables so they can be changed in runtime.
- Changed colors across all the subpackages

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-38077)


packages/assets/scss/02-molecules/_relationship-indicators.scss
```diff
@@ -192,7 +192,7 @@
       }
 
       a {
-        color: $c-primary-alt;
+        color: var(--mf-c-primary-alt);
         box-shadow: inset 0 0 $c-white, inset 0px -1px rgba($c-primary-alt, 0.4);// This can be replaced with border-bottom. See L.163.
```
DmitriiTartynovLCM, line 196: Should `$c-white` be replaced too?

Please let me know if the idea here is to replace both `$c-white` and `$c-primary-alt` at once in [DP-39213](https://massgov.atlassian.net/browse/DP-39213)

packages/assets/scss/02-molecules/_relationship-indicators.scss
```diff
@@ -250,7 +250,7 @@
     &.single [aria-labelledby="secondary"] {
 
       a {
-        color: $c-primary-alt;
+        color: var(--mf-c-primary-alt);
         box-shadow: inset 0 0 $c-white, inset 0px -1px rgba($c-primary-alt, 0.4);
```
DmitriiTartynovLCM, line 254: Same question here.

changelogs/DP-38077.yml
```diff
@@ -0,0 +1,13 @@
...
+  - project: Assets
+    component: _color-tokens, _colors
+    description: Added css variables for colors (#1981)
+    issue: DP-38077
+    impact: Patch
+
+Changed:
+  - project: Assets, Core, Patternlab, React, Site
+    component: scss files
+    description: Changed colors to use the new css variables rather than scss variables (#1981)
```
DmitriiTartynovLCM: Not sure if it's important here, but CSS and SCSS should probably be spelled in uppercase letters.

packages/assets/scss/00-base/_colors.scss
```diff
@@ -49,3 +49,50 @@ $c-bd-input-dark:        $c-gray !default;
 
 // Box shadow
 $c-box-shadow-black:      rgba($c-black, 0.25) !default;
```
DmitriiTartynovLCM, line 51: The `$box-shadow` SCSS variable in a different file [still references](https://github.com/massgov/mayflower/blob/058177f43a4ca9496b46a3762fe74a365044ff22/packages/assets/scss/00-base/_variables.scss#L110) `$c-box-shadow-black` while it should probably just use the CSS variable.

DmitriiTartynovLCM (commented):
Left a couple of minor comments. Nice and solid work! I'll check what's going on with the test failure a bit later if that's OK.

packages/assets/scss/02-molecules/_relationship-indicators.scss
notraiyan: Yes, the idea was to replace both $c-white and $c-primary-alt at once in [DP-39213](https://massgov.atlassian.net/browse/DP-39213)

packages/assets/scss/02-molecules/_relationship-indicators.scss
notraiyan: Yes, the idea was to replace both $c-white and $c-primary-alt at once in [DP-39213](https://massgov.atlassian.net/browse/DP-39213)

packages/assets/scss/00-base/_colors.scss
notraiyan: My bad, I missed this one. Fixed it.

changelogs/DP-38077.yml
notraiyan: Updated it.

clairesunstudio:
Patternlab test failures are false positives (one of them is in a ticket to resolve)  - please ignore

DmitriiTartynovLCM:
No unresolved feedback left from my side. @notraiyan Thanks for the updates!

@clairesunstudio It's ready for your review, I believe.

clairesunstudio:
The changes look good in general! I have a couple of questions/comments:

1. Is it necessary or is it a good practice to keep both the SCSS and CSS variables? Is this mainly for backward compatibility? My main concern here is that this may lead to inconsistency and fragmentation across the codebase, as both types of variables are exposed. If we decide to keep both, we need to clarify their purposes and document the public API for developers, potentially moving the scss variables into one file (color-tokens-scss?) and make sure they are not directly referenced in the component styles, and document when they should be used and what are some antipatterns. We probably also want to update the variable naming in the core documentation - https://mayflower.digital.mass.gov/core/index.html?path=/docs/foundation-color--page or keep them as more tech agnostic. - $c-error-red --> error-red. 

2. How should we handle variable overrides at a high level? I'm not sure if it's in used in other repos but we do have the `!default` flag applied all the SCSS variable declarations, allowing them to be overriden.

3. I'd like to better understand how this change will impact other implementations. You mentioned runtime compilation which enables color modes. One benefit of directly storing tokens into native CSS variables is that it removes the requirement of SASS compilation, and makes the primitive tokens in mayflower more framework agnostic, which may potentially enable mayflower-assets package to be used in Tailwind and CSS-in-JS? 

We want to take incremental steps to improve the adoption of mayflower-assets and broaden its use cases, so want to make sure the decision we make here is in line with that goal, while maintaining the single source truth.

Ethansev:
> The changes look good in general! I have a couple of questions/comments:
> 
> 1. Is it necessary or is it a good practice to keep both the SCSS and CSS variables? Is this mainly for backward compatibility? My main concern here is that this may lead to inconsistency and fragmentation across the codebase, as both types of variables are exposed. If we decide to keep both, we need to clarify their purposes and document the public API for developers, potentially moving the scss variables into one file (color-tokens-scss?) and make sure they are not directly referenced in the component styles, and document when they should be used and what are some antipatterns. We probably also want to update the variable naming in the core documentation - https://mayflower.digital.mass.gov/core/index.html?path=/docs/foundation-color--page or keep them as more tech agnostic. - $c-error-red --> error-red.
> 2. How should we handle variable overrides at a high level? I'm not sure if it's in used in other repos but we do have the `!default` flag applied all the SCSS variable declarations, allowing them to be overriden.
> 3. I'd like to better understand how this change will impact other implementations. You mentioned runtime compilation which enables color modes. One benefit of directly storing tokens into native CSS variables is that it removes the requirement of SASS compilation, and makes the primitive tokens in mayflower more framework agnostic, which may potentially enable mayflower-assets package to be used in Tailwind and CSS-in-JS?
> 
> We want to take incremental steps to improve the adoption of mayflower-assets and broaden its use cases, so want to make sure the decision we make here is in line with that goal, while maintaining the single source truth.

@clairesunstudio Raiyan is going to be OOO for a while so I'll try my best here to answer some of these questions.

1. The plan is to completely transition away from SCSS variables and only use CSS variables. I'm actually working on [this ticket](https://massgov.atlassian.net/browse/DP-39213) to finish the transition but I think it may make sense to merge this in first when we can.
2. Css variables can be overriden by styling at the component level, so it should be easy to work within any particular context without needing to make changes to the global default.
3. Yeah so SCSS variables are hard-coded at build time so the variable becomes non-existent and renders static css anyways. We cannot make changes to the variable if it doesn't exist in the first place and therefore the lack of flexibility becomes an issue. Because css variables exist in the browser and are resolved at runtime, we're free to manipulate them anyway we want and even update them with javascript if necessary. The tailwind config allows us to use these css variables to define its theme as well and overall provides flexibility that other frameworks can use because at the end of the day it's still primitive css.

I'd appreciate any fact checking from @DmitriiTartynovLCM in case I'm wrong here 😅

DmitriiTartynovLCM:
@Ethansev @clairesunstudio Just adding my two cents here.

1. Complete removal of SCSS variables in the next (even major) Mayflower release may be a breaking change that is too disruptive for existing projects. In the long run, they can still be useful for projects that do import SCSS from Mayflower Assets. For example, SCSS mixins may still need SCSS variables for different color transformations, etc. But, we could also call them obsolete and ask all the projects to stop using those so that they could be completely removed in one of the future versions.

2. Projects can always replace values of the CSS variables in the default rules by using CSS cascading. The CSS vars added in this PR are applied to the `:root` selector. The projects can always add rules with higher specificity and set different values to the same CSS vars, e.g.
```css
:root body {
  --mf-c-primary: #FFFFFF;
}
```

clairesunstudio:
Next step: To get rid of SCSS color variables inside of mayflower, we need to replace scss color variables in 1) mixins and 2) rgba() functions. 
For example:
1) In mixins: 
```
background-color: $color; --> background-color: var(#{$color}); 
```
2) In rgba(), use [color-mix()](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix)
```
background-color: rgba($c-primary,.75); --> background-color: color-mix(in srgb, var(--mf-c-primary) 75%, transparent);
```

In the codebase, we should standardize on the CSS variables going forward. Can we create a follow up ticket for this? @Ethansev

clairesunstudio (approved):
Approved! Waiting on Mass.gov release hotfix to go in before merging

PR #1990, opened 2025-05-21
DP-39495-eliminate-horizontal-page-scroll

clairesunstudio:
Fixed:
  - project: Patternlab,Assets
    component: Elements
    description: Eliminate page horizontal scroll. (#1990)
    issue: DP-39495
    impact: Patch

### DIFF
--- a/changelogs/DP-39495.yml
+++ b/changelogs/DP-39495.yml
@@ -1,0 +1,59 @@
+# Every PR must have a changelog. To create a changelog:
+# 1. Make a copy of this file and name it with the ticket number.
+# 2. Keep the original format of one of the examples at the bottom, and override each element of it using ___TEMPLATE___ for reference
+# 3. Remove all comments from the copied file
+
+#___CHANGE TYPES___
+# - Added for new features.
+# - Changed for changes in existing functionality.
+# - Deprecated for soon-to-be removed features.
+# - Removed for now removed features.
+# - Fixed for any bug fixes.
+# - Security in case of vulnerabilities.
+# e.g. Added
+
+#___PROJECT___
+# - Patternlab
+# - React
+# - Docs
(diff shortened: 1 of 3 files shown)


Ethansev:
Hi after inspecting it, looks like the overlays may be the culprit. I believe you can update the `menu-overlay` and `alert-overlay` divs to set `width` to `100%` instead of `100vw`.

^ Update: should be done on the body's width instead

clairesunstudio:
Setting `width: 100vw` in the body causes the horizontal scroll because vw ignores the vertical scroll bar. The solution is to set it to 100%. It has to be explicit, removing the width causes the body to shift when the menu is expanded.
Doing this instead of `overflow-x:hidden` is safer, but it doesn't completely eliminate horizontal scrolls. However, the slight shifts horizontally are common on other websites. For example, [www.google.com](https://www.google.com/), [www.apple.com](https://www.apple.com/), and [www.linkedin.com](https://www.linkedin.com/) don't set a width on the body and [www.x.com](https://www.x.com/) simply sets it to 100%; It's worth noting that https://designsystem.digital.gov/ and boston.gov sets `overflow-x:hidden` on the body tag to completely disable the shifts - I'm going to check with the access team to see if the shift costs any accessibility concerns. If not, we can go with body width:100% instead of `overflow-x:hidden`

clairesunstudio:
Unintended horizontal scrolling or layout shift is an accessibility issue. The main indicator is the scrollbar. Setting the body width to 100% eliminated the horizontal scrollbar, a scrollbar was the main issue and this fix should be sufficient. 

Before:
Screenshot 2025-05-23 at 1 46 27 PM

After:
Screenshot 2025-05-23 at 1 46 36 PM

arthurbaghdas (approved):

PR #1997, opened 2025-06-30
DP-40113: Pagination Links Not Functional on Upcoming and Past Events Page

arthurbaghdas:

## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

### DIFF
--- a/packages/patternlab/styleguide/source/assets/js/helpers/getHandlebarTemplate.js
+++ b/packages/patternlab/styleguide/source/assets/js/helpers/getHandlebarTemplate.js
@@ -1,3 +1,7 @@
+Handlebars.registerHelper('eq', function (a, b) {
+  return a === b;
+});
+
 module.exports = function(name) {
   if (Handlebars.templates === undefined || Handlebars.templates[name] === undefined) {
       jQuery.ajax({
--- a/packages/patternlab/styleguide/source/assets/js/templates/pagination.html
+++ b/packages/patternlab/styleguide/source/assets/js/templates/pagination.html
@@ -1,0 +1,49 @@
+<div class="ma__pagination js-pagination">
+    <nav class="ma__pagination__container" aria-label="Pagination Navigation">
+        <a
+                class="ma__pagination__prev js-pagination-prev {{#if pagination.prev.disabled}}disabled{{/if}}"
+                role="button"
+                {{#if pagination.prev.disabled}}
+                aria-disabled="true"
+                {{else}}
+                aria-label="Go to {{pagination.prev.text}} page"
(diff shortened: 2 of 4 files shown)


packages/patternlab/styleguide/source/assets/js/templates/pagination.html
```diff
@@ -0,0 +1,49 @@
+<div class="ma__pagination js-pagination">
```
clairesunstudio, line 1: I think we should avoid maintaining two duplicate templates in the system - this is a copy of packages/patternlab/styleguide/source/_patterns/02-molecules/pagination.twig in handlebar. We want to maintain a source of truth for the pagination pattern and this will makes it challenging to update and debug in the future. Can we do what the [location-listing template](https://github.com/massgov/mayflower/blob/a1fa86334484a434f77f29337f83a6ab9c949a92/packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/location-listing.twig) is doing - by including the pagination twig in the template directly without using twiggy?

packages/patternlab/styleguide/source/assets/js/helpers/getHandlebarTemplate.js
```diff
@@ -1,3 +1,7 @@
+Handlebars.registerHelper('eq', function (a, b) {
```
clairesunstudio, line 1: Please add a changelog

packages/patternlab/styleguide/source/assets/js/templates/results-heading.html
```diff
@@ -0,0 +1,50 @@
+<div class="ma__results-heading js-results-heading">
```
clairesunstudio, line 1: According to @arthurbaghdas:
> I think its related to Acquia cloud next, its blocking the template rendering with JS, that molecule might be used in other places as well, only for the event listing where we call twiggy I replace it with handlebars. So the issue is that acquia is blocking twig template rendering called from JS code

[Results heading](packages/patternlab/styleguide/source/_patterns/02-molecules/results-heading.twig) and [AJAX pattern](/Users/minghuasun/Documents/Github/mayflower/packages/patternlab/styleguide/source/assets/js/helpers/jQueryExtend_ajaxPattern.js) (which is used in accordion and alerts) are the only other pattern that's currently relying on twiggy. Also the icon function relies on twiggy. Should we try to move away from twiggy into handlebar for all dynamic templates? If so, I think we should also clean up the unused code to prevent confusion.

PR #2007, opened 2025-07-31
patternlab/DP-40850-Multiple-language-bars-on-the-same-page-generate-inaccessible-links

jarielrrz:
Fixes accessibility issue with multiple language translation links on the same page


clairesunstudio (commented):
Functionally works well in openmass. Can you please update the corresponding JSON file and the .md file to include the new changes? The access team reviews components in mayflower patternlab and the changes are currently not reflected.  

Also, can we move packages/patternlab/styleguide/source/_patterns/03-organisms/by-author/inline-links-language.twig into molecules instead? It's an anti pattern in atomic design to include organisms in molecules. (please update the [openmass PR](https://github.com/massgov/openmass/pull/3068/files) accordingly.)

clairesunstudio (approved):

PR #2037, opened 2026-01-21
DP-44415-A11Y-Broken-aria-link-context-for-directions-link-on-org-page

jarielrrz:
Only generate linkContext when item.label exists


packages/patternlab/styleguide/source/_patterns/02-molecules/contact/contact-item.twig
```diff
@@ -9,6 +9,10 @@
 {% else %}
   {% set link = item.link %}
 {% endif %}
+{# Remove info if it's "directions" to prevent redundant title #}
+{% if link.info and link.info|lower == 'directions' %}
```
clairesunstudio: @jarielrrz , is it possible to just remove the link.info value from openmass instead of adding this logic in the template?

packages/patternlab/styleguide/source/_patterns/02-molecules/contact/contact-item.twig
```diff
@@ -54,12 +58,18 @@
     {# Address - RTE version of value and look for directions link #}
   {% elseif item.type == "address" %}
     <div class="ma__contact-group__address">
```
clairesunstudio, line 55: I think this might be a simpler fix -->`{{ address|nl2br|replace({'<br />': '<br aria-hidden="true"/>'})|raw }}`

clairesunstudio (approved):

PR #2055, opened 2026-03-09
DP-45023: Google Translate widget updates for accessibility

tfleming-ma:
## Description
- New `modal` Twig template based on Figma design and built from example code
- New `google-translate` Twig template based on Figma design and built from example code
- Added translate SVG icon
- Various changes to remove code related to old Google Translate widget that did things like add top to the body
- Made best effort to use new spacing and color tokens, used existing color tokens when available

## Related Issue / Ticket

- [JIRA issue](https://massgov.atlassian.net/browse/DP-45023)

## Additional Notes:

- When loaded on mass.gov, it should be noted that the modal and translate markup load twice, once for the main menu seen at larger screen sizes and once for the mobile menu seen on smaller screen sizes. When you click on Language in either menu it is showing a different modal. This might cause some issues for some screen sizes, so this is an area we should definitely test.
- The modal JS and Google Translate JS are based on the example provided in the ticket. It has been modified to dynamically work with multiple modal instances on a page.
- Some design elements for the modal, such as the button group and content are rolled into the Google Translate Twig template, but I included styles in the Modal template's SCSS that will style the button group.
- The new languages.json includes all languages that are available in openmass's theme configuration
- I have done little testing on the React component

* Header, pages, utility nav

#### @TODO
- React component
- Adding/updating tests
- I don't see a way of hiding the Google Translate spinner that appears. It looks like Google intentionally uses random classnames to prevent hiding of the spinner, but maybe nothing needs to be done to address this.


packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
```diff
@@ -2,12 +2,9 @@
   <ul class="ma__utility-nav__items">
     {% if googleLanguages %}
       <li class="ma__utility-nav__item">
-        <div class="ma__utility-nav__translate">
```
clairesunstudio, line 5: On mixed headers appendChild(undefined) fails because mainNavMixed.js still assumes that element exists

Image

packages/patternlab/styleguide/source/_patterns/02-molecules/modal.twig
```diff
@@ -0,0 +1,27 @@
+{% set modalId = modal.id|default('ma__modal-' ~ random()) %}
+{% set modalTitleId = modalId ~ '-title' %}
+
+<button type="button" class="ma__utility-nav__link" data-modal-trigger="{{ modalId }}" aria-haspopup="dialog" aria-controls="{{ modalId }}">
+    {{ icon(modal.icon, "20", "20") }}
+    <span>{{ modal.buttonText }}</span>
```
clairesunstudio: This is not visible in patternlab 

Image

packages/assets/scss/03-organisms/_modal.scss
```diff
@@ -0,0 +1,215 @@
+@use "00-base/configure" as *;
+
+.ma__note {
+  background-color: var(--mf-c-gray-light);
```
clairesunstudio: Actually, could we bring in the color tokens and start referencing to the semantic color tokens in this component? EEA might be consuming this component after it's shipped

packages/patternlab/styleguide/source/_patterns/02-molecules/modal.twig
```diff
@@ -0,0 +1,27 @@
+{% set modalId = modal.id|default('ma__modal-' ~ random()) %}
+{% set modalTitleId = modalId ~ '-title' %}
+
+<button type="button" class="ma__utility-nav__link" data-modal-trigger="{{ modalId }}" aria-haspopup="dialog" aria-controls="{{ modalId }}">
```
clairesunstudio: I wonder if should use a different classname that detaches from the utility nav in the modal component.

packages/patternlab/styleguide/source/_patterns/02-molecules/modal.twig
clairesunstudio: This might be related to the `.ma__utility-nav__translate` being removed
On desktop, we no longer render the old .ma__utility-nav__translate structure, so the wide-header translate layout no longer matches the markup.

packages/patternlab/styleguide/source/_patterns/03-organisms/modal.twig
```diff
@@ -0,0 +1,27 @@
+{% set modalId = modal.id|default('ma__modal-' ~ random()) %}
```
clairesunstudio: Can we make this into an organism instead?

packages/react/src/components/base/Icon/index.js
```diff
@@ -212,6 +212,7 @@ export { default as IconTrafficSignal } from './IconTrafficSignal'
 export { default as IconTrainRegional } from './IconTrainRegional'
 export { default as IconTrain } from './IconTrain'
 export { default as IconTransit } from './IconTransit'
+export { default as IconTranslate } from './IconTranslate'
```
clairesunstudio, line 215: rush start:react fails on:

ERROR in ./src/components/base/Icon/index.js
Module not found: Error: Can't resolve './IconTranslate' in '/Users/minghuasun/Documents/Github/mayflower/packages/react/src/components/base/Icon'

ERROR in ./src/components/atoms/buttons/GoogleTranslateElement/index.js
Module not found: Error: Can't resolve 'MayflowerReactBase/Icon/IconTranslate' in '/Users/minghuasun/Documents/Github/mayflower/packages/react/src/components/atoms/buttons/GoogleTranslateElement'

clairesunstudio: If this is not a straightforward fix, let's leave the React part out and create a separate ticket this

packages/patternlab/styleguide/source/_patterns/02-molecules/google-translate.twig
```diff
@@ -0,0 +1,55 @@
+<!-- Status message for screen readers -->
+<div id="ma__translate-status" class="ma__status-message" role="status" aria-live="polite" aria-atomic="true"></div>
+
+<!-- Accessible language selector -->
+<div class="ma__translate-container">
+  <div class="ma__translate-select-wrapper">
+    <select
+        id="ma__translate-select"
```
clairesunstudio, line 8: I'm seeing these IDs `ma__translate-select, ma__translate-help, ma__translate-status, and google_translate_element` duplicated in header-mixed and header-hamurger because they are being rendered multiple times.

clairesunstudio: https://pr3290-smzou3nj96sjvqxpmqbxpye5rmatu5cu.tugboatqa.com/topics/finding-a-job

clairesunstudio:
Some styles are mismatched with the figma design specs:
1. The dialog is missing the border. Figma specifies a 1.5px neutral border
2. top spacing is smaller than Figma specs, also the close button target area is smaller than 44x44 in design 
Screenshot 2026-03-30 at 11 42 48 AM
3. There seems to be a transparency on the modal background --> http://localhost:3000/?p=pages-Homepage
4. Screenshot 2026-04-01 at 3 22 44 PM
5. On mobile, there are two language buttons
Screenshot 2026-04-01 at 3 22 57 PM

packages/assets/scss/03-organisms/_modal.scss
clairesunstudio: Done

packages/patternlab/styleguide/source/_patterns/02-molecules/modal.twig
clairesunstudio: This is fixed

packages/patternlab/styleguide/source/_patterns/03-organisms/by-template/utility-nav.twig
clairesunstudio: This is fixed

packages/patternlab/styleguide/source/_patterns/03-organisms/modal.twig
clairesunstudio: done

packages/patternlab/styleguide/source/_patterns/02-molecules/modal.twig
clairesunstudio: done

packages/patternlab/styleguide/source/_patterns/02-molecules/google-translate.twig
clairesunstudio: I updated the scripts so that now we keep a single translate modal instance and move that same DOM node between wide/narrow utility-nav containers, we avoid the problems that came from rendering multiple identical modals at once, such as duplicate IDs and mismatch references

clairesunstudio:
All the comments above have been fixed. Main fixes:

- Brought in all tokens, and update all the css variable references in modal.scss
- Refactor modal.scss, added variants matching the figma design and cleaned up unused styles
- Matched modal responsive design in figma 
- Moved modal from molecules into organisms (probably need to update the path in openmass)
- Fixed modal example rendering in Patternlab (cleaned up classnames to detach from utility nav)
- FIxed duplicate ID and potential mismatch of aria references by keeping a single translate modal instance and move that same DOM node between wide/narrow utility-nav containers, we avoid the problems that came from rendering multiple identical modals at once
- Implemented focus trapping

packages/assets/scss/02-molecules/_main-nav.scss
```diff
@@ -161,12 +161,6 @@
       right: -300px;
       transition: visibility .5s ease, right .5s ease;
 
-      .translated-rtl &,
```
clairesunstudio, line 164: I didn't test the removals of these styles - unsure why we had these to begin with. Can you please verify there's no impact removing the .translated-rtl &, .translated-ltr & styling?

tfleming-ma: @clairesunstudio The removed styles are all related to adding height and moving the entire body down to accommodate the language bar when a language is selected. With those styles in place, you end up with a gap at the top of the page.

packages/assets/static/images/icons/translate.svg
tfleming-ma, line 1: @clairesunstudio This icon is being imported at a much smaller size than the other icons I see that you added. Should those icons be fixed or this one?

packages/patternlab/styleguide/source/_patterns/03-organisms/modal/modal.twig
```diff
@@ -0,0 +1,34 @@
+{% set modalId = modal.id|default('ma__modal-' ~ random()) %}
+{% set modalTitleId = modalId ~ '-title' %}
+{% set modalWrapperClass = modal.wrapperClass|default('') %}
+{% set modalTriggerClass = modal.triggerClass|default('') %}
+{% set modalStatus = modal.status|default(modal.type|default('info')) %}
```
tfleming-ma:
```suggestion
{% set modalStatus = modal.status|default('info') %}
```

packages/assets/static/images/icons/translate.svg
clairesunstudio: Screenshot 2026-04-07 at 9 23 35 AM
They look the same size to me in the svg viewbox, and in Patternlab. Can you show me what you mean?

packages/assets/scss/02-molecules/_main-nav.scss
clairesunstudio: wow don't remember we were manually accommodating the translate bar height. Good riddance!

clairesunstudio: I wonder if we should clean up the variable `$header-mobile-controls-height`

clairesunstudio: I think we can remove that, and its references

packages/assets/scss/02-molecules/_google-translate.scss
```diff
@@ -0,0 +1,170 @@
...
+}
+
+.ma__translate-reset-button {
+  background-color: var(--mds-inverse);
+  border-color: var(--mds-border-brand-primary-mid);
+  color: var(--mds-text-and-icons-brand-primary-mid);
+}
+
+.ma__translate-reset-button:hover {
+  background-color: var(--mds-background-adaptive-brand-neutral-lowest);
```
clairesunstudio:
```suggestion
  background-color: var(--mds-background-adaptive-brand-primary-high);
```

packages/assets/scss/02-molecules/_google-translate.scss
```diff
@@ -0,0 +1,170 @@
...
+  border-color: var(--mds-background-adaptive-brand-primary-high);
```
clairesunstudio:
```suggestion
  border-color: var(--mds-border-brand-primary-high);
```

packages/assets/scss/02-molecules/_google-translate.scss
```diff
@@ -0,0 +1,170 @@
...
+  color: var(--mds-background-adaptive-brand-primary-high);
```
clairesunstudio:
```suggestion
  color: var(--mds-inverse);
```

packages/assets/scss/02-molecules/_google-translate.scss
```diff
@@ -0,0 +1,170 @@
...
+}
+
+.ma__translate-button:disabled {
+  background-color: var(--mds-background-section-brand-neutral-lowest);
+  border-color: var(--mds-border-brand-neutral-low);
+  cursor: not-allowed;
+}
+
+.ma__translate-reset-button {
+  background-color: var(--mds-inverse);
```
clairesunstudio:
```suggestion
  background-color: var(--mds-static-white);
```

clairesunstudio:
Accessibility review feedback (WIP):
- [x] Color contrast for select dropdown border doesn’t meet 3:1 requirement #BABABA against white background [WebAIM: Contrast Checker](https://webaim.org/resources/contrastchecker/) 
- [x] When activating the Change language blue button using the keyboard it changes the background color to a light gray and the text color remains white while the widget updates the language, the 4.5:1 contrast should remain during this temporary state
- [x] If I change the language to let’s say German, then reopen the modal and move the dark blue change language button, nothing happens, I would expect it to just close the modal and move focus back to the the trigger button
- [x] Whenever updating the language using the widget with either the update or reset button after the modal is closed the focus should move to the trigger button
> I've already fixed the focus setting part. But on mobile this means we need to keep the mobile tray open after clicking on the "translate" or "show original" button - please roll back auto closing the menu tray after modal actions @tfleming-ma 
- [x] Move the trigger button of the modal outside of the mobile menu so that users can easily find it to change the language especially when using browser zoom, there is currently a bug when using browser zoom that the language button doesn’t show right away in the mobile menu when zooming at 175%
> For this, let's table the part of moving translate trigger button outside of mobile tray (This is a design decision, which will happen down the line with the new header/utility menu redesign). I've already fix the bug described in the latter part.
- [x] The modal is not usable when zooming because it doesn’t show the full height of the modal (see screenshot) - for me the issue starts at 250% browser zoom on Google Chrome, 22” external monitor. 
> This is a crucial issue @tfleming-ma 
- [x] Currently the screen reader doesn’t announce a label for the select element, use aria-labelledby to point to the h2 heading ID
> With the current setup, this might be tricky because the heading is in the modal component and the select is in the google translate component @tfleming-ma I will leave this to you to decide what we want to do here
- [x] Remove the screen reader only text it is verbose and redundant: “Select a language from the dropdown menu and press the Apply button to translate the page”
> This came from Alexa's example. But we will remove it
- [x] Optional for future improvement: change the h2 heading text to “Translate page” and use a <label> tag with text “Language” for the select dropdown.
> Ignore this for now. This is a content/design decision

clairesunstudio:
I think we need to think about what goes into the modal vs modal internal block (such as the google translate). The buttons in my opinion should be move to the modal level, because the keyboard interaction ties the focus back and forth between those buttons and the modal trigger button (I added that in the google translate JS which is not idea). Also from a component matching standpoint, the Figma component does include those action buttons.  
However, in this case, the heading of the modal content needs to be a label for the select field (this is an edge case in my opinion) Let me know if you want to discuss the approach for that one

clairesunstudio:
- Fixed the label association by setting a modal id (it needs to match in openmass)
- Fixed the high zoom modal display 
before:
Screenshot 2026-04-15 at 9 21 57 AM
after:
Screenshot 2026-04-15 at 9 21 07 AM

tfleming-ma:
Fixed the focus in mobile when switching languages. Now, the menu will reopen with the focus on the Language menu item.

dstorozhuk (approved):

dstorozhuk:
A failed test is not related; it's simply that the map isn't fully rendered. 
image

PR #2065, opened 2026-05-04
DP-46593-Blue-focus-borders-on-blue-site-banner-on-safari

jarielrrz:
Set a white focus outline on site banner to improve focus visibility in Safari.


tfleming-ma (approved):
Looks good to me. Some unrelated changes causing test failures.

PR #2076, opened 2026-06-09
SSR: update GitHub Actions to latest versions

justinLCM:
Updated common actions (checkout, docker/login-action, etc.) to resolve deprecation warnings.


JimmyIL:
looks like there is a newer pull request making the same or even more compatible changes in both actions and the app code.  We should probably proceed with that one.

PR #2084, opened 2026-09-01
DP-41516: Updated site-logo to use picture

tfleming-ma:

## Related Issue / Ticket

- [JIRA issue]()
- [Github issue]()

## Steps to Test

1.

## Screenshots

## Additional Notes:

Anything else to add?

*

#### @TODO

*

### DIFF
--- a/packages/patternlab/styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
+++ b/packages/patternlab/styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
@@ -1,7 +1,10 @@
 <div class="ma__site-logo">
   <a href="{{ url ? url : '/' }}">
     {% if sealImage %}
-      <img src="{{ sealImage }}" alt="{{ altTitle ? altTitle : 'Mass.gov home' }}" width="45" height="45" />
+      <picture>
+        <source srcset="{{ sealWhiteImage }}" media="(prefers-color-scheme: dark), (forced-colors: active)" />
+        <img src="{{ sealImage }}" alt="{{ altTitle ? altTitle : 'Mass.gov home' }}" width="45" height="45" />
+      </picture>
     {% endif %}
 
     <span aria-hidden="true">{{ title ? title : 'Mass.gov' }}</span>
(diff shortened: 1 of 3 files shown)


packages/patternlab/styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
```diff
@@ -1,7 +1,10 @@
 <div class="ma__site-logo">
   <a href="{{ url ? url : '/' }}">
     {% if sealImage %}
-      <img src="{{ sealImage }}" alt="{{ altTitle ? altTitle : 'Mass.gov home' }}" width="45" height="45" />
+      <picture>
+        <source srcset="{{ sealWhiteImage }}" media="(prefers-color-scheme: dark), (forced-colors: active)" />
```
jarielrrz, line 5: Image

This media query may be too broad. (prefers-color-scheme: dark) matches macOS/Windows Dark Mode and Windows high contrast. Mass.gov does not invert in dark mode, so this loads stateseal-white.png on the light header and the same query in _header-search.scss sets search text/placeholder to --mf-c-font-inverse (white) on a white field.

Verified on the Tugboat with Dark Mode on and forced-colors off: header currentSrc is stateseal-white.png and typed search text is invisible.

AI suggested that forced-colors: active is the right target.

jarielrrz (changes requested):

packages/patternlab/styleguide/source/_patterns/01-atoms/09-media/site-logo.twig
tfleming-ma: @jarielrrz Let me know how this looks now. What I saw in my testing was that removing `prefers-color-scheme: dark` does prevent the OS light/dark mode from deciding what should be used, but the white logo and white text will be chosen in all high contrast modes, dark or light (e.g. Night Sky, Desert). With that said, there doesn't seem to be a way to detect whether a darker or lighter mode was selected with CSS. The `prefers-color-scheme` CSS seems to only care about the OS-level light or dark mode setting, at least from what I could tell in BrowserStack.

jarielrrz (approved):
LGTM

